PR #7, opened 2018-11-19
Upgrade to 2018 Edition

Hoverbear:
Follows from https://github.com/tikv/rfcs/pull/7.

PR #10, opened 2018-11-19
WIP Docs

Hoverbear:
(This is a WIP!)

Based off #7.

Adds some documentation, along with tested examples.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -16,57 +16,75 @@ use quick_error::quick_error;
 use std::{error, result};
 
 quick_error!{
+    /// An error originating from Raft or dependencies.
+    ///
+    /// This client currently uses [`quick_error`](https://docs.rs/quick-error/1.2.2/quick_error/)
+    /// for errors. *This may change in future versions.*
     #[derive(Debug)]
     pub enum Error {
+        /// Wraps a a `std::io::Error`.
         Io(err: ::std::io::Error) {
             from()
             cause(err)
             description(err.description())
         }
+        /// Wraps a `grpcio::Error`.
         Grpc(err: grpcio::Error) {
             from()
             cause(err)
             description(err.description())
         }
+        /// Represents that a futures oneshot channel was cancelled.
         Canceled(err: ::futures::sync::oneshot::Canceled) {
             from()
             cause(err)
             description(err.description())
         }
+        /// An unknown error.
+        ///
+        /// Generally, this is not an expected error.
         Other(err: Box<error::Error + Sync + Send>) {
             from()
             cause(err.as_ref())
             description(err.description())
             display("unknown error {:?}", err)
         }
+        /// A region was not found for the given key.
         RegionForKeyNotFound(key: Vec<u8>) {
             description("region is not found")
             display("region is not found for key {:?}", key)
         }
+        /// A region was not found.
         RegionNotFound(id: u64) {
             description("region is not found")
             display("region {:?} is not found", id)
         }
+        /// The peer is not a leader of the given region.
         NotLeader(region_id: u64) {
             description("peer is not leader")
             display("peer is not leader for region {:?}.", region_id)
         }
+        /// The store does not match.
         StoreNotMatch {
             description("store not match")
             display("store not match")
         }
+        /// The given key is not eithin the given region.
         KeyNotInRegion(key: Vec<u8>, region_id: u64, start_key: Vec<u8>, end_key: Vec<u8>) {
             description("region is not found")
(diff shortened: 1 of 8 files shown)


src/raw.rs
```diff
@@ -372,59 +443,281 @@ impl Future for Connect {
...
+    /// let req = connected_client.scan(inclusive_range, 2);
+    /// req.wait();
+    /// 
+    /// // Other possibilities:
+    /// let exclusive_range = KeyRef::from("TiKV")..KeyRef::from("TiDB");
+    /// let req = connected_client.scan(exclusive_range, None);
+    /// ```
+    pub fn scan<'client, 'keys, Bounds>(&'client self, range: Bounds, limit: impl Into<Option<u32>>) -> Scan<'client, 'keys, Bounds> 
+    where Bounds: RangeBounds<KeyRef<'keys>> {
+        use std::u32::MAX;
```
sunxiaoguang: If the limit could be potentially large, Scan and BatchScan should definitely be Stream so it can do multiple round of scan. Otherwise the default value of MAX will blow up memory.

src/lib.rs
```diff
@@ -97,6 +218,15 @@ impl Config {
...
     }
 
+    /// Set the certificate authority, certificate, and key locations for the [`Config`](struct.Config.html).
+    /// 
+    /// By default, TiKV connections do not have utilize transport layer security. Enable it by setting these values.
+    /// 
+    /// ```rust
+    /// # use tikv_client::Config;
+    /// let config = Config::new(vec!["192.168.0.100:2379", "192.168.0.101:2379"])
+    ///     .with_security("root.ca", "internal.cert", "internal.key");
```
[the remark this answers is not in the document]
sunxiaoguang: Having a dedicated type with constructor is equally tricky as this one. Therefore the type will have to make everything public to make it possible to construct by field name. 
How about break it into three methods to set them separately.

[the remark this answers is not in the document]
gregwebs: Three separate methods doesn't capture the nature of this: there are three arguments, and they must be given as all or none. The concern with the constructor is making it public? Won't any solution need to make something public?

src/raw.rs
```diff
@@ -270,15 +318,15 @@ impl<'a> Future for Scan<'a> {
 
 pub struct BatchScan<'a> {
     client: &'a Client,
-    ranges: Vec<(Key, Key)>,
+    ranges: Vec<(Bound<Key>, Bound<Key>)>,
```
sticnarf: What's your concern about collecting the iterator into a Vec? Is there any drawback of storing an iterator here (like the previous version)?

Hoverbear: I'd like to revisit doing that :)

src/raw.rs
sticnarf: Unfortunately, the scan rpc returns RepeatedField of KvPair. To make it a Stream, https://github.com/tikv/tikv/issues/3727 needs to be solved first.

src/errors.rs
```diff
@@ -11,61 +11,80 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::error;
-use std::result;
+use grpcio;
+use quick_error::quick_error;
+use std::{error, result};
 
 quick_error!{
```
siddontang: IMO, I want to try Failure instead of quick_error :-)

sunxiaoguang: Let me take a look into failure.

src/lib.rs
```diff
@@ -122,3 +288,23 @@ impl Config {
...
 }
+
+/// Unfortunately due to the API limitations of `RangeBound` we can only ever get an `&` (not a
+/// `&mut` or owned value) of a `Bound<T>`. So in order to coherce the value into the bound we want
+/// we must clone the data.
+/// 
+/// The only way to avoid this would be to only accept two `Bound<T>` arguments, not one
+/// `RangeBound<T>` argument. This would mean `"abc"..="xyz"` would not be possible.
+/// 
+// TODO: If you feel tricky please try to remove this clone!
```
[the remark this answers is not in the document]
siddontang: PTAL @sunxiaoguang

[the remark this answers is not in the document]
sunxiaoguang: Yea, I think this way is better it can avoid extra clone when bound is already a Key however make it possible to use non Key type which can be converted to Key.

src/errors.rs
Hoverbear: I also do! :)

src/lib.rs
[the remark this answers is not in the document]
Hoverbear: Great idea, I'll integrate this @sticnarf

Hoverbear:
Closed in favor of #13

PR #13, opened 2018-12-28
Docs migrated to initial version

Hoverbear:
@sunxiaoguang PTAL and merge freely, this isn't going to master, but your branch. :)


src/raw.rs
```diff
@@ -11,15 +11,296 @@
...
+impl Client {
+    /// Create a new [`Client`](struct.Client.html) once the [`Connect`](struct.Connect.html) resolves.
+    ///
+    /// ```rust,no_run
+    /// use tikv_client::{Config, raw::Client};
+    /// use futures::Future;
+    /// let connect = Client::new(&Config::default());
+    /// let client = connect.wait();
+    /// ```
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::new_ret_no_self))]
```
sunxiaoguang: I guess `feature = "cargo-chippy"` can be removed now?

src/raw.rs
```diff
@@ -11,15 +11,296 @@
...
+            self.rpc(),
+            BatchDeleteInner::new(keys.into_iter().map(Into::into).collect()),
+        )
+    }
+
+    /// Create a new [`Scan`](struct.Scan.html) request.
+    ///
+    /// Once resolved this request will result in a scanner over the given keys.
+    ///
+    /// If not passed a `limit` parameter, it will default to `u32::MAX`.
```
sunxiaoguang: Shell we put a rational limit on this? Like what Go client does.

src/raw.rs
```diff
@@ -11,15 +11,296 @@
...
+                each_limit.into().unwrap_or(u32::MAX),
+            ),
+        )
+    }
+
+    /// Create a new [`DeleteRange`](struct.DeleteRange.html) request.
+    ///
+    /// Once resolved this request will result in the deletion of all keys over the given range.
+    ///
+    /// If not passed a `limit` parameter, it will default to `u32::MAX`.
```
sunxiaoguang: DeleteRange has no number of pairs limit, so we can remove this line.

PR #22, opened 2019-01-14
Refine CI

Hoverbear:
Use cross platform CI. Linux Stable is red ATM due to a clippy problem, don't block merge for that.

PR #24, opened 2019-01-19
Change raw::Client::get to return Option<Value>

sunxiaoguang:
As what @sticnarf suggested in #23, I also agree that returning Option<Value> seems to be better for get. PTAL @Hoverbear when you have time, thanks.


Hoverbear:
Force merging since #22 is not merged.

PR #25, opened 2019-01-20
Remove unwraps in raw client methods

Kerollmops:
Do not panic when using unbounded ranges on certain methods.


Hoverbear:
Right. @Kerollmops Thank you for this PR, you found a bug. :)

I will close this and we can open a bug for this.

PR #26, opened 2019-01-20
AsRef/Mut and From traits implementation on Key and Value

Kerollmops:
Adding some missing traits and fixing unsound lifetimes restrictions on `Key/Value::From` impls.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,11 +108,9 @@ impl From<String> for Key {
     }
 }
 
-impl<'a> From<&'static str> for Key {
-    fn from(v: &'static str) -> Key {
-        let mut vec = Vec::new();
-        vec.extend_from_slice(v.as_bytes());
-        Key(vec)
+impl From<&str> for Key {
+    fn from(v: &str) -> Key {
+        Key(v.as_bytes().to_vec())
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -206,11 +216,9 @@ impl From<String> for Value {
     }
 }
 
-impl From<&'static str> for Value {
-    fn from(v: &'static str) -> Value {
-        let mut vec = Vec::new();
-        vec.extend_from_slice(v.as_bytes());
-        Value(vec)
+impl From<&str> for Value {
```
sunxiaoguang: If I understand it correctly, the extra static lifetime requirement is trying to avoid copying without user actually noticing it. Creating Key/Value from &str can be done by calling to_string() explicitly or simply move the String where the &str was created from.

Hoverbear: Right. :) The reason `&'static` is supported is because this is almost always a `const` or a raw string in the code, which the user would expect to be cloned to be used in this way.

Maybe it should be changed? 🤔

Hoverbear: @Kerollmops Any chance you have time to follow up on this?

Kerollmops: The problem is that it will not fix the problem, the `Value` type require a `Vec` internally, so the static str will need to be transformed into a `Vec` anyway.

Hoverbear: Right, but the intention is that `&'static str` are used for examples or testing code, we don't want people to use it for `&'a str` or other random borrows, just for, for example

This is ok:

```rust
let test_key = "blah"`;
Value::from(test_key);
```

While this shouldn't work, since it introduces invisible copying of data which is probably real:

```rust
let test_key = String::from("blah");
Value::from(&test_key)
```

Hoverbear:
Forcing merge due to CI.

THanks so much for this contribution! :) Really appreciate it.

PR #28, opened 2019-01-20
Improve the raw example

Kerollmops:
This example panic when encounting the scan key results.

Spotting two bugs:
 - the `raw::Client::scan` does not understand inclusive bounds correctly
 - the `raw::Client::delete_range` does not delete anything

### DIFF
--- a/examples/raw.rs
+++ b/examples/raw.rs
@@ -33,7 +33,7 @@ fn main() -> Result<()> {
         Config::new(args.pd)
     };
 
-    // When we first create a client we recieve a `Connect` structure which must be resolved before
+    // When we first create a client we receive a `Connect` structure which must be resolved before
     // the client is actually connected and usable.
     let unconnnected_client = Client::new(&config);
     let client = unconnnected_client.wait()?;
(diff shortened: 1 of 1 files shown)


ice1000:
Please remove the unused import:

```
  --> examples/raw.rs:18:61
   |
18 | use tikv_client::{raw::Client, Config, Key, KvPair, Result, Value};
   |                                                             ^^^^^
```

Hoverbear:
@sunxiaoguang that was the intention. Do you think it's better to remove them?

sunxiaoguang:
> @sunxiaoguang that was the intention. Do you think it's better to remove them?

I personally had difficulties when I just started learning Rust and had to turn to IDE to know the real types. So as a vim user, I think type annotation is more friendly for new learners of Rust.

sunxiaoguang:
Maybe two separate examples for rust new learners and TiKV new users? They probably like different styles.

Kerollmops:
I rebased on master to retrieve the modifications about the raw get method returning an optional value and modified the example in consequence.

I also kept the type annotations to make it clear for new users, I think it is not necessary to create multiple example for the same think. Advanced Rust users will understand the beginner example.

I added some comments too.

examples/raw.rs
```diff
@@ -56,49 +55,60 @@ fn main() -> Result<()> {
...
         .wait()
-        .expect_err("Get returned value for not existing key");
+        .expect("Could not get just deleted entry");
+    assert!(value.is_none());
 
-    let pairs: Vec<KvPair> = (1..3)
-        .map(|i| KvPair::from((Key::from(format!("k{}", i)), Value::from(format!("v{}", i)))))
-        .collect();
+    // You can ask to write multiple key-values at the same time, it is much more
+    // performant because it is passed in one request to the key-value store.
```
Hoverbear: Great comment. :)

Hoverbear:
Thanks so much for this contribution @Kerollmops !

Hoverbear:
@siddontang It's failing because Raft uses a `feature()` https://github.com/pingcap/raft-rs/pull/180, an upgrade is pending release https://github.com/pingcap/raft-rs/pull/182

brson:
I do notice that this is failing on rustfmt, though there are existing rustfmt failures. Since the CI is reportedly busted, and it's a pre-existing problem, and this has been waiting for too long, I'm going to force merge. Someone else will clean up the testing mess later.

PR #36, opened 2019-03-18
Add `Send` trait bound on `KvFuture`

Snnappie:
Currently none of the `raw` API is usable with tokio, since the `KVFuture` `dyn Future` isn't required to implement `Send`.

This pull request simply adds the `Send` bound. This doesn't seem to cause any problems as far as I can tell, but maybe it would?


nrc:
I think it would only cause problems for clients, not inside the library itself. Any future that is not `Send` will no longer be usable in `KvFuture`.

Hoverbear:
@Snnappie Can I ask you to signoff your commit? If you do a `git commit --amend --signoff` onto your last commit it will do this. :)

Snnappie:
Hi @Hoverbear - I've tried the sign-off a few different ways and it doesn't seem to be working. I'm probably doing something wrong. Any suggestions?

ice1000:
> Hi @Hoverbear - I've tried the sign-off a few different ways and it doesn't seem to be working. I'm probably doing something wrong. Any suggestions?

You may try rebasing on master. The problem is that you have one commit not signoffed before a "merge master" commit, which makes dco unhappy.

Hoverbear:
Hi @Snnappie, sorry the DCO is giving you problems. :(

You've successfully signed off https://github.com/tikv/client-rust/pull/36/commits/c8cf77ba1405ebaefc310ec4db868bb79b357380, however you did not set your Git config's user.name and user.email. You can see https://confluence.atlassian.com/bitbucket/configure-your-dvcs-username-for-commits-950301867.html#ConfigureyourDVCSusernameforcommits-ConfigureyourGitusername/email

This is kind of a mess. :(

I recommend you:

* set your git user.name and user.email globally (this will resolve the problem for the future on all projects)
* do `git rebase -i origin/master` and just delete the merge commits. 
* at this point you should only have c8cf77b  in this PR
* You can reset the author of c8cf77b by doing `git commit --amend --reset-author --signoff` (https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---reset-author) to set the author correctly. 
* You can then force push an update to this PR with `git push --force`.


Sorry about this complication, but it's due to lack of configuration and it shouldn't be a problem with future PRs. You can see here why we use the DCO: https://stackoverflow.com/a/35238070. It protects our project from legal action.

If you'd prefer, we could just make this change as well since it's a one-line change.

nrc:
> I think it would only cause problems for clients, not inside the library itself. Any future that is not Send will no longer be usable in KvFuture.

I think this is incorrect. I assumed users would implement `KvFuture`, but I don't think they ever would. So r+ :-)

nrc:
The easiest way to fix the commit mess is something like:

```
git checkout master
git reset HEAD~9 --hard
git pull upstream master
git checkout -b send
git cherry-pick 01cf70166a08374c83f38d3ab687616111e92476
git reset HEAD~1
git commit -a -s -m "Add Send trait bound on KvFuture"
git push origin send:master -f
```

Snnappie:
@Hoverbear I do have `user.email` and `user.name` set globally. `user.email` was updated to a new address which wasn't in Github at first. That may have been the issue.

Thanks @nrc that seems to have fixed it.

Hoverbear:
🎉 Thanks =D

PR #38, opened 2019-03-20
Fix infinite loop on raw-kv requests that span multiple regions

Snnappie:
When working with a DB that has enough data to have multiple regions (that is, one of `start_key, end_key` is not empty for the region), any multi-region request (`batch_put`, `batch_get`, etc.) will loop infinitely, repeatedly asking PD for the region info. This is because the `loop_fn` in `Rpc::group_tasks_by_region` can't make progress due to a bug in `KeyLocation::contains`. `contains` is incorrectly comparing `start <= key && end < key` (should be `key < end`).
Since `KeyLocation: Deref<Target = Region>`, and `Region::contains` is a correct implementation, this merge request simply removes the buggy `KeyLocation::contains`.


Hoverbear:
Hi @Snnappie , can you signoff your commits please? You can do this with `git rebase -i master --signoff`

Hoverbear:
Hi @Snnappie, seems we have the same DCO problem here. :( https://github.com/tikv/client-rust/pull/36#issuecomment-490256874 has some potential help, but before we can merge on any TiKV (or DCO adhering repo on Github) we need you Git `user.name` and `user.email` configured correctly and the commit signed off with `--signoff`

Otherwise, it seems this PR contains stuff from #36. Could you separate them and we can merge this one? It LGTM while the other one seems to have pending discussion.

Snnappie:
Closing in favor of #47

PR #40, opened 2019-04-29
Some very minor improvements

nrc:
Mostly to get us compiling without warnings


ice1000:
Missing sign-offs

nrc:
@brson ptal

src/rpc/security.rs
```diff
@@ -77,8 +77,8 @@ impl SecurityManager {
     {
         info!("connect to rpc server at endpoint: {:?}", addr);
         let addr = addr
-            .trim_left_matches("http://")
-            .trim_left_matches("https://");
+            .trim_start_matches("http://")
+            .trim_start_matches("https://");
```
ice1000: Can we use a regex here? Like `https?://`

nrc: We could! There are about a million things we could improve. Could you file an issue for this (motivation here is just to clear the deprecation warnings rather than make this awesome).

PR #41, opened 2019-05-06
Migrate to futures 0.3 from 0.1

nrc:
~~And some minor refactoring. Based on #40, it is worth us reviewing and landing that first.~~

PTAL @sunxiaoguang

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,17 +12,23 @@ With this crate you can easily connect to any TiKV deployment, interact with it,
 
 This is an open source (Apache 2) project hosted by the Cloud Native Computing Foundation (CNCF) and maintained by the TiKV Authors. *We'd love it if you joined us in improving this project.*
 
-## Install
+## Using the client
 
-There are no special requirements to use this. It is a Rust 2018 edition crate supporting stable and nightly.
+The TiKV client is a Rust library (crate). It requires version 1.36 of the compiler and standard libraries (which will be stable from the 4th July 2019, see below for ensuring compatibility).
 
-To use this crate in your project, add it as a dependency in the `Cargo.toml` of your Rust project:
+To use this crate in your project, add it as a dependency in your `Cargo.toml`:
 
 ```toml
 [dependencies]
 # ...Your other dependencies...
-tikv-client = "~0.1"
+tikv-client = { git = "https://github.com/tikv/client-rust.git" }
 ```
+
+The client requires a Git dependency until we can [publish it](https://github.com/tikv/client-rust/issues/32).
+
+There are [examples](examples) which show how to use the client in a Rust program.
+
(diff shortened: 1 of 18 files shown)


nrc:
Rebased to remove conflict, merged the last three commits to make it easier to review

nrc:
Rebased on to #42 (since I expect that to land first). PTAL at the second and third commits @Hoverbear and @sunxiaoguang

Hoverbear:
This does cause us not to follow https://github.com/tikv/rfcs/pull/7/files#diff-3333e76d7edf3ed642c27ac78f1795a6R25 until these features are stabilized.

That said: The RFC isn't accepted.

I think it's fine. What do you think @sunxiaoguang?

ice1000:
async/await! Cool!

README.md
```diff
@@ -37,3 +43,31 @@ You can access the documentation on your machine by running the following in any
...
+
+## Toolchain versions
+
+To check what version of Rust you are using, run
+
+```bash
+rustc --version
+```
+
+you'll see something like `rustc 1.36.0-nightly (a784a8022 2019-05-09)` where the `1.36.0` is the toolchain version, and `nightly` is the channel (stable/beta/nightly). To install another toolchain use
```
Hoverbear: These sentences have inconsistent capitalization. ('To build', '(where `nightly`', 'you'll see')

README.md
```diff
@@ -12,18 +12,24 @@ With this crate you can easily connect to any TiKV deployment, interact with it,
...
 # ...Your other dependencies...
-tikv-client = "~0.1"
+tikv-client = { git = "https://github.com/tikv/client-rust.git" }
 ```
 
+The client requires a Git dependency until we can [publish it](https://github.com/tikv/client-rust/issues/32).
+
+There are [examples](examples) which show how to use the client in a Rust program.
+
+The examples and documentation use async/await syntax. This is a new feature in Rust and is currently unstable. To use async/await you'll need to add the feature flag `#![]` to your crate and use a nightly compiler (see below).
```
Hoverbear:
```suggestion
The examples and documentation use async/await syntax. This is a new feature in Rust and is currently unstable. To use async/await you'll need to add the feature flag `#![feature(async_await, await_macro)]` to your crate and use a nightly compiler (see below).
```

src/lib.rs
```diff
@@ -31,7 +31,7 @@
 //! [dependencies]
```
Hoverbear: A couple lines above this we say we support stable but we don't now!

nrc: Hmm, looks like we duplicate the README in the crate docs. IMO, it is better not to include installation instructions in the crates docs, just how to use the API. I've updated the PR with that change, what do you think?

Issue #37, opened 2019-03-20
Panicked when running example/raw.rs


c4pt0r:
Output:

```
Put key "TiKV", value "Rust".
Get key Key(54694B56) returned value Some(Value("Rust")).
Key: Key(54694B56) deleted
Found values: [KvPair(6B31, "v1"), KvPair(6B32, "v2")] for keys: [Key(6B31), Key(6B32)]
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[Key(6B31)]`,
 right: `[Key(6B31), Key(6B32)]`', src/main.rs:103:5
```

I think the problem is (in raw.rs):
```rust
    // Scanning a range of keys is also possible giving it two bounds
    // it will returns all entries between these two.
    let start = "k1";
    let end = "k2";
    let pairs = client
        .scan(start..=end, 10)
        .key_only()
        .wait()
        .expect("Could not scan");

    let keys: Vec<_> = pairs.into_iter().map(|p| p.key().clone()).collect();
    assert_eq!(&keys, &[Key::from("k1"), Key::from("k2")]);
    println!("Scaning from {:?} to {:?} gives: {:?}", start, end, keys);
```
        
Does `scan` in this case return the right border?

sunxiaoguang:
Hm, looks like inclusive range isn't implemented correctly. 
Thanks for digging out this issue. I will try fixing this.

c4pt0r:
@sunxiaoguang still opening?

PR #44, opened 2019-05-06
Add simple unit tests for our KeyRange implementations.

Hoverbear:
Adds some unit tests for KeyRange to check their validity.

Also sets the behavior of `..x` to go from `[0]` to whatever x is, instead of erroring.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -607,6 +607,7 @@ pub trait KeyRange: Sized {
 fn range_to_keys(range: (Bound<Key>, Bound<Key>)) -> Result<(Key, Option<Key>)> {
     let start = match range.0 {
         Bound::Included(v) => v,
+        // TODO: Is this even possible?
         Bound::Excluded(mut v) => {
             match v.last_mut() {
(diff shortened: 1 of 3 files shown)


Hoverbear:
This was originally to fix #37 but seems the bug is not here.

src/lib.rs
```diff
@@ -640,43 +641,106 @@ impl<T: Into<Key>> KeyRange for Range<T> {
...
 
+#[test]
+fn test_bound_bound() -> Result<()> {
+    let input = (Bound::Included(vec![0]), Bound::Included(vec![10]));
+    let output = input.into_keys()?;
+    let expected = (Key::from(vec![0]), Some(Key::from(vec![10])));
+    assert_eq!(output, expected);
+    Ok(())
+}
+
```
nrc: Could you move all the tests to the test mod please?

Hoverbear: Yes!

Hoverbear:
Let's hold on merging this until after #45

Hoverbear:
#45 adds tested doctests.

PR #45, opened 2019-05-06
Fix Range bug.

### LINKED ISSUE #37: Panicked when running example/raw.rs

Hoverbear:
This includes some logic to deal with TiKV intrinsics.

Fixes #37 .


src/rpc/tikv/client.rs
```diff
@@ -556,7 +556,19 @@ impl KvClient {
         start_key
             .map(|k| req.set_start_key(k.into_inner()))
             .unwrap();
-        end_key.map(|k| req.set_end_key(k.into_inner())).unwrap();
+        end_key
+            .map(|k| {
+                // Scan in TiKV has a weird particularity to it.
+                // The **start** of a scan is inclusive, unless appended with an '\0', then it is exclusive.
+                // The **end** of a scan is exclusive, unless appended with an '\0', then it is inclusive.
```
nrc: I would expect the scan range to be inclusive/exclusive and if TiKV does that, then the client should too, I think. Using the `\0`s is weird and we should work around it. Also Rust ranges are by default inclusive/exclusive.

Hoverbear: The Client handles ranges from all external code by taking `x..y` or `x..=y` or whatever else. The user doesn't need to think  about it. The way this works is by always building a set of tuples that create (inclusive) bounds and pass those around. This means all internal APIs just deals with inclusive range bounds instead of range type objects. (This is primarily because ranges are challenging to work with if you want to avoid copying all over the place. So we just copy once to take ownership and pass it around

nrc: Hmm, so that seems like a mistake in the implementation of `KeyRange`, but we should probably land this as is and address that later (I have also refactored `KeyRange` in one of the commits I have queued up, so I think we can land this and that one and then look at `KeyRange`'s treatment of the end key).

Hoverbear: What seems like a mistake in `KeyRange`?

nrc: Having the upper bound be inclusive

Hoverbear:
@nrc I think this is a better solution, please look again?

src/lib.rs
```diff
@@ -599,6 +588,33 @@ pub type KvFuture<Resp> = Box<dyn Future<Item = Resp, Error = Error> + Send>;
 /// which means all of the above types can be passed directly to those functions.
 pub trait KeyRange: Sized {
     fn into_bounds(self) -> (Bound<Key>, Bound<Key>);
+    /// Ranges used in scanning TiKV have a particularity to them.
+    ///
+    /// The **start** of a scan is inclusive, unless appended with an '\0', then it is exclusive.
```
nrc:
```suggestion
    /// The **start** of a scan is inclusive, unless prepended with an '\0', then it is exclusive.
```

src/lib.rs
```diff
@@ -608,24 +624,19 @@ fn range_to_keys(range: (Bound<Key>, Bound<Key>)) -> Result<(Key, Option<Key>)>
     let start = match range.0 {
         Bound::Included(v) => v,
         Bound::Excluded(mut v) => {
-            match v.last_mut() {
-                None | Some(&mut U8_MAX) => v.push(0),
-                Some(v) => *v += 1,
-            }
-            v
+            let mut buf = b"\0".to_vec();
```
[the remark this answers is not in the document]
Hoverbear: Great idea!

src/lib.rs
Hoverbear: Please see https://github.com/tikv/client-go/blob/master/rawkv/rawkv.go#L267

src/kv.rs
```diff
@@ -467,10 +477,14 @@ impl<T: Into<Key>> KeyRange for (Bound<T>, Bound<T>) {
...
+    K: Into<Key>,
+{
+    use std::ops::Bound::*;
     match b {
-        Bound::Included(k) => Bound::Included(k.into()),
-        Bound::Excluded(k) => Bound::Excluded(k.into()),
-        Bound::Unbounded => Bound::Unbounded,
+        Included(k) => Included(k.into()),
+        Excluded(k) => Excluded(k.into()),
+        Unbounded => Unbounded,
```
nrc: Are these changes intentional? I prefer the original version

PR #47, opened 2019-05-13
Fix infinite loop in KeyLocation::contains (round 2)

Snnappie:
Fix bug in KeyLocation::contains which would cause an infinite loop on any batch operation which spanned multiple regions.

Signed-off-by: Jason Boatman <jason.boatman@multiscale.io>


Snnappie:
@Hoverbear Instead of fiddling with git until I could get the DCO accepted, I just opened a new pull request with the changes from #38

Hoverbear:
Thanks @Snnappie. IT LGTM. :) Some development on kvproto broke our CI (Fixed by https://github.com/tikv/client-rust/pull/49) is breaking the CI. Once we fix that I'll shepherd this to merge.

Snnappie:
> From #38 it sounds like the fix here is easy (reversing the order of comparison on line 545, so why are we deleting the `contains` function rather than fixing it (sorry if this has been discussed, I couldn't find discussion here or on #38).

The `contains` function is redundant with `Region::contains`, which is available via `Deref`. If you think it is better to have a separate contains function I think it should at least be implemented as

```rust
fn contains(&self, key: &Key) -> bool {
  Region::contains(&self.0, key)
}
```
But since the `KeyLocation` struct isn't used in any other way than to be a wrapper around `Region`, it made more sense to me to just remove the `KeyLocation::contains` method.

nrc:
Thanks for the explanation, that all makes sense!

PR #42, opened 2019-05-06
Migrate to Prost

nrc:
~~Based on #41, lets not land this until after #41.~~

@Hoverbear PTAL at commit https://github.com/tikv/client-rust/commit/24e12022eb5ebb30239711e434d3602dc2084fc1

PR #49, opened 2019-05-15
Use new kvproto raft branch update

Hoverbear:
Use the new version of kvproto on the branch we're depending on (raft update https://github.com/pingcap/kvproto/pull/357) so that we can build properly still.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,10 @@ name = "tikv_client"
 failure = "0.1"
 futures = "0.1"
 fxhash = "0.2"
-grpcio = { version = "0.4", features = [ "secure" ] }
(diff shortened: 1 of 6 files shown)


Hoverbear:
@nrc I'm aware of this and unfortunately #42 depends on #41 which is unlikely to get merged quickly (it makes us nightly only and uses currently unstable features)

The goal of this is to make it possible to *use* this library again as a result of changes on the kvproto repo. I avoided taking out `protobuf` since that requires the use of things like `new_()` which have a learning burden and some mental overhead. It'd be better do do a full migration to Prost without the wrapper I think.

Hoverbear:
Yay! Please put eyes on #41 / #42 as they're a long term solution.

PR #52, opened 2019-05-23
rpc: user regexp replace string match scheme

hello2dj:
Signed-off-by: dengjie <dengjie.hello2dj@bytedance.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ integration-tests = []
 name = "tikv_client"
 
 [dependencies]
+regex = "1"
 failure = "0.1"
 futures = "0.1"
 fxhash = "0.2"
--- a/src/rpc/security.rs
+++ b/src/rpc/security.rs
@@ -1,17 +1,17 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
 use std::{
+    borrow::Borrow,
     fs::File,
     io::Read,
     path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
-    borrow::Borrow,
 };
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -18,6 +18,7 @@ integration-tests = []
 name = "tikv_client"
 
 [dependencies]
+regex = "1"
```
ice1000: Could you please use a more generic version? Like `^1` (or probably something better)?

ice1000:
Please run `cargo fmt`

ice1000:
Missing signoffs, LGTM

hello2dj:
thank you

Cargo.toml
nrc: I think `1` is fine - it means any version back compat with 1 (i.e., the same as `0.1` for the other deps)

src/rpc/security.rs
```diff
@@ -65,9 +72,10 @@ impl SecurityManager {
         Factory: FnOnce(Channel) -> Client,
     {
         info!("connect to rpc server at endpoint: {:?}", addr);
-        let addr = addr
-            .trim_start_matches("http://")
-            .trim_start_matches("https://");
+
+        let cow_addr = SCHEME_REG.replace(addr, "");
+        let addr: &str = cow_addr.borrow();
```
nrc: Could you use `&addr` below and avoid the explicit call to `borrow`?

hello2dj: done

nrc:
CI is failing due to needing https://github.com/tikv/client-rust/pull/60

PR #56, opened 2019-05-30
Some trivial tidying up

nrc:
The first commit might have some small performance impact, but it should be negligible and I think it is better to use the standard library until it is proven to have performance implications.


nrc:
PTAL @Hoverbear and/or @brson

PR #58, opened 2019-05-30
Use an enum for `ColumnFamily`

nrc:
Means we have proper typing rather than stringly typing.

PTAL @Hoverbear

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -9,14 +9,9 @@
 //!
 //! **Warning:** It is not advisable to use both raw and transactional functionality in the same keyspace.
 //!
-use crate::{rpc::RpcClient, Config, Error, Key, KeyRange, KvPair, Result, Value};
+use crate::{rpc::RpcClient, Config, Error, ErrorKind, Key, KeyRange, KvPair, Result, Value};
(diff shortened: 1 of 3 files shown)


src/raw.rs
```diff
@@ -292,54 +287,64 @@ impl Future for Connect {
...
-///
-/// This is a *wrapper type* that implements `Deref<Target=String>` so it can be used like one transparently.
-///
 /// **But, you should not need to worry about all this:** Many functions which accept a
 /// `ColumnFamily` accept an `Into<ColumnFamily>`, which means all of the above types can be passed
 /// directly to those functions.
-#[derive(Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
-pub struct ColumnFamily(String);
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum ColumnFamily {
```
Hoverbear: This is user configurable: https://github.com/tikv/tikv/wiki/TiKV-FAQ#what-are-the-features-of-tikv-block-cache

So I don't think this is appropriate.

PR #62, opened 2019-06-04
Fix a warning

nrc:
Caused by an incorrect code conversion in #41

Signed-off-by: Nick Cameron <nrc@ncameron.org>


nrc:
This patch is needed to pass CI with current master (I guess a compiler update means this is warned when it wasn't previously). The problem is blocking any PR in the queue.

PR #63, opened 2019-06-04
Make `ColumnFamily` a little more opaque

nrc:
Derived from #58
There is not much practical effect here, we're still wrapping a String, but implementing `Display` rather than `into_inner` which encapsulates the implementation of `ColumnFamily` a bit better. Removes the `Deref` impl because `ColumnFamily` is not a pointer type. We need to change the `From` impl a little due to coherence with `Display`, but I think it is better in any case, `ToString` is a very broad trait.

PTAL @Hoverbear

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -11,12 +11,7 @@
 //!
 use crate::{rpc::RpcClient, Config, Error, Key, KeyRange, KvPair, Result, Value};
 use futures::{future, task::Context, Future, Poll};
-use std::{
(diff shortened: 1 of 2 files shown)


Hoverbear:
@nrc You mentioned this as a caveat:

> So `"write".from()` would work but `"foo"` would require something more explic

nrc:
> @nrc You mentioned this as a caveat:

>    So "write".from() would work but "foo" would require something more explic

I meant that about a potential future version, not this PR.

PR #67, opened 2019-06-18
Force local and Travis builds to use an old nightly

nrc:
Work around an ICE (e.g., in #66)

(Also fixes the Travis link from the README)


Hoverbear:
Merging since this is a blocker and a minor workaround.

PR #71, opened 2019-06-19
Refactor KvRange

nrc:
The end result is a concrete type `BoundRange` and using `Into<BoundRange>`
where functions previously took `KeyRange`. This refactoring means that errors
with invalid ranges are found earlier (in fact, they are mostly prevented statically,
where they can still occur, they do so when ranges are created, not when used)
and there is less conversion between range types in some cases. Hopefully,
code is simplified somewhat too.

PTAL @Hoverbear @brson @sunxiaoguang

### DIFF
--- a/src/kv.rs
+++ b/src/kv.rs
@@ -1,8 +1,8 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::ops::{
-    Bound, Deref, DerefMut, Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, RangeToInclusive,
-};
+use std::cmp::{Eq, PartialEq};
+use std::convert::TryFrom;
+use std::ops::{Bound, Deref, DerefMut, Range, RangeFrom, RangeInclusive};
 use std::{fmt, str, u8};
 
 use crate::{Error, Result};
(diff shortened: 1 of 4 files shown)


src/kv.rs
```diff
@@ -320,171 +350,179 @@ impl fmt::Debug for KvPair {
...
+        // Debug assert because this function is private.
+        debug_assert!(from != Bound::Unbounded);
+        BoundRange { from, to }
+    }
+
     /// Ranges used in scanning TiKV have a particularity to them.
     ///
     /// The **start** of a scan is inclusive, unless appended with an '\0', then it is exclusive.
     ///
     /// The **end** of a scan is exclusive, unless appended with an '\0', then it is inclusive.
```
Hoverbear: Perhaps we should be noting this function adds these `\0`s?

src/rpc/client.rs
```diff
@@ -373,14 +374,14 @@ impl RpcClient {
                     .map_ok(|(region, client)| {
                         (scan, region.range(), RawContext::new(region, client, cf))
                     })
-                    .and_then(|(mut scan, region_range, context)| {
-                        let (start_key, end_key) = scan.range();
+                    .and_then(|(scan, region_range, context)| {
+                        let (ref start_key, ref end_key) = scan.range;
```
Hoverbear: Any way we can take here to avoid the clone?

nrc: I think we can, but it requires some significant refactoring of `ScanRegionsContext`. That will clash quite a lot with https://github.com/tikv/client-rust/pull/70, so I'd prefer to leave it until after we land both these PRs. I would like to do that refactoring though.

src/kv.rs
```diff
@@ -320,171 +350,179 @@ impl fmt::Debug for KvPair {
...
-/// let from_range_full = range_full.into_bounds();
-/// assert_eq!(
-///     (Bound::Unbounded, Bound::Unbounded),
-///     from_range_full
+///     (Bound::Included(Key::from("Rust")), Bound::Unbounded),
 /// );
 /// ```
 ///
-/// **But, you should not need to worry about all this:** Many functions accept a `impl KeyRange`
+/// **But, you should not need to worry about all this:** Many functions accept a `impl Into<BoundRange>`
```
brson:
```suggestion
/// **But, you should not need to worry about all this:** Most functions which operate
/// on ranges will accept any types  which implement `Into<BoundRange>`.
```

brson: This gets rid of "a impl" by repeating the more clearly written line from the first paragraph.

PR #73, opened 2019-06-20
Refactor everything

nrc:
~~I'm not sure about this. The only way I could think to avoid boxing futures is to use existential types, however, that is unstable (will require nightly) and is unlikely to be stabilised soon (afaik). On the other hand, if we want the client to be as fast as possible, we have to avoid allocating for our futures. Long term we will probably want to use `async` methods rather than existential types, we could try that now or put off doing this until `async` is stable or migrate from this PR to async methods.~~

See https://github.com/tikv/client-rust/pull/73#issuecomment-504794389

WDYT @Hoverbear @brson @sunxiaoguang ?


Hoverbear:
So once async is stabilized this PR isn't needed?

nrc:
> So once async is stabilized this PR isn't needed?

Maaaaaaybe. Moving to async is going to be a big thing and will change the code quite a lot. I'm not sure how it will pan out.

> Maybe you could put nightly only behind a feature flag?

Unfortunately not, these changes are quite invasive

I think I will look at whether we need `RequestInner` at all. If we can replace that with plain functions, then we should just be able to use (stable) impl Trait, rather than (unstable) existential types.

nrc:
So, this turned into 'refactor all the things' - just to try and get rid of the boxing and a few clones. On the bright side, I've ticked off most of the things I wanted to refactor and I don't think the API has changed much (technically the raw API is returning opaque types rather than custom futures, but I don't think that makes a difference to users). The major change is that to use column families (or `key_only`) you call a builder-style method on the client instead of a future. I think this is better since it makes making multiple calls with the same CF more ergonomic and it means the request futures are not doing double duty as builder.

This also addresses https://github.com/tikv/client-rust/pull/71/files/9b72d47dce2aa58a9898b10bb8ea8b53fa7a02b8#r295903593 as best as possible - in fact it removes quite a few clones from various places and makes some others more explicit.

PTAL @Hoverbear and @brson. Apologies for the large PR, but you can review each commit at a time to make things easier.

examples/raw.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
 #![feature(async_await, await_macro)]
+#![type_length_limit = "3081103"]
```
Hoverbear: This is a [long, long](https://www.youtube.com/watch?v=6-1Ue0FFrHY) type!

src/raw/client.rs
```diff
@@ -1,15 +1,112 @@
...
 pub struct Client {
     rpc: Arc<RpcClient>,
+    cf: Option<ColumnFamily>,
+    key_only: bool,
+}
+
+// The macros below make writing `impl Client` concise and hopefully easy to
+// read. Otherwise, the important bits get lost in boilerplate.
+
+// `request` and `scan_request` define public functions which return a future for
```
Hoverbear: Maybe these can be doc comments on each macro?

src/raw/client.rs
```diff
@@ -28,12 +125,55 @@ impl Client {
         Connect::new(config)
     }
 
-    #[inline]
-    fn rpc(&self) -> Arc<RpcClient> {
-        Arc::clone(&self.rpc)
+    /// Set the column family of requests.
+    ///
+    /// This function returns a new `Client`, requests created with it will have the
+    /// supplied column family constraint. The original `Client` can still be used.
```
Hoverbear: Oh! What a neat idea.

src/raw/client.rs
```diff
@@ -28,12 +125,55 @@ impl Client {
...
+    ///
+    /// ```rust,no_run
+    /// # #![feature(async_await)]
+    /// # use tikv_client::{Config, raw::Client};
+    /// # use futures::prelude::*;
+    /// # futures::executor::block_on(async {
```
Hoverbear: 👍

src/rpc/client.rs
```diff
@@ -258,193 +118,259 @@ impl RpcClient {
...
-                            Loop::Break(scan.result)
-                        } else {
-                            match scan.next(region_range) {
-                                ScanRegionsStatus::Continue => Loop::Continue((inner, scan)),
-                                ScanRegionsStatus::Break => Loop::Break(scan.result),
-                            }
-                        }
-                    })
+        self.regions_for_range(range)
+            .try_fold(Vec::new(), move |mut result, context| {
```
Hoverbear: I wonder if we can guessimate the required capacity for the vector to save reallocations?

src/rpc/client.rs
```diff
@@ -258,193 +118,259 @@ impl RpcClient {
...
-            })
+    ) -> impl Stream<Item = Result<RawContext>> {
+        let (start_key, end_key) = range.into_keys();
+        stream_fn(Some(start_key), move |start_key| {
+            let start_key = match start_key {
+                None => return Either::Right(ready(Ok(None))),
+                Some(sk) => sk,
+            };
+            let end_key = end_key.clone();
+            let this = self.clone();
```
Hoverbear: I feel like I'm in javascript land. >.>

src/rpc/client.rs
```diff
@@ -258,193 +118,259 @@ impl RpcClient {
...
+    }
+
+    fn group_tasks_by_region<Task>(
+        self: Arc<Self>,
+        tasks: Vec<Task>,
+    ) -> impl Stream<Item = Result<(RegionId, Vec<Task>)>>
+    where
+        Task: GroupingTask,
+    {
+        let tasks: VecDeque<Task> = tasks.into();
```
Hoverbear: Is this efficient? It *looks* expensive.

src/transaction/transaction.rs
```diff
@@ -11,27 +12,12 @@ use std::ops::RangeBounds;
 /// particular timestamp obtained from the placement driver.
 ///
 /// Once a transaction is commited, a new commit timestamp is obtained from the placement driver.
+#[derive(new)]
 pub struct Transaction {
     snapshot: Snapshot,
 }
 
 impl Transaction {
-    /// Create a new transaction operating on the given snapshot.
```
Hoverbear: Seems this documentation will be gone now, could we keep it?

examples/raw.rs
nrc: What did I just watch? :-p

src/raw/client.rs
nrc: Do you mean the macro uses or definitions? If the latter, I usually only put doc comments on public API, but I don't mind either way.

src/rpc/client.rs
nrc: Given my experience with Prost, I'd be hesitant to do this before we can prove that it improves performance. I'm also not sure what that would look like, but perhaps there is a rough heuristic for how many regions we might expect for a given number of tasks?

src/rpc/client.rs
nrc: It is very cheap, it just copies the pointer (yay for ownership!) as long as the capacity of `tasks` is a power of two (which it should be if it has not been created using `with_capacity`.

src/transaction/transaction.rs
nrc: Sure, we can move it to the struct definition.

src/raw/client.rs
Hoverbear: Hm, okay, let's not worry about it.

nrc:
Comment addressed in last commit.

PR #79, opened 2019-07-03
Some more refactoring

nrc:
This is mostly just low level stuff. The most interesting change is removing one of the two timestamp types.

PTAL @Hoverbear @brson


brson:
Does it matter that the removed timestamp type was only 64-bits, while the one replacing it is 128?

nrc:
Hard to tell because the one I removed wasn't really used anywhere. I can't imagine that 8bytes is going to make a big difference to performance here, so I figured more bits is better than less bits.

PR #81, opened 2019-07-12
Implement a unit test by mocking lower level clients

nrc:
This needed a lot of refactoring. I think that improves things though.


nrc:
PTAL @Hoverbear @brson

brson:
And I'm guessing there is not true mocking here right? The RetryClient remains the production PdClient?

brson:
All the `pub (in foo)` annotations give me a bad smell. Better than pure pub everywhere, but it indicates disorganization to me. Better for everything to be daggy, such that mods could be crates later. Not something to fix here though.

nrc:
>  Is PdClient the only thing that needs to be mocked to enable unit testing? 

No, it is just the first step, I suspect we will need to be able to mock each layer of client to test the layer above.

>  It seems like the KvClient / TikvClient (don't know the difference) could be usefully mocked as well

Yes, for sure.

> And I'm guessing there is not true mocking here right? The RetryClient remains the production PdClient?

I'm not sure what you mean. Yes, the `RetryClient` is the production client and `MockPdClient` is the mock client.

> All the pub (in foo) annotations give me a bad smell

Yeah, I'm not a fan. I'll try and get rid of them in the future...

brson:
> I'm not sure what you mean. Yes, the `RetryClient` is the production client and `MockPdClient` is the mock client.

Oh, I must have missed `MockPdClient` behind one of the big diff load links.

Issue #82, opened 2019-07-16
cargo doc fails due to too deep recursion


sticnarf:
`cargo doc` produces errors like this:

```
error[E0275]: overflow evaluating the requirement `alloc::raw_vec::RawVec<(syn::Lifetime, syn::token::Add)>: std::marker::Unpin`
  |
  = help: consider adding a `#![recursion_limit="128"]` attribute to your crate
  = note: required because it appears within the type `std::vec::Vec<(syn::Lifetime, syn::token::Add)>`
  = note: required because it appears within the type `syn::punctuated::Punctuated<syn::Lifetime, syn::token::Add>`
  = note: required because it appears within the type `syn::LifetimeDef`
  = note: required because it appears within the type `(syn::LifetimeDef, syn::token::Comma)`
  = note: required because it appears within the type `*const (syn::LifetimeDef, syn::token::Comma)`
  = note: required because it appears within the type `std::ptr::Unique<(syn::LifetimeDef, syn::token::Comma)>`
  = note: required because it appears within the type `alloc::raw_vec::RawVec<(syn::LifetimeDef, syn::token::Comma)>`
  = note: required because it appears within the type `std::vec::Vec<(syn::LifetimeDef, syn::token::Comma)>`
  = note: required because it appears within the type `syn::punctuated::Punctuated<syn::LifetimeDef, syn::token::Comma>`
  = note: required because it appears within the type `syn::BoundLifetimes`
  = note: required because it appears within the type `std::option::Option<syn::BoundLifetimes>`
  = note: required because it appears within the type `syn::PredicateType`
  = note: required because it appears within the type `syn::WherePredicate`
  = note: required because it appears within the type `(syn::WherePredicate, syn::token::Comma)`
  = note: required because it appears within the type `*const (syn::WherePredicate, syn::token::Comma)`
  = note: required because it appears within the type `std::ptr::Unique<(syn::WherePredicate, syn::token::Comma)>`
  = note: required because it appears within the type `alloc::raw_vec::RawVec<(syn::WherePredicate, syn::token::Comma)>`
  = note: required because it appears within the type `std::vec::Vec<(syn::WherePredicate, syn::token::Comma)>`
  = note: required because it appears within the type `syn::punctuated::Punctuated<syn::WherePredicate, syn::token::Comma>`
  = note: required because it appears within the type `syn::WhereClause`
  = note: required because it appears within the type `std::option::Option<syn::WhereClause>`
[...]

PR #83, opened 2019-07-16
Upgrade rustc version to solve cargo doc failure

sticnarf:
Signed-off-by: Yilin Chen <sticnarf@gmail.com>

Fixes https://github.com/tikv/client-rust/issues/82

PTAL @nrc


nrc:
PTAL @Hoverbear

PR #87, opened 2019-07-29
Refactor into a request-style API

nrc:
PTAL @Hoverbear @sticnarf 

The interesting bit is the last commit really. The idea is that rather than a method for each request in every layer, the clients are fairly light weight, and there is a 'command' object (an implementation of the `RawRequest` trait which does the request-specific stuff. I think this should make things a lot simpler, allow us to share more code, and mock out the pd and tikv clients more easily.

What do you think?


nrc:
PTAL @Hoverbear and @sticnarf It's probably best to start by looking at https://github.com/tikv/client-rust/pull/87/files#diff-0cf04669ac46a0e0b8f4f7679145b2d2 then perhaps review each commit. Sorry for the huge PR. I believe each commit builds, but things may have got a little bit messed up when rebasing.

src/lib.rs
```diff
@@ -102,3 +109,7 @@ pub use crate::errors::ErrorKind;
 pub use crate::errors::Result;
 #[doc(inline)]
 pub use crate::kv::{BoundRange, Key, KvPair, ToOwnedRange, Value};
+#[doc(inline)]
+pub use crate::raw::{Client as RawClient, ColumnFamily};
```
Hoverbear: Is this move purely for discovery purposes? I think it's logical. :)

nrc: It's an effort to have a simpler module structure with fewer re-exports and fewer deep paths. I don't think it is optimal yet, but you should be able to use the client for the most common tasks with only the top-level items.

src/transaction/mod.rs
```diff
@@ -10,20 +10,363 @@
...
+    pub logical: i64,
+}
 
 pub enum Mutation {
     Put(Key, Value),
     Del(Key),
     Lock(Key),
     Rollback(Key),
 }
+
```
sticnarf: I feel good about moving them to the root of the transaction mod. But there are already some modifications. You have to resolve the conflicts. Really sorry.

src/transaction/client.rs
```diff
@@ -12,7 +12,7 @@ use std::sync::Arc;
 
 /// The TiKV transactional `Client` is used to issue requests to the TiKV server and PD cluster.
 pub struct Client {
-    rpc: Arc<RpcClient>,
+    rpc: Arc<PdRpcClient>,
```
sticnarf: The field name should be changed too.

src/kv/bound_range.rs
```diff
@@ -182,6 +183,16 @@ impl<T: Into<Key> + Eq> TryFrom<(Bound<T>, Bound<T>)> for BoundRange {
     }
 }
 
+impl Into<kvrpcpb::KeyRange> for BoundRange {
+    fn into(self) -> kvrpcpb::KeyRange {
+        let (start, end) = self.into_keys();
+        let mut range = kvrpcpb::KeyRange::default();
+        range.set_start_key(start.into());
+        end.map(|k| range.set_end_key(k.into())).unwrap();
```
sticnarf: What if `end` is `None`?

src/kv_client/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+use std::time::Duration;
+
+/// A trait for connecting to TiKV stores.
+pub trait KvConnect: Sized {
+    type KvClient: KvClient + Clone + Send + Sync + 'static;
+
+    fn connect(&self, address: &str) -> Result<Self::KvClient>;
+}
+
+pub type RpcFnType<Req, Resp> = &'static dyn for<'a, 'b> Fn(
```
sticnarf: What about change it to a function pointer?
```
pub type RpcFnType<Req, Resp> = for<'a, 'b> fn(
    &'a TikvClient,
    &'b Req,
    CallOption,
) -> std::result::Result<
    ::grpcio::ClientUnaryReceiver<Resp>,
    ::grpcio::Error,
>
```

src/kv_client/client.rs
```diff
@@ -0,0 +1,335 @@
...
+                .kv_delete_range_async_opt(&req, self.call_options()),
+        )
+    }
+
+    fn call_options(&self) -> CallOption {
+        CallOption::default().timeout(self.timeout)
+    }
+}
+
+fn map_errors<Resp, RpcFuture>(
```
sticnarf: To be honest, I think the way we call this function is not intuitively clear. And the request name is used for stats instead of errors, which is misleading.
```
map_errors("request_name", grpc_result_of_a_future)
```

I think it would better if we can write in some way like this:
```
grpc_result_of_a_future.map_errors().with_stats(request_name)
```

But then perhaps we need a util trait which has no exact meaning. What do you think? @nrc

src/raw/requests.rs
```diff
@@ -0,0 +1,587 @@
...
+    pub cf: Option<ColumnFamily>,
+}
+
+impl RawRequest for RawGet {
+    type Result = Option<Value>;
+    type RpcRequest = kvrpcpb::RawGetRequest;
+    type RpcResponse = kvrpcpb::RawGetResponse;
+    type KeyType = Key;
+    const REQUEST_NAME: &'static str = "raw_get";
+    const RPC_FN: RpcFnType<Self::RpcRequest, Self::RpcResponse> = &TikvClient::raw_get_async_opt;
```
sticnarf: What's your consideration of using an associated const instead of a trait function?

src/pd/client.rs
```diff
@@ -0,0 +1,341 @@
...
+    ) -> BoxFuture<'static, Result<Store<Self::KvClient>>> {
+        self.region_for_id(id)
+            .and_then(move |region| self.clone().map_region_to_store(region).boxed())
+            .boxed()
+    }
+
+    fn group_keys_by_region<K: AsRef<Key> + Send + Sync + 'static>(
+        self: Arc<Self>,
+        keys: impl Iterator<Item = K> + Send + Sync + 'static,
+    ) -> BoxStream<'static, Result<(RegionId, Vec<K>)>> {
```
sticnarf: Stream may not be the best idea. I think we can get them concurrently.

src/pd/client.rs
```diff
@@ -0,0 +1,341 @@
...
+            }
+        })
+        .boxed()
+    }
+
+    // Returns a Steam which iterates over the contexts for each region covered by range.
+    fn stores_for_range(
+        self: Arc<Self>,
+        range: BoundRange,
+    ) -> BoxStream<'static, Result<Store<Self::KvClient>>> {
```
sticnarf: ditto

src/raw/requests.rs
```diff
@@ -0,0 +1,587 @@
...
+        store: &Store<KvC>,
+    ) -> Self::RpcRequest {
+        let mut req = store.request::<Self::RpcRequest>();
+        req.set_key(key.into());
+        req.maybe_set_cf(self.cf);
+
+        req
+    }
+
+    fn store_stream<PdC: PdClient>(
```
sticnarf: A stream prevents us sending requests to multiple stores concurrently...

src/raw/requests.rs
nrc: It would just be a redirection, so it seems like unnecessary boilerplate.

src/pd/client.rs
nrc: I don't understand what you mean exactly, streams can be handled concurrently

src/transaction/mod.rs
nrc: no problem!

src/kv_client/client.rs
nrc: I can rename it, but using a trait is difficult because the transactonal API isn't using `BoxFuture` and `impl Trait` is not supported in traits. Once we are done moving the transactional API to async/await (and the command-oriented design) then we can use a trait.

src/pd/client.rs
sticnarf: I mean in a single request, if we have a stream of stores, then we will send requests in a stream. I expect that we can send all requests concurrently.
Am I misunderstanding how a `Stream` works?

nrc: It depends exactly what you mean. We can execute requests concurrently using  [for_each_concurrent](https://docs.rs/futures-preview/0.3.0-alpha.17/futures/stream/trait.StreamExt.html#method.for_each_concurrent). That doesn't support using different threads, but I think that is an issue for `for_each_concurrent`, not streams in general. That means we can write our own Stream processor which uses threads (although I expect there will be one available soon, if not already).

sticnarf: Cool. I didn't notice this function. I think it should satisfy our needs.

nrc:
Seems the CI failure is not due to this PR. We pass on Mac and fail on Linux with `Cannot add PPA: 'ppa:~maarten-fonville/ubuntu/protobuf'.` which seems like a setup issue. I tried restarting the tests a few times, so I don't think it is a transient error.

Hoverbear:
@nrc Thanks, I'll fix this.

Hoverbear:
@nrc Last CI fail is due to this:

image


Please run `rustfmt` :)

examples/raw.rs
```diff
@@ -6,7 +6,7 @@
 mod common;
 
 use crate::common::parse_args;
-use tikv_client::{raw::Client, Config, Key, KvPair, Result, ToOwnedRange, Value};
+use tikv_client::{Config, Key, KvPair, RawClient as Client, Result, ToOwnedRange, Value};
```
Hoverbear: I gotta say, I really don't think this is better than `raw::Client`. :(

nrc: Yeah, it might be reasonable to change it back. I felt it was a bit odd making people use a module for just a couple of things, but it is also sub-optimal to have RawClient and TransactoinClient when you should only use one.

Hoverbear: When you should **definitely** only use one.

Hoverbear:
Agreed. :)

PR #90, opened 2019-07-31
[DNM] Correct transaction interfaces

sticnarf:
1. Remove invalid interfaces such as
- rollback: an optimistic transaction doesn't need it
- set_isolation_level: we don't support various isolation levels now (pessimistic transaction, which we will possibly support later, assumes RR but it should not be configurable either)
2. Use async/await
3. Fix some documents. Operations like set and delete should not be a `Future` which needs to be resolved later. They are just buffered in memory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ os:
   # - windows # TODO: https://github.com/pingcap/kvproto/issues/355
   - osx
 rust:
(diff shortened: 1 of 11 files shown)


sticnarf:
Previously, CI failed because a clippy bug. https://github.com/rust-lang/rust-clippy/pull/4266 fixed it. Now, I have upgraded toolchain to latest.

PR #91, opened 2019-07-31
Use async/await in the transaction APIs

sticnarf:
After this commit, the library requires a nightly Rust compiler to build.

The required toolchain version is also updated in README.md and .travis.yml.


README.md
```diff
@@ -28,8 +28,6 @@ The client requires a Git dependency until we can [publish it](https://github.co
 
 There are [examples](examples) which show how to use the client in a Rust program.
 
-The examples and documentation use async/await syntax. This is a new feature in Rust and is currently unstable. To use async/await you'll need to add the feature flag `#![async_await]` to your crate and use a nightly compiler (see below).
```
nrc: Could you alter this comment rather than delete it please? I think it is important for users to know

src/transaction/client.rs
```diff
@@ -92,11 +96,11 @@ impl Client {
...
     /// let connect = Client::connect(Config::default());
     /// let client = connect.await.unwrap();
-    /// let timestamp = client.current_timestamp();
+    /// let timestamp = client.current_timestamp().await.unwrap();
     /// # });
     /// ```
-    pub fn current_timestamp(&self) -> Timestamp {
-        unimplemented!()
+    pub async fn current_timestamp(&self) -> Result<Timestamp> {
+        Arc::clone(&self.rpc).get_timestamp().await
```
nrc: prefer `self.rpc.clone()`

sticnarf: I'm OK with both, but the official doc says "The Arc::clone(&from) syntax is the most idiomatic": https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#cloning-references

nrc: In practice, I don't see the explicit form unless it is necessary for disambiguation (I had a quick search though the Rust repo and although it is hard to tell where `.clone` is used on an `Arc`, it does seem to be used more often than the explicit version).

sticnarf:
Previously, CI failed because of a clippy bug. rust-lang/rust-clippy#4266 fixed it. Now, I have upgraded toolchain to latest.

PR #96, opened 2019-08-06
Correct transaction api

sticnarf:
1. Remove invalid interfaces such as
- rollback: an optimistic transaction doesn't need it
- set_isolation_level: we don't support various isolation levels now (pessimistic transaction, which we will possibly support later, assumes RR but it should not be configurable either)
2. Fix some documents. Operations like set and delete should not be a `Future` which needs to be resolved later. They are just buffered in memory.


MyonKeminta:
Could you please explain why rollback is unnecessary?

sticnarf:
> 
> 
> Could you please explain why rollback is unnecessary?

In the user's perspective, dropping the `Transaction` instance means rollback. Internal rollback operation will be added later.

PR #97, opened 2019-08-07
Record mutations in transactions

sticnarf:
A `BTreeMap` (prior to `HashMap` because scan needs supporting) is added to record all key mutations. `Mutation` inside the library now doesn't contain the key. The `Mutation`s used in RPC is generated in prewrite.

Mutations in a transaction should be queried first on read operations. Although `Snapshot` is not implemented yet, we can return values in the mutation map now.


src/transaction/transaction.rs
```diff
@@ -48,8 +51,13 @@ impl Transaction {
     /// txn.commit().await.unwrap();
     /// # });
     /// ```
-    pub async fn get(&self, _key: impl Into<Key>) -> Result<Value> {
-        unimplemented!()
+    pub async fn get(&self, key: impl Into<Key>) -> Result<Value> {
+        let key = key.into();
+        if let Some(value) = self.get_from_mutations(&key) {
```
nrc: The raw API returns `Result<Option<Value>>` and I think the transactional one should too. That means `get_value` will need to be changed to handle deleted values properly.

src/transaction/transaction.rs
```diff
@@ -71,9 +79,38 @@ impl Transaction {
...
     ) -> Result<Vec<KvPair>> {
-        unimplemented!()
+        let mut result = Vec::new();
+        let mut keys_from_snapshot = Vec::new();
+        let mut result_indices_from_snapshot = Vec::new();
+
+        // Try to fill the result vector from mutations
+        for key in keys {
+            let key = key.into();
+            if let Some(value) = self.get_from_mutations(&key) {
```
nrc: nit: prefer `match` to `if let ... else`

src/transaction/mod.rs
```diff
@@ -14,16 +14,45 @@ pub use self::requests::Scanner;
...
-    Del(Key),
-    Lock(Key),
-    Rollback(Key),
+    Put(Value),
+    Del,
+    Lock,
+}
+
+impl Mutation {
+    fn with_key(self, key: impl Into<Key>) -> kvrpcpb::Mutation {
```
nrc: I think this will only ever be called with `Key`s so you can skip the `impl Into`.

src/transaction/transaction.rs
```diff
@@ -71,9 +79,38 @@ impl Transaction {
...
+                result.push((key, value).into());
+            } else {
+                keys_from_snapshot.push(key);
+                result_indices_from_snapshot.push(result.len());
+                // Push a placeholder
+                result.push(KvPair::default());
```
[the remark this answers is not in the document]
sticnarf: Finally we have to return a `Vec<KvPair>`. If we use `Option`, we need one more `Vec`.

src/transaction/mod.rs
```diff
@@ -14,16 +14,45 @@ pub use self::requests::Scanner;
...
+
 use crate::{Key, Value};
 
 mod client;
 mod requests;
 #[allow(clippy::module_inception)]
 mod transaction;
 
+#[derive(Debug, Clone)]
 pub enum Mutation {
```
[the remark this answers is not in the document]
sticnarf: No. I'll remove it.

src/transaction/transaction.rs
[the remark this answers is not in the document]
nrc: I think that we should be returning a pair of a Key and Option<Value>, and that that would be fine to use here too. There might be opportunity later to do some interesting optimisations here (if we end up using streams, or with whatever the type signatures end up looking like).

Anyway, I think it is better to use an `Option` for now since that is the 'right' thing to do, using a null value is premature optimisation, IMO. We can optimise later.

src/transaction/transaction.rs
```diff
@@ -71,9 +79,38 @@ impl Transaction {
     /// ```
     pub async fn batch_get(
         &self,
-        _keys: impl IntoIterator<Item = impl Into<Key>>,
+        keys: impl IntoIterator<Item = impl Into<Key>>,
     ) -> Result<Vec<KvPair>> {
-        unimplemented!()
+        let mut result = Vec::new();
+        let mut keys_from_snapshot = Vec::new();
+        let mut result_indices_from_snapshot = Vec::new();
```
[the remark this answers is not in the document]
sticnarf: Tt's not convenient to use a `Vec<(KvPair, usize)>` now. If we want to avoid cloning the key, we have to consume the `Vec` to get an iterator. Meanwhile, the indices are also consumed but we want to use them sometime later.

src/transaction/transaction.rs
```diff
@@ -71,9 +79,38 @@ impl Transaction {
...
+
+        // Try to fill the result vector from mutations
+        for key in keys {
```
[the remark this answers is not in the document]
sticnarf: Sorry I don't get what you mean...

sticnarf:
Now, batch_get API returns a HashMap. I hope this will make it easier for both users and us. Unit tests for the added part of this PR is missing. However, I want to refactor the kv api to be request based first and make kv client a trait.
PTAL @nrc @youjiali1995

src/transaction/transaction.rs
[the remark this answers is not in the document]
nrc: Don't worry, I think this comment is not true any more

nrc:
> Now, batch_get API returns a HashMap. I hope this will make it easier for both users and us

Do you have any examples of how real users use `batch_get`? Another possibility is to return an iterator, I'm not sure which would be more ergonomic.

sticnarf:
> > Now, batch_get API returns a HashMap. I hope this will make it easier for both users and us
> 
> Do you have any examples of how real users use `batch_get`? Another possibility is to return an iterator, I'm not sure which would be more ergonomic.

BatchGet in the go client of TiKV returns a map: https://github.com/tikv/client-go/blob/master/txnkv/txn.go#L81

I have little idea how real users use `batch_get`. The only examples I can find is TiDB: https://github.com/pingcap/tidb/blob/master/ddl/index.go#L767 and titan in Meitu: https://github.com/distributedio/titan/blob/3af3679f4ed5d1ec6cc0527ec17243f010539faf/db/db.go#L87

I think it's also fine to return an Iterator of (Key, Option<Value>). Do you think the type will be too complex here?

sticnarf:
I've changed the return type of batch_get to an iterator of `(Key, Option<Value>)`. And I also add a unit test for it.
@nrc @youjiali1995 PTAL again! Thanks.

src/transaction/mod.rs
```diff
@@ -24,11 +26,43 @@ pub struct Timestamp {
...
+                pb.set_op(kvrpcpb::Op::Put);
+                pb.set_value(v.into());
+            }
+            Mutation::Del => pb.set_op(kvrpcpb::Op::Del),
+            Mutation::Lock => pb.set_op(kvrpcpb::Op::Lock),
+        };
+        pb
+    }
+
+    /// Returns a `Some` if the value can be determined by this mutation. Otherwise, returns `None`.
```
mapleFU: Maybe `Determined` and `Undetermined` here?

sticnarf: Oh, I forgot to modify it.

sticnarf: I think the type itself is clear enough so let's just remove this comment.

src/transaction/mod.rs
```diff
@@ -24,11 +26,42 @@ pub struct Timestamp {
...
-    Rollback(Key),
+#[derive(Debug, Clone)]
+enum Mutation {
+    Put(Value),
+    Del,
+    Lock,
+}
+
+impl Mutation {
+    fn with_key(self, key: Key) -> kvrpcpb::Mutation {
```
nrc: I would rename this `into_proto` or `into_proto_with_key` (`with_*` is the idiomatic name for a method which takes a function to operate on an internal value).

src/transaction/mod.rs
```diff
@@ -53,52 +86,91 @@ pub enum Mutation {
...
+            .map(|key| {
+                let key = key.into();
+                let mutation_value = self.get_from_mutations(&key);
+                if let MutationValue::Undetermined = mutation_value {
+                    keys_from_snapshot.push(key.clone());
+                }
+                (key, mutation_value)
+            })
+            .collect::<Vec<_>>()
+            .into_iter();
```
nrc: There should be a better way than to `collect` then immediately iterate again. Maybe `cloned()` or add a `to_owned` to the map function? If you're doing this for the side effects, then please add a comment

src/transaction/mod.rs
```diff
@@ -53,52 +86,91 @@ pub enum Mutation {
...
+            .await?
+            .peekable();
+        Ok(std::iter::from_fn(move || {
+            let (key, mutation_value) = results_in_buffer.next()?;
+            match mutation_value {
+                MutationValue::Determined(value) => Some((key, value)),
+                MutationValue::Undetermined => match results_from_snapshot.peek() {
+                    Some((key_from_snapshot, _)) if &key == key_from_snapshot => {
+                        results_from_snapshot.next()
+                    }
```
nrc: I think it would be good to add a comment inside this method somewhere to explain what you're doing.

src/transaction/mod.rs
sticnarf: It's just an eager evaluation of the iterator. Then, the new iterator does not hold a reference to `self`.

sticnarf: Yes, and there is side effect too. Now I think it's better to use a for loop...

src/transaction/mod.rs
sticnarf: Sorry, I find I am doing wrong here. These logic should belong to the batch_get in snapshot instead of here.

sticnarf:
@nrc @youjiali1995  PTAL again.

PR #98, opened 2019-08-14
Request based MVCC APIs

sticnarf:
The original txn requests are removed. Instead, I want to make all requests use the request based style.
There is little difference between raw api and mvcc apis (except that mvcc apis need to resolve lock), so I combine their request traits to `KvRequest`.

- [x] merge #97 first


src/kv_client/request.rs
```diff
@@ -0,0 +1,105 @@
...
+use std::sync::Arc;
+
+pub trait KvRequest: Sync + Send + 'static + Sized + Clone {
+    type Result;
+    type RpcRequest;
+    type RpcResponse: HasError + Clone + Send + 'static;
+    /// A single `KvRequest` can be divided into a number of RPC requests because the keys span
+    /// several regions or a single RPC request is too large. Most of the fields in these requests
+    /// share the same content while `Payload` is the part which differs among the requests.
+    type Payload;
```
nrc: I don't think `Payload` is a great name here - it is very generic and to me feels more like a value-like thing than a key-like thing. It is maybe better than `KeyType`, but I think we could still do better.

src/kv_client/request.rs
```diff
@@ -0,0 +1,105 @@
...
+};
+
+use futures::future::BoxFuture;
+use futures::prelude::*;
+use futures::stream::BoxStream;
+use grpcio::CallOption;
+use kvproto::kvrpcpb;
+use std::sync::Arc;
+
+pub trait KvRequest: Sync + Send + 'static + Sized + Clone {
```
nrc: Do we still need the `clone` bound?

src/kv_client/request.rs
```diff
@@ -0,0 +1,105 @@
...
+                .boxed(),
+        )
+    }
+
+    fn store_stream<PdC: PdClient>(
+        &mut self,
+        pd_client: Arc<PdC>,
+    ) -> BoxStream<'static, Result<(Self::Payload, Store<PdC::KvClient>)>>;
+
+    fn into_request<KvC: KvClient>(
```
nrc: This should not be `into_` if it takes `self` by reference. Maybe `make_rpc_request`?

src/kv_client/request.rs
sticnarf: Seems not. Let me remove it.

src/kv_client/request.rs
sticnarf: Sounds good.

src/kv_client/request.rs
sticnarf: I change it to `Entries`. Any better suggestions?

nrc: What are the possible instantiations? Key, Vec<Key>, BoundRange, and Vec<BoundRange>, are there any more?

sticnarf: `Vec<kvrpcpb::Mutations>`

nrc: Operand? KeyData? InputData? Data? (in decreasing order of preference)

sticnarf: It's changed to `KeyData`.

mapleFU:
I have a naive question: maybe `MockDispatch` and `self.mock_dispatch` should be in `#[cfg(test)]`, or you should just use `Dispatch` trait.

sticnarf:
> I have a naive question: maybe `MockDispatch` and `self.mock_dispatch` should be in `#[cfg(test)]`, or you should just use `Dispatch` trait.

Just because we are reusing `KvRequest::execute` function.

In fact, I don't think `MockDispatch` is very useful. To test the `execute` function, we can just create a mock request and use `MockKvClient`. Other trait functions in `KvRequest` should be tested separately. The whole process should be tested by an integration test. Then we don't need  `MockDispatch`. @nrc

nrc:
MockDispatch is a trait which lets the mocks hook in a dummy dispatch before doing a proper dispatch. I've renamed it in https://github.com/tikv/client-rust/pull/101. The impls need to be cfg(test), but the trait itself shouldn't be.

I think it is useful, it lets us unit test the high-level parts of a request without the complications of rpc, network errors, retrying requests, etc.

PR #100, opened 2019-08-15
Implement get and batch_get for snapshot

sticnarf:
- [ ]  merge #98 first

Integration tests are missing because we don't have have txn write support now.

Unit tests are missing because we need a mock tikv with mvcc and cf support.


sticnarf:
covered by #104

PR #101, opened 2019-08-20
Add some more tests

nrc:
And do a little more mocking.

PTAL @sticnarf @Hoverbear


src/mock.rs
```diff
@@ -0,0 +1,131 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```
Hoverbear: This could use mod level docs.

src/pd/client.rs
```diff
@@ -204,116 +216,24 @@ impl<KvC: KvConnect + Send + Sync + 'static> PdRpcClient<KvC> {
...
-            let result = match id {
-                1 => Ok(Self::region1()),
-                2 => Ok(Self::region2()),
-                _ => Err(Error::region_not_found(id, None)),
-            };
 
-            Box::pin(ready(result))
-        }
+    #[test]
+    fn test_kv_client() {
```
Hoverbear: Maybe this could have a more specific name?

nrc:
@Hoverbear addressed comments

I've also done some more refactoring.

PR #103, opened 2019-08-20
Add error handling to KvRequest::execute

sticnarf:
The main part of this PR is removing custom request types and implementing `KvRequest` directly for RPC request types.

The reason comes from how we handle region errors. Region errors mean the `Store` associated with the request is wrong and we need to get the right `Store` from PD and resend the request. As region information may change, the failed request may need spliting to smaller parts. So, it means failed requests have to complete the whole execution process of an RPC request.

Currently, the execution process can be applied to our custom request types. But when an RPC request is made, we only have gRPC request types. One choice is to add a new trait function converting an RPC request type back to the custom request type. But it is too verbose. I think it's OK to implement `KvRequest` directly for gRPC request types, although it makes the meaning of gRPC request types not so clear.

Currently, we always retry immediately on region errors. Backoff and retry limit are planned to be added later.


src/raw/client.rs
```diff
@@ -325,14 +337,50 @@ impl Client {
...
+    fn set_cf(&mut self, cf: String);
+
+    fn maybe_set_cf(&mut self, cf: Option<ColumnFamily>) {
+        if let Some(cf) = cf {
+            self.set_cf(cf.to_string());
+        }
+    }
 }
+
+macro_rules! impl_raw_rpc_request {
```
nrc: Can we put this macro and the uses in raw/requests.rs instead of here?

src/raw/requests.rs
```diff
@@ -14,27 +13,17 @@ use kvproto::tikvpb::TikvClient;
...
-
-impl KvRequest for RawGet {
+impl KvRequest for kvrpcpb::RawGetRequest {
     type Result = Option<Value>;
-    type RpcRequest = kvrpcpb::RawGetRequest;
     type RpcResponse = kvrpcpb::RawGetResponse;
     type KeyData = Key;
     const REQUEST_NAME: &'static str = "raw_get";
-    const RPC_FN: RpcFnType<Self::RpcRequest, Self::RpcResponse> = TikvClient::raw_get_async_opt;
+    const RPC_FN: RpcFnType<Self, Self::RpcResponse> = TikvClient::raw_get_async_opt;
```
nrc: I think it would be good to have ctor functions in this module for each request so that the code in the client can be clearer, e,g,, `pub fn new_get_request(key: impl Into<Key>, ...) -> kvrpcpb::RawGetRequest`

src/raw/requests.rs
```diff
@@ -14,27 +13,17 @@ use kvproto::tikvpb::TikvClient;
...
 
-    fn make_rpc_request<KvC: KvClient>(
-        &self,
-        key: Self::KeyData,
-        store: &Store<KvC>,
-    ) -> Self::RpcRequest {
-        let mut req = store.request::<Self::RpcRequest>();
+    fn make_rpc_request<KvC: KvClient>(&self, key: Self::KeyData, store: &Store<KvC>) -> Self {
```
nrc: We don't need this method any more, we can have a `set_store` method instead

src/raw/requests.rs
```diff
@@ -100,12 +78,11 @@ impl KvRequest for RawBatchGet {
...
         pd_client: Arc<PdC>,
     ) -> BoxStream<'static, Result<(Self::KeyData, Store<PdC::KvClient>)>> {
-        let mut keys = Vec::new();
-        mem::swap(&mut keys, &mut self.keys);
+        let keys = mem::replace(&mut self.keys, Default::default());
 
         pd_client
             .clone()
-            .group_keys_by_region(keys.into_iter())
+            .group_keys_by_region(keys.into_iter().map(Into::into))
```
nrc: Why is this needed? Can we change `group_keys_by_region`so it is not necessary?

src/raw/requests.rs
sticnarf: seems good

src/raw/requests.rs
sticnarf: We still need to know how to set `KeyData` because they can be split to smaller parts. And it's also not so convenient to fill other option fields with reused code (for example, CF is a String which is not Copy).

src/raw/requests.rs
sticnarf: I've tried and think we shouldn't. 
```
fn group_keys_by_region<K: AsRef<Key> + Send + Sync + 'static>(
        self: Arc<Self>,
        keys: impl Iterator<Item = K> + Send + Sync + 'static,
    )
```
If we introduce a second generic parameter (Into<K>), for most cases, the compiler will fail to  compute what type it is. However, if the type is `impl Into<Key>` here, it will succeed (because there will be no other possibility).

src/transaction/requests.rs
```diff
@@ -43,9 +32,9 @@ impl KvRequest for MvccGet {
         &mut self,
         pd_client: Arc<PdC>,
     ) -> BoxStream<'static, Result<(Self::KeyData, Store<PdC::KvClient>)>> {
-        let key = self.key.clone();
+        let key = mem::replace(&mut self.key, Vec::default()).into();
```
AndreMouche: Why we consume the `self.key`? Would it be used again later?

sticnarf: Never mind, the one in the original request will not be used again.

src/raw/requests.rs
nrc: I don't understand - I thought we set the key data and family when we create the request, and so here all we need to do is set the store/context (and maybe clone).

sticnarf: `make_rpc_request` is now used in the `response_stream` function when a small part of key data is generated. So we must know how to set key data at least.

PR #104, opened 2019-08-21
Use the get and batch get requests in the transaction client, plus some refactoring.

nrc:
~~Based on https://github.com/tikv/client-rust/pull/101 (probably best to land that before reviewing this).~~


src/lib.rs
```diff
@@ -5,6 +5,7 @@
 #![allow(clippy::redundant_closure)]
 #![allow(clippy::type_complexity)]
 #![feature(async_await)]
+#![feature(async_closure)]
```
sticnarf: It won't be stablized in the near future, will it? Let's avoid it first.

src/transaction/transaction.rs
```diff
@@ -28,9 +37,104 @@ use std::{collections::BTreeMap, ops::RangeBounds};
 /// ```
 #[derive(new)]
 pub struct Transaction {
-    pub snapshot: Snapshot,
+    pub timestamp: Timestamp,
```
sticnarf: shoudn't be pub

src/transaction/transaction.rs
```diff
@@ -28,9 +37,104 @@ use std::{collections::BTreeMap, ops::RangeBounds};
...
+            MutationValue::Undetermined => {
+                let value = f(key.clone()).await?;
+                let mut mutations = self.mutations.lock().unwrap();
+                mutations.insert(key, Mutation::Cached(value.clone()));
+                Ok(value)
+            }
+        }
+    }
+
+    async fn batch_get_or_else<F, Fut>(
```
sticnarf: I hope the order can be retained if we . You may refer to my pr: https://github.com/tikv/client-rust/pull/100/files#diff-2dca5666975e7d44fe0f80bcd94c551aR412

src/transaction/transaction.rs
```diff
@@ -28,9 +37,104 @@ use std::{collections::BTreeMap, ops::RangeBounds};
...
+            .map(Mutation::get_value)
+            .unwrap_or(MutationValue::Undetermined)
+    }
+
+    fn lock(&self, key: Key) {
+        let mut mutations = self.mutations.lock().unwrap();
+        let value = mutations
+            .entry(key)
+            // Mutated keys don't need a lock.
+            .or_insert(Mutation::Lock);
```
sticnarf: Cached results need locks

src/lib.rs
nrc: It is kind of hard to avoid :-( I think lets wait until async/await is stable and then remove it if async_closure is unlikely to stabilise soon.

src/transaction/transaction.rs
nrc: I don't think it is necessary - the result should be treated like a map, not an ordered list. If we had a `Map` type in Rust I would use it

src/transaction/transaction.rs
nrc: :+1:

src/transaction/transaction.rs
sticnarf: So why not a `HashMap`?

src/lib.rs
Hoverbear: I worry about this change as well. :(

src/transaction/mod.rs
```diff
@@ -29,40 +28,64 @@ pub struct Timestamp {
...
         match self {
+            Mutation::Cached(value) => MutationValue::Determined(value.clone()),
             Mutation::Put(value) => MutationValue::Determined(Some(value.clone())),
             Mutation::Del => MutationValue::Determined(None),
             Mutation::Lock => MutationValue::Undetermined,
         }
     }
 }
 
+#[derive(Eq, PartialEq, Debug)]
```
Hoverbear: I feel like this struct could use some documentation.

src/transaction/transaction.rs
nrc: By returning the iterator we give the user the choice about whether to use a BTreeMap or HashMap, we'd only collect the iterator into a HashMap anyway. Having said that, I'd be fine with this as a HashMap, tbh.

nrc:
Comments addressed. PTAL @sticnarf

src/transaction/transaction.rs
```diff
@@ -52,10 +62,15 @@ impl Transaction {
...
+                MvccGet {
+                    key,
+                    version: self.timestamp.into_version(),
+                }
+                .execute(self.rpc.clone())
+            })
+            .await
     }
 
     /// Gets the values associated with the given keys. The returned iterator is in the same order
```
sticnarf: Oh, this line should be modified too. We don't keep the order now.

nrc:
Updated to address comments and rebased

PR #106, opened 2019-08-22
Add Snapshot

nrc:
Unlike the previous snapshot, this one wraps the transaction rather than vice versa.

PTAL @sticnarf @Hoverbear

### DIFF
--- a/src/transaction/snapshot.rs
+++ b/src/transaction/snapshot.rs
@@ -1,0 +1,38 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
+
+use crate::{Key, KvPair, Result, Transaction, Value};
+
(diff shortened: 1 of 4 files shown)


src/transaction/snapshot.rs
```diff
@@ -0,0 +1,38 @@
...
+    transaction: Transaction,
+}
+
+impl Snapshot {
+    /// Gets the value associated with the given key.
+    pub async fn get(&self, key: impl Into<Key>) -> Result<Option<Value>> {
+        self.transaction.get(key).await
+    }
+
+    /// Gets the values associated with the given keys. The returned iterator is in the same order
```
sticnarf:
```suggestion
    /// Gets the values associated with the given keys.
```

PR #108, opened 2019-08-25
Resolve locks

sticnarf:
When locks are returned, we need to resolve the locks and try again. Locks are resolved in the unit of tuple (transaction, region). We use `CleanupRequest` to clean the primary lock (if the lock is not committed) or get the commit version. Then, use the commit version to resolve the locks.

In this PR, variants of region errors are removed from the library's error type because the behavior is the same in the library. Also, we shouldn't mix errors from PD with KV region errors. This is fixed too.


src/transaction/lock.rs
```diff
@@ -0,0 +1,106 @@
...
+            },
+        }
+    }
+}
+
+pub trait HasLocks {
+    fn take_locks(&mut self) -> Vec<kvrpcpb::LockInfo>;
+}
+
+macro_rules! dummy_impl_has_locks {
```
nrc: Could you use a default implementation of `take_locks` rather than use this macro?

src/transaction/lock.rs
```diff
@@ -0,0 +1,106 @@
+use super::requests;
+use crate::pd::{PdClient, RegionVerId};
+use crate::request::KvRequest;
+use crate::{ErrorKind, Key, Result, Timestamp};
+
+use kvproto::kvrpcpb;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+
+pub async fn resolve_locks(
```
nrc: It looks to me like these functions should be methods on `PdClient`. Is there a reason they're not?

nrc: Could you document what 'resolve' means

src/transaction/lock.rs
```diff
@@ -0,0 +1,106 @@
...
+    Ok(())
+}
+
+async fn resolve_lock_with_retry(
+    key: Key,
+    start_version: u64,
+    commit_version: u64,
+    pd_client: Arc<impl PdClient>,
+) -> Result<RegionVerId> {
+    // TODO: Add backoff and retry limit
```
nrc: Seems like a retry limit is very easy to add and prevents us getting stuck in infinite loops. Could you add it please?

src/transaction/lock.rs
sticnarf: Seems we can't, unless we use specialization?

src/transaction/lock.rs
sticnarf: > It looks to me like these functions should be methods on `PdClient`. Is there a reason they're not?

The requests are sent to TiKV instead of PD. So I think it should belong to TiKV requests.

sticnarf: > Could you document what 'resolve' means

Ok.

src/transaction/lock.rs
sticnarf: I hoped to add this backoff and retry limit to this lock resolving and normal KV requests together. Anyway, it's fine to simply add a retry limit here.

sticnarf:
@nrc The tools we have now (a stateless mock tikv) is not enough to unit test the `resolve_locks` function. I feel like we defer it a bit.

src/transaction/requests.rs
```diff
@@ -160,3 +179,144 @@ impl KvRequest for kvrpcpb::ScanRequest {
...
+impl HasLocks for kvrpcpb::ScanResponse {
+    fn take_locks(&mut self) -> Vec<kvrpcpb::LockInfo> {
+        self.pairs
+            .iter_mut()
+            .filter_map(|pair| pair.error.as_mut().and_then(|error| error.locked.take()))
+            .collect()
+    }
+}
+
+// For a ResolveLockRequest, it does not contain enough information about which region it should be
```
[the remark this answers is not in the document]
sticnarf: It is for the whole request, describing why`on_region_error` behavior differs from other requests.

src/transaction/lock.rs
```diff
@@ -0,0 +1,152 @@
...
+    Ok(has_live_locks)
+}
+
+async fn resolve_lock_with_retry(
+    key: Key,
+    start_version: u64,
+    commit_version: u64,
+    pd_client: Arc<impl PdClient>,
+) -> Result<RegionVerId> {
+    // TODO: Add backoff
```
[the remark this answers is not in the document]
sticnarf: Of course

src/transaction/requests.rs
```diff
@@ -160,3 +179,144 @@ impl KvRequest for kvrpcpb::ScanRequest {
...
+    }
+
+    fn map_result(_: Self::RpcResponse) -> Self::Result {}
+
+    fn reduce(
+        results: BoxStream<'static, Result<Self::Result>>,
+    ) -> BoxFuture<'static, Result<Self::Result>> {
+        results
+            .into_future()
+            .map(|(f, _)| f.expect("no results should be impossible"))
```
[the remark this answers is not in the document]
sticnarf: This error message is used when there is only one result so "at least one" is not so good. Let's think of a better error message later.

rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2019-07-31
+nightly-2019-08-25
```
AndreMouche: I think we should only do one thing in one PR. Please pay attention to this rule in the future.

src/transaction/requests.rs
```diff
@@ -160,3 +179,141 @@ impl KvRequest for kvrpcpb::ScanRequest {
...
+        req.set_start_version(self.start_version);
+
+        req
+    }
+
+    fn store_stream<PdC: PdClient>(
+        &mut self,
+        pd_client: Arc<PdC>,
+    ) -> BoxStream<'static, Result<(Self::KeyData, Store<PdC::KvClient>)>> {
+        let key = mem::replace(&mut self.key, Default::default()).into();
```
AndreMouche: It seems repeated with L36-L41, could we avoid it?

rust-toolchain
sticnarf: Sorry, just find some dependencies cannot  compile because they remove the feature gate.

src/transaction/requests.rs
sticnarf: Done

src/transaction/lock.rs
nrc: ISTM that this is a PD 'method' since we first query the timestamp, then `resolve_lock_with_retry` is a kv client method, although it is a bit weird that it has to figure out the region from pd (though I see why you'd do it this way). I think that there might be a nice way to factor these functions, at the moment it doesn't quite feel like the abstractions are right.

src/transaction/lock.rs
nrc: Why doesn't

```
pub trait HasLocks {
    fn take_locks(&mut self) -> Vec<kvrpcpb::LockInfo> {
        Vec::new()
    }
}
```

work? It doesn't look like it should need specialisation to me.

(And no, specialisation is a long way from being stable, I think there are some complex outstanding questions around soundness?)

sticnarf: Having that default implementation still requires us to impl HasLocks for the response types one by one. (impl<T: KvRequest> doesn't work)

nrc: But it would mean that you could just have a one line impl for each one, rather than a macro.

sticnarf: Oh, I see what you mean. That makes sense.

PR #109, opened 2019-08-27
Implement 2PC

sticnarf:
- [x] Merge #112 first

This PR implements two-phase commit. It is a very initial implementation but should work well.

After this PR, the transaction part is barely enough to be functional.

For the added part of this PR, there are some todos (will update to issues when the pr gets merged):
- When committing the primary key, also carry some other keys. (split requests by size)
- Increase the lock TTL for big transactions
- Enable transactional API integration test
- Add as many unit tests as possible (to be discussed about how to do this)

### DIFF
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -40,6 +40,7 @@ use std::{fmt, u8};
 /// functions.
 #[derive(Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 #[cfg_attr(test, derive(Arbitrary))]
+#[repr(transparent)]
 pub struct Key(
     #[cfg_attr(
         test,
@@ -114,6 +115,12 @@ impl<'a> Into<&'a [u8]> for &'a Key {
 impl AsRef<Key> for Key {
     fn as_ref(&self) -> &Key {
         self
+    }
+}
+
+impl AsRef<Key> for Vec<u8> {
+    fn as_ref(&self) -> &Key {
+        unsafe { std::mem::transmute(&self) }
     }
 }
 
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -120,7 +120,7 @@ impl AsRef<Key> for Key {
 
 impl AsRef<Key> for Vec<u8> {
     fn as_ref(&self) -> &Key {
-        unsafe { std::mem::transmute(self) }
+        unsafe { &*(self as *const Vec<u8> as *const Key) }
     }
 }
 
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -120,7 +120,7 @@ impl AsRef<Key> for Key {
 
 impl AsRef<Key> for Vec<u8> {
     fn as_ref(&self) -> &Key {
-        unsafe { std::mem::transmute(&self) }
+        unsafe { std::mem::transmute(self) }
     }
 }
 
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -99,6 +99,16 @@ impl Buffer {
         self.mutations.lock().unwrap().insert(key, Mutation::Del);
     }
 
+    /// Converts the buffered mutations to the proto buffer version
+    pub fn to_proto_mutations(&self) -> Vec<kvrpcpb::Mutation> {
+        self.mutations
+            .lock()
+            .unwrap()
+            .iter()
+            .flat_map(|(key, mutation)| mutation.to_proto_with_key(key))
+            .collect()
+    }
+
     fn get_from_mutations(&self, key: &Key) -> MutationValue {
         self.mutations
             .lock()
@@ -111,7 +121,7 @@ impl Buffer {
 
 // The state of a value in the buffer.
 #[derive(Debug, Clone)]
-enum Mutation {
+pub enum Mutation {
     // The value has been read from the server. None means there is no entry.
     Cached(Option<Value>),
     // Value has been written.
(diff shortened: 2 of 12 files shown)


src/transaction/requests.rs
```diff
@@ -294,7 +294,176 @@ pub fn new_cleanup_request(key: impl Into<Key>, start_version: u64) -> kvrpcpb::
     req
 }
 
-dummy_impl_has_locks!(PrewriteResponse);
+impl AsRef<Key> for kvrpcpb::Mutation {
```
nrc: How is this used? I generally don't like this sort of thing - I think it is better to be explicit about conversions rather than using AsRef/Deref, etc.

src/transaction/transaction.rs
```diff
@@ -173,31 +168,130 @@ impl Transaction {
...
+    rpc: Arc<PdRpcClient>,
+    #[new(default)]
+    committed: bool,
+    #[new(default)]
+    undetermined: bool,
+}
+
+impl TwoPhaseCommitter {
+    async fn commit(mut self) -> Result<()> {
+        if self.mutations.is_empty() {
```
nrc: I think we should check if `mutations` is read only, i.e., we can return here if the only values in mutations are cached values.

src/transaction/client.rs
```diff
@@ -24,12 +24,15 @@ impl Client {
...
     /// # futures::executor::block_on(async {
-    /// let connect = TransactionClient::connect(Config::default());
-    /// let client = connect.await.unwrap();
+    /// let client = TransactionClient::new(Config::default()).await.unwrap();
     /// # });
     /// ```
-    pub fn connect(config: Config) -> Connect {
-        Connect::new(config)
+    pub async fn new(config: Config) -> Result<Client> {
+        let bg_worker = ThreadPool::new()?;
```
nrc: This is fine for now, but long term we will want to let the user pass in a thread pool (or callback) of some kind.

sticnarf: Exactly. But there isn't a uniform abstraction of a threadpool in the async ecosystem. Futures has Spawn trait, tokio has Executor trait...

src/transaction/requests.rs
sticnarf: It's for implementing AsRef<Key> for Mutation. Any better idea?

sticnarf: Mutation is used for store_stream_for_keys, which requires it to implement `AsRef<Key>`

nrc: Hmm, that is my fault I guess. Maybe we should use a different trait for that. I would leave it for now, but add a comment.

src/kv/key.rs
```diff
@@ -117,6 +118,12 @@ impl AsRef<Key> for Key {
     }
 }
 
+impl AsRef<Key> for Vec<u8> {
```
nrc: Where is this one used?

src/transaction/requests.rs
sticnarf: Agree. A different trait seems better.

src/kv/key.rs
sticnarf: Here: https://github.com/tikv/client-rust/pull/109/files#diff-d01850033b0d2fc084df1829a7789724R299

nrc: Hmm, ok right. So I think there is a bit of an issue that we have to convert back and forth between our `Key` abstraction and the underlying `Vec<u8>`, in this case because we are using the protobuf `Mutation` rather than our own kind of mutation. Not sure what would be better, but worth thinking about for the future

src/transaction/transaction.rs
AndreMouche: If the mutations are empty, means the transaction is  read-only

nrc: We are storing some cached read values in the mutations table too, so even if it is non-empty, it might still be a read-only transaction

sticnarf: The buffer may contain cached read values. But when the buffer is transformed to the `mutations` here, it should contain mutations only.

src/transaction/buffer.rs
```diff
@@ -99,6 +99,16 @@ impl Buffer {
         self.mutations.lock().unwrap().insert(key, Mutation::Del);
     }
 
+    /// Converts the buffered mutations to the proto buffer version
+    pub fn to_proto_mutations(&self) -> Vec<kvrpcpb::Mutation> {
+        self.mutations
+            .lock()
+            .unwrap()
+            .iter()
+            .flat_map(|(key, mutation)| mutation.to_proto_with_key(key))
```
nrc: I think it is clearer to use `filter_map` rather than `flat_map` here.

src/transaction/transaction.rs
nrc: Oh yeah, so it does. I added a comment there to hopefully make that clearer.

src/transaction/transaction.rs
```diff
@@ -173,31 +168,130 @@ impl Transaction {
...
-    async fn commit_secondary(&mut self) -> Result<()> {
-        unimplemented!()
+    async fn commit_secondary(mut self, commit_version: u64) -> Result<()> {
+        let mutations = mem::replace(&mut self.mutations, Vec::default());
+        let keys = mutations
+            .into_iter()
+            .skip(1) // skip primary key
+            .map(|mutation| mutation.key.into())
+            .collect();
+        new_commit_request(keys, self.start_version, commit_version)
```
sticnarf: Don't do the secondary commit if `keys` is empty.

sticnarf:
The logic in this PR must be interacted with TiKV. I think an integration test is more suitable. If necessary, we can add unit test for specific requests in the future.

mapleFU:
Looks awesome. I want to know whether our rust-client would send cleanup requests now.

Issue #105, opened 2019-08-21
Test `util::tests::test_global_timer` fails intermittently on CI


nrc:
Example output

```
thread 'util::tests::test_global_timer' panicked at 'assertion failed: timer.elapsed() >= ::std::time::Duration::from_millis(100)', src/util.rs:72:9

stack backtrace:

   0: backtrace::backtrace::libunwind::trace

             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88

   1: backtrace::backtrace::trace_unsynchronized

             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66

   2: std::sys_common::backtrace::_print

             at src/libstd/sys_common/backtrace.rs:47

   3: std::sys_common::backtrace::print

             at src/libstd/sys_common/backtrace.rs:36

   4: std::panicking::default_hook::{{closure}}

             at src/libstd/panicking.rs:200

   5: std::panicking::default_hook

             at src/libstd/panicking.rs:214

   6: std::panicking::rust_panic_with_hook

             at src/libstd/panicking.rs:477

   7: std::panicking::begin_panic

             at /rustc/dddb7fca09dc817ba275602b950bb81a9032fb6d/src/libstd/panicking.rs:411

   8: tikv_client::util::tests::test_global_timer

             at src/util.rs:72

   9: tikv_client::util::tests::test_global_timer::{{closure}}

             at src/util.rs:66

  10: core::ops::function::FnOnce::call_once

             at /rustc/dddb7fca09dc817ba275602b950bb81a9032fb6d/src/libcore/ops/function.rs:235

  11: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once

             at /rustc/dddb7fca09dc817ba275602b950bb81a9032fb6d/src/liballoc/boxed.rs:770

  12: __rust_maybe_catch_panic

             at src/libpanic_unwind/lib.rs:80

  13: std::panicking::try

             at /rustc/dddb7fca09dc817ba275602b950bb81a9032fb6d/src/libstd/panicking.rs:275

  14: std::panic::catch_unwind

             at /rustc/dddb7fca09dc817ba275602b950bb81a9032fb6d/src/libstd/panic.rs:394

  15: test::run_test::run_test_inner::{{closure}}

             at src/libtest/lib.rs:1471

[...]

PR #112, opened 2019-08-28
Replace tokio-timer with futures_timer

sticnarf:
With futures_timer, we don't need to maintain a timer thread ourselves. I think #105 will be fixed too.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,7 +1,6 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::{sync::mpsc, thread, time::Duration};
-use tokio_timer::{self, timer::Handle};
+use std::{thread, time::Duration};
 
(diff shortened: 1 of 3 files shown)


src/util.rs
```diff
@@ -38,37 +37,3 @@ pub fn duration_to_sec(d: Duration) -> f64 {
...
-        .unwrap();
-    rx.recv().unwrap()
-}
-
-#[cfg(test)]
-mod tests {
-    use futures::compat::Compat01As03;
-
-    #[test]
-    fn test_global_timer() {
```
sticnarf: futures_timer has similar test already. We don't need to repeat here.

sticnarf:
> I found a `use tokio_timer::timer::Handle` in `src/pd/request.rs`.
> Should we remove this import?

`src/pd/request.rs` is not used. Let me remove it later.

PR #115, opened 2019-09-04
Document fully qualified `into()` for keys and values

Hoverbear:
This adds the ability for downstreams to access contained values as `Vec<u8>`. It was previously allowed but removed in #77 .

### DIFF
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -34,15 +34,17 @@ use std::{fmt, u8};
 /// let from_bytes = Key::from(bytes);
 /// assert_eq!(from_static_str, from_bytes);
 /// ```
-/// 
-/// In order to get the wrapped value you can use `to_inner`:
-/// 
+///
+/// While `.into()` is usually sufficient for obtaining the buffer itself, sometimes type inference
+/// isn't able to determine the correct type. Notably in the `assert_eq!()` and `==` cases. In
+/// these cases using the fully-qualified-syntax is useful:
+///
 /// ```rust
 /// use tikv_client::Key;
-/// 
+///
 /// let buf = "TiKV".as_bytes().to_owned();
 /// let key = Key::from(buf.clone());
-/// assert_eq!(key.to_inner(), buf);
+/// assert_eq!(Into::<Vec<u8>>::into(key), buf);
 /// ```
 ///
 /// Many functions which accept a `Key` accept an `Into<Key>`, which means all of the above types
@@ -93,12 +95,6 @@ impl Key {
         } else {
             Bound::Excluded(self)
         }
-    }
-        
-    /// Unwrap the key into the byte vector contained within.
-    #[inline]
-    pub fn to_inner(self) -> Vec<u8> {
-        self.0
     }
 }
 
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -34,10 +34,19 @@ use std::{fmt, u8};
 /// let from_bytes = Key::from(bytes);
 /// assert_eq!(from_static_str, from_bytes);
 /// ```
+/// 
+/// In order to get the wrapped value you can use `to_inner`:
+/// 
+/// ```rust
+/// use tikv_client::Key;
+/// 
+/// let buf = "TiKV".as_bytes().to_owned();
+/// let key = Key::from(buf.clone());
+/// assert_eq!(key.to_inner(), buf);
+/// ```
 ///
-/// **But, you should not need to worry about all this:** Many functions which accept a `Key`
-/// accept an `Into<Key>`, which means all of the above types can be passed directly to those
-/// functions.
+/// Many functions which accept a `Key` accept an `Into<Key>`, which means all of the above types
+/// can be passed directly to those functions.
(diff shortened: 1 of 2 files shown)


Hoverbear:
@nrc It seems like our friend Rustc has trouble resolving sometimes, even with more than typically necessary type hints. I do think you're right in that this should be solvable from the context of `from` and `into` though, hrm...

Take a look at this:

```rust
use tikv_client::{Config, RawClient, Error, Value};

#[tokio::main]
async fn main() -> Result<(), Error> {
    let config = Config::new(vec![
        "192.168.0.100:2379",
    ]);
    let client = RawClient::new(config)?;
    let key = String::from("TiKV").into_bytes();
    let value: Vec<u8> = String::from("Works!").into_bytes();

    client.put(key, value).await?;

    let returned: Value = client.get(key).await?
        .expect("Value should be present.");
    
    assert_eq!(returned.into(), value);
    Ok(())
}
```

Shows:

```
   Compiling tikv-example v0.1.0 (/home/hoverbear/git/tikv-example)
error[E0283]: type annotations required: cannot resolve `tikv_client::kv::value::Value: std::convert::Into<_>`
  --> src/main.rs:17:25
   |
17 |     assert_eq!(returned.into(), value);
   |                         ^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0283`.
error: Could not compile `tikv-example`.

To learn more, run the command again with --verbose.
```

Similarly,

```rust
use tikv_client::{Config, RawClient, Error, Value};

#[tokio::main]
async fn main() -> Result<(), Error> {
    let config = Config::new(vec![
        "192.168.0.100:2379",
    ]);
    let client = RawClient::new(config)?;
    let key = String::from("TiKV").into_bytes();
    let value: Vec<u8> = String::from("Works!").into_bytes();

    client.put(key, value).await?;

    let returned: Value = client.get(key).await?
        .expect("Value should be present.");
    
    assert_eq!(Vec::<u8>::from(returned), value);
    Ok(())
}
```

Shows:

```
   Compiling tikv-example v0.1.0 (/home/hoverbear/git/tikv-example)
error[E0277]: the trait bound `std::vec::Vec<u8>: std::convert::From<tikv_client::kv::value::Value>` is not satisfied
  --> src/main.rs:17:16
   |
17 |     assert_eq!(Vec::<u8>::from(returned), value);
   |                ^^^^^^^^^^^^^^^ the trait `std::convert::From<tikv_client::kv::value::Value>` is not implemented for `std::vec::Vec<u8>`
   |
   = help: the following implementations were found:
             <std::vec::Vec<T> as std::convert::From<&[T]>>
             <std::vec::Vec<T> as std::convert::From<&mut [T]>>
             <std::vec::Vec<T> as std::convert::From<std::borrow::Cow<'a, [T]>>>
             <std::vec::Vec<T> as std::convert::From<std::boxed::Box<[T]>>>
           and 5 others
   = note: required by `std::convert::From::from`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `tikv-example`.

To learn more, run the command again with --verbose.
```

nrc:
Using `assert_eq` (or operator `==`) is basically worst case scenario for type inference (because there is a trait lookup, and Rust can't infer based on trait lookups). I think in more realistic code this would be fine.

Hoverbear:
@nrc I think `client.get("foo").await? == expected_value` is a perfectly valid use case. As with both those examples, but it seems they are actually impossible?

nrc:
They're not impossible, you just need a variable:

```
let value: Vec<u8> = client.get("foo").await?.into();
value == expected_value
```

Or just `client.get("foo").await? == expected_value` where `expected_value` has type `Value` rather than `Vec<u8>` (which is the style I would encourage).

Hoverbear:
Can you explain why you encourage that style? So you'd prefer the former over the latter?

```rust
fn main() {
    // Comes from wherever....
    let x = String::from("foo");
    // ...
    assert_eq!(x, String::from("foo"));
    // vs
    assert_eq!(&x, "foo");
}
```

nrc:
Sorry, to clarify, I don't much care about using value vs ref, but I think users should prefer using `Value` to `Vec<u8>` because the former is a bit more type-safe and abstract. (Although I have to admit I have been contemplating what it would look like if we dumped `Value` and `Key` and just use `Vec<u8>` everywhere).

Hoverbear:
I mean, if #77 has rendered our code unable to do these simple things because of Rust's deficiencies, perhaps it is worthwhile to just remove it. I know @sticnarf mentioned a similar feeling.

sticnarf:
`Vec::<u8>::from(value)` fails maybe because we implement `Into` instead of `From`?

I think `into` should be good for most cases. But I still think `into_inner` can be considered an option as it's clearer to the user that this operation is not costly.

Hoverbear:
Seems this can work for now and is an acceptable workaround

```rust
assert_eq!(Into::<Vec<u8>>::into(key), buf);
```

Hoverbear:
I updated this PR to include a functional doctest that demonstrates fully qualified `into` to workaround these cases.

PR #118, opened 2019-09-06
Replace the retry implementation.

nrc:
The changes outside of retry.rs are to support the change from futures to async/await. It is probably easier to review the whole file rather than the diff.

PTAL @sticnarf @Hoverbear

### DIFF
--- a/src/pd/retry.rs
+++ b/src/pd/retry.rs
@@ -8,14 +8,10 @@ use std::{
     time::Duration,
 };
 
-use futures::future::BoxFuture;
 use futures::prelude::*;
-use futures::ready;
-use futures::task::{Context, Poll};
 use futures_timer::Delay;
 use grpcio::Environment;
 use kvproto::metapb;
-use std::pin::Pin;
 
 use crate::{
     pd::{
@@ -27,6 +23,8 @@ use crate::{
     Result,
 };
 
+// FIXME: these numbers and how they are used are all just cargo-culted in, there
+// may be more optimal values.
 const RECONNECT_INTERVAL_SEC: u64 = 1;
 const MAX_REQUEST_COUNT: usize = 3;
 const LEADER_CHANGE_RETRY: usize = 10;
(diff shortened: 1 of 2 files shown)


src/pd/retry.rs
```diff
@@ -131,99 +109,153 @@ impl fmt::Debug for RetryClient {
...
 
-    // A function which makes an async request.
-    func: Func,
+    fn with_cluster<T, F: Fn(&Cluster) -> T>(&self, f: F) -> T {
+        f(&self.cluster.read().unwrap())
+    }
 }
 
-impl<Resp, Fut> Future for RetryRequest<Fut>
+async fn retry_request2<Rc, Resp, Func, RespFuture>(client: Arc<Rc>, func: Func) -> Result<Resp>
```
sticnarf: Why there is a `2`?

nrc: whoops, should have taken that out...

src/pd/retry.rs
```diff
@@ -131,99 +109,153 @@ impl fmt::Debug for RetryClient {
...
-    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Result<Resp>> {
-        unsafe {
-            let this = Pin::get_unchecked_mut(self);
-            if this.reconnect_count == 0 {
-                return Poll::Ready(Err(internal_err!("failed to send request")));
+    let mut last_err = Ok(());
+    for _ in 0..LEADER_CHANGE_RETRY {
+        // Reconnect if necessary.
+        let mut reconnect_count = MAX_REQUEST_COUNT;
+        while let Err(e) = client.reconnect(RECONNECT_INTERVAL_SEC) {
```
sticnarf: Looks like we always reconnect on the first try. We should reconnect only if there is an error.

nrc: addressed

nrc:
@sticnarf added another test

Issue #114, opened 2019-09-02
UCP: Add region cache for PD


sticnarf:
## Description
It costs too much if we query PD for region information in every request. We can save region information in cache and query if necessary (on region error or the region for the key is unknown)

It's a good to idea to use the TiDB implementation as a reference: https://github.com/pingcap/tidb/blob/master/store/tikv/region_cache.go

@nrc has a PR which gives an incomplete implementation: https://github.com/tikv/client-rust/pull/121. You can continue that PR as well.

## Difficulty
- Medium

## Score
- 3000

## Mentor(s)
- @sticnarf 
- @nrc 

## Recommended Skills
- Rust
- TiKV Architecture

SSebo:
@sticnarf Could you explain a little bit what is the difficult part of this issue? 3000 points is really high.

sticnarf:
@SSebo This task is not just porting the region cache implementation. The difficulty here is the associating request error handling with cache manipulation.

So if you want to try this task, you may not only look at the region cache in TiDB itself, but also how the cache is used and refreshed on errors.

SSebo:
@sticnarf I'll have a try after I got my 200 points. Hope this issue is unpick until then.

SSebo:
Is there any resources that should read before pick this issue?

sticnarf:
> Is there any resources that should read before pick this issue?

I'm sorry but I don't know any text resource about it. TiDB code https://github.com/pingcap/tidb/blob/master/store/tikv/region_cache.go and https://github.com/pingcap/tidb/blob/master/store/tikv/region_request.go may be helpful but you don't necessarily copy the exact implementation of TiDB.

SSebo:
unformally picked up

SSebo:
@sticnarf  Can I continue nrc's work? I just merge nrc `region-cache` branch to `master`, Could you reopen that PR?

sticnarf:
@SSebo I've reopened it.

SSebo:
@sticnarf bot close It again, I'll wait until my easy PR merged

PR #121, opened 2019-10-07
Add a region caches

nrc:
Builds on https://github.com/nrc/client-rust/pull/new/region-cache. It is pretty hard to test this. I think we need integration tests on mock TiKV so that we can cause region failures. There are some unit tests, but not as many as I would like.

PTAL @sticnarf @Hoverbear

### DIFF
--- a/src/mock.rs
+++ b/src/mock.rs
@@ -111,7 +111,7 @@ impl PdClient for MockPdClient {
         ))))
     }
 
-    fn region_for_key(&self, key: &Key) -> BoxFuture<'static, Result<Region>> {
+    fn region_for_key(&self, key: &Key, _skip_cache: bool) -> BoxFuture<'static, Result<Region>> {
         let bytes: &[_] = key.into();
         let region = if bytes.is_empty() || bytes[0] < 10 {
             Self::region1()
@@ -122,7 +122,7 @@ impl PdClient for MockPdClient {
         Box::pin(ready(Ok(region)))
     }
 
-    fn region_for_id(&self, id: RegionId) -> BoxFuture<'static, Result<Region>> {
+    fn region_for_id(&self, id: RegionId, _skip_cache: bool) -> BoxFuture<'static, Result<Region>> {
         let result = match id {
             1 => Ok(Self::region1()),
             2 => Ok(Self::region2()),
--- a/src/pd/retry.rs
+++ b/src/pd/retry.rs
@@ -21,7 +21,7 @@ use crate::{
     },
     security::SecurityManager,
     transaction::Timestamp,
-    Result,
+    Key, Result,
 };
 
 // FIXME: these numbers and how they are used are all just cargo-culted in, there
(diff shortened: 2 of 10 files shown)


src/mock.rs
```diff
@@ -111,7 +111,7 @@ impl PdClient for MockPdClient {
         ))))
     }
 
-    fn region_for_key(&self, key: &Key) -> BoxFuture<'static, Result<Region>> {
+    fn region_for_key(&self, key: &Key, _: bool) -> BoxFuture<'static, Result<Region>> {
```
Hoverbear:
```suggestion
    fn region_for_key(&self, key: &Key, _skip_cache: bool) -> BoxFuture<'static, Result<Region>> {
```

src/mock.rs
```diff
@@ -122,7 +122,7 @@ impl PdClient for MockPdClient {
         Box::pin(ready(Ok(region)))
     }
 
-    fn region_for_id(&self, id: RegionId) -> BoxFuture<'static, Result<Region>> {
+    fn region_for_id(&self, id: RegionId, _: bool) -> BoxFuture<'static, Result<Region>> {
```
Hoverbear:
```suggestion
    fn region_for_id(&self, id: RegionId, _skip_cache: bool) -> BoxFuture<'static, Result<Region>> {
```

src/request.rs
```diff
@@ -44,7 +44,7 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
         mut self,
         pd_client: Arc<impl PdClient>,
     ) -> BoxStream<'static, Result<Self::RpcResponse>> {
-        let stores = self.store_stream(pd_client.clone());
+        let stores = self.store_stream(pd_client.clone(), false);
```
sticnarf: I think it's not correct to always set it to false. Parameter `is_retry` should be also added to this method and set `is_retry` to `true` on region error.

nrc:
Comments addressed, PTAL @sticnarf @Hoverbear

src/pd/retry.rs
```diff
@@ -66,21 +69,47 @@ impl RetryClient<Cluster> {
...
             timeout,
+            cache: RwLock::new(RegionCache::new()),
         })
     }
 
     // These get_* functions will try multiple times to make a request, reconnecting as necessary.
-    pub async fn get_region(self: Arc<Self>, key: Vec<u8>) -> Result<Region> {
+    pub async fn get_region(self: Arc<Self>, key: Key, skip_cache: bool) -> Result<Region> {
+        if skip_cache {
+            self.cache.write().unwrap().remove_region_by_start_key(&key);
```
sticnarf: The `key` here isn't always the start key of a region. Maybe we can add a new method to the region cache.

Hoverbear:
So this fixes #114 ?

sticnarf:
> So this fixes #114 ?

If this PR gets merged (without changing a lot), maybe #114 can be closed but we need to create new issues to track the unsolved problems.

src/pd/retry.rs
Hoverbear: @nrc PTAL

sadroeck:
Any update on this?

nrc:
I'll come back to this some day, but closing for now since I don't have time.

PR #123, opened 2019-11-07
Compiling on stable. Also fixed panic when region ID is not found

Snnappie:
Compiles on stable 1.39.0. Tests still require nightly because of the use of specialization.


Hoverbear:
@Snnappie I used some of this in #124 which gets `cargo test` working too.

Hoverbear:
@Snnappie Could you resolve the conflicts? Then we should be good to merge!

Snnappie:
@Hoverbear Done

nrc:
@Snnappie thanks for updating the branch. There is a compilation error, could you fix it please?

PR #124, opened 2019-11-19
update nightly

Hoverbear:
Update to latest nightly. :)


src/pd/timestamp.rs
```diff
@@ -125,7 +125,7 @@ struct TsoRequestStream<'a> {
 }
 
 impl<'a> Stream for TsoRequestStream<'a> {
-    type Item = (TsoRequest, WriteFlags);
+    type Item = Result<(TsoRequest, WriteFlags)>;
```
nrc: Why do we need this? Seems like we'll never use the `Err` case

Hoverbear: It's needed for `join!` to work since it wants a `TryStream`.

nrc: I'm not sure why `join` needs a `TryStream`, it should work fine with a regular stream. Do we use `join` with a different `TryStream` and we need this one to match? I'm not sure if it makes it any better, but you could use `.map(Ok)` to turn the stream into one which implements `TryStream`.

Hoverbear: Should be fixed :)

Issue #125, opened 2019-12-25
build failed with kvproto


zyctree:
here is the error message
```shell
error: no matching package named `protobuf` found
location searched: https://github.com/nrc/rust-protobuf?branch=v2.8
perhaps you meant: protoc
required by package `kvproto v0.0.2 (https://github.com/pingcap/kvproto.git?rev=2fc6229ed6097b59dbe51525c7a65b19543a94ca#2fc6229e)`
    ... which is depended on by `tikv-client v0.0.0 (/data/git/rust/ping_cap/client-rust)`
```
found it can build after changing
```toml
kvproto = { git = "https://github.com/pingcap/kvproto.git", rev = "2fc6229ed6097b59dbe51525c7a65b19543a94ca", features = [ "prost-codec" ], default-features = false }
```
to
```toml
kvproto = { git = "https://github.com/pingcap/kvproto.git", features = [ "prost-codec" ], default-features = false }
```
maybe it needs a new rev

divinerapier:
After replacing `kvproto` gets error

```
error[E0433]: failed to resolve: use of undeclared type or module `LogType`
  --> client-rust/target/debug/build/kvproto-358e45cab457932c/out/protos/wrapper_cdcpb.rs:94:53
   |
94 | #[inline] pub fn get_type(&self) -> LogType { match LogType::from_i32(self.r#type) {Some(e) => e,
   |                                                     ^^^^^^^ use of undeclared type or module `LogType`

error[E0433]: failed to resolve: use of undeclared type or module `row`
  --> client-rust/target/debug/build/kvproto-358e45cab457932c/out/protos/wrapper_cdcpb.rs:98:40
   |
98 | #[inline] pub fn get_op_type(&self) -> row::OpType { match row::OpType::from_i32(self.op_type) {Some(e) => e,
   |                                        ^^^ use of undeclared type or module `row`

error[E0433]: failed to resolve: use of undeclared type or module `row`
  --> client-rust/target/debug/build/kvproto-358e45cab457932c/out/protos/wrapper_cdcpb.rs:98:60
   |
98 | #[inline] pub fn get_op_type(&self) -> row::OpType { match row::OpType::from_i32(self.op_type) {Some(e) => e,
   |                                                            ^^^ use of undeclared type or module `row`

error[E0412]: cannot find type `LogType` in this scope
  --> client-rust/target/debug/build/kvproto-358e45cab457932c/out/protos/wrapper_cdcpb.rs:94:37
   |
[...]

PR #126, opened 2020-01-15
Fix protobuf dependency

zyctree:
fix https://github.com/tikv/client-rust/issues/125

the rev of protobuf is copied from kvproto,

seems the `[replace]` section takes effect for itself, but not the crates that depend it


nrc:
Looks good! Could you sign the commit please?

nrc:
@sticnarf PTAL

zyctree:
just force pushed to add sign-off

PR #128, opened 2020-02-02
Convert part of futures to `async/await`

### LINKED ISSUE #119: Convert futures to async/await
We have a lot of async code now, but are still using some futures, converting them all (or as much as possible) would make the code clearer.

silathdiir:
Tries to convert part of futures to `async/await`. Fixes #119.

Signed-off-by: Steven Gu <asongala@163.com>

### DIFF
--- a/src/kv_client/mod.rs
+++ b/src/kv_client/mod.rs
@@ -78,23 +78,27 @@ impl KvClient for KvRpcClient {
     }
 }
 
-fn map_errors_and_trace<Resp, RpcFuture>(
+async fn map_errors_and_trace<Resp, RpcFuture>(
     request_name: &'static str,
     fut: ::grpcio::Result<RpcFuture>,
-) -> impl Future<Output = Result<Resp>>
+) -> Result<Resp>
 where
     Compat01As03<RpcFuture>: Future<Output = std::result::Result<Resp, ::grpcio::Error>>,
     Resp: HasError + Sized + Clone + Send + 'static,
 {
     let context = tikv_stats(request_name);
 
-    // FIXME should handle the error, not unwrap.
-    Compat01As03::new(fut.unwrap())
-        .map(|r| match r {
-            Err(e) => Err(ErrorKind::Grpc(e).into()),
-            Ok(r) => Ok(r),
-        })
-        .map(move |r| context.done(r))
+    let fut = match fut {
+        Err(e) => return context.done(Err(ErrorKind::Grpc(e).into())),
+        Ok(f) => f,
+    };
+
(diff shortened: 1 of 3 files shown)


src/kv_client/mod.rs
```diff
@@ -78,23 +78,27 @@ impl KvClient for KvRpcClient {
...
-    Compat01As03::new(fut.unwrap())
-        .map(|r| match r {
-            Err(e) => Err(ErrorKind::Grpc(e).into()),
-            Ok(r) => Ok(r),
-        })
-        .map(move |r| context.done(r))
+    let fut = match fut {
+        Err(e) => return context.done(Err(ErrorKind::Grpc(e).into())),
+        Ok(f) => f,
+    };
```
nrc: Could this be written as `context.done(fut.and_then(async |f| Compat01As03::new(fut).await).map_err(|e| ErrorKind::Grpc(e).into()))` or something similar?

src/raw/client.rs
```diff
@@ -256,7 +259,9 @@ impl Client {
                 requests::new_raw_scan_request(range, limit, self.key_only, self.cf.clone())
                     .execute(self.rpc.clone()),
             )
-        }
+        };
+
+        request.await
```
nrc: I think this function can be further simplified by directly returning the error in the first case and using `...execute.await` in the second.

src/raw/client.rs
```diff
@@ -275,12 +280,12 @@ impl Client {
...
     /// ```
-    pub fn batch_scan(
+    pub async fn batch_scan(
         &self,
         ranges: impl IntoIterator<Item = impl Into<BoundRange>>,
         each_limit: u32,
-    ) -> impl Future<Output = Result<Vec<KvPair>>> {
-        if each_limit > MAX_RAW_KV_SCAN_LIMIT {
+    ) -> Result<Vec<KvPair>> {
+        let request = if each_limit > MAX_RAW_KV_SCAN_LIMIT {
```
nrc: Similarly to above

src/kv_client/mod.rs
silathdiir: @nrc It seems that the closure `async |f| Compat01As03::new(fut).await` could not return a Result (it returns a Future).

I tries to simplify the code as below for code review.

```
    let res = match fut {
        Ok(f) => Compat01As03::new(f).await,
        Err(e) => Err(e),
    };

    let context = tikv_stats(request_name);
    context.done(res.map_err(|e| ErrorKind::Grpc(e).into()))
```

nrc:
@sticnarf PTAL

PR #129, opened 2020-02-22
Add backoff and retry limit to requests

### LINKED ISSUE #110: UCP: Add backoff and retry limit to requests
## Description

We shouldn't allow infinite or immediate retries. It is better to add backoff retry delay.

## Difficulty
- Easy

## Score
- 50

## Mentor(s)
- @sticnarf 
- @nrc

## Recommended Skills
- Rust

silathdiir:
Adds backoff and retry limit to requests. Fixes #110.

Signed-off-by: Steven Gu <asongala@163.com>


src/request.rs
```diff
@@ -16,6 +16,14 @@ use std::sync::Arc;
 use std::time::Duration;
 
 const LOCK_RETRY_DELAY_MS: u64 = 10;
+const REGION_RETRY_DEFAULT_DELAY_MS: u64 = 500;
+const REGION_RETRY_MAX_COUNT: usize = 3;
+
+#[derive(Clone, Copy)]
+pub struct RetryState {
+    region_retry_count: usize,
+    region_retry_delay_ms: u64,
```
nrc: It looks like the delay is never changed, so we should just be able to use `REGION_RETRY_MAX_COUNT` and remove this field

silathdiir:
@sticnarf Thanks for the reference from `tidb` implementation. I will try to update for implementing a similar one of exponential backoff.

silathdiir:
@nrc @sticnarf Hi, I update the code according to previous comments, please help review if you have time. Thanks.

src/backoff.rs
```diff
@@ -0,0 +1,242 @@
...
+pub trait Backoff: Clone + Copy + Send + Sync + 'static {
+    fn next_delay_ms(&mut self) -> Option<u64>;
+
+    fn exponential_delay_ms(base_delay_ms: u64, max_delay_ms: u64, attempts: u32) -> u64 {
+        max_delay_ms.min(2u64.pow(attempts) * base_delay_ms)
+    }
+}
+
+#[derive(Clone, Copy)]
+pub struct BackoffState {
```
sticnarf: I think it might be a better design to put only mutable things in the state. Immutable fields can be saved in the backoff type directly.

src/backoff.rs
```diff
@@ -0,0 +1,242 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
+
+// https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+
+use rand::{thread_rng, Rng};
+
+pub trait Backoff: Clone + Copy + Send + Sync + 'static {
+    fn next_delay_ms(&mut self) -> Option<u64>;
+
+    fn exponential_delay_ms(base_delay_ms: u64, max_delay_ms: u64, attempts: u32) -> u64 {
```
sticnarf: Seems that it can be simply a free function instead of a trait method.

src/backoff.rs
```diff
@@ -0,0 +1,242 @@
...
+        max_delay_ms.min(2u64.pow(attempts) * base_delay_ms)
```
sticnarf: It could be more efficient to simply multiply the last delay_ms by 2 to get the next one.

sticnarf: And then, it looks that we needn't save `base_delay`.

src/backoff.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        let base_delay_ms = match base_delay_ms {
+            0 | 1 => 2,
+            v => v,
+        };
+
+        let max_delay_ms = match max_delay_ms {
+            0 | 1 => 2,
+            v => v,
+        };
```
sticnarf: These checks only apply on jitter backoffs? Maybe we can check them only when creating those kinds of backoffs.

sticnarf: I think it's all right to just panic on bad arguments. Implicit fixes may bring unexpected result.

src/request.rs
```diff
@@ -41,8 +45,26 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
...
+        self,
+        pd_client: Arc<impl PdClient>,
+    ) -> BoxStream<'static, Result<Self::RpcResponse>> {
+        let backoff_state = BackoffState::new(
+            REGION_RETRY_BASE_DELAY_MS,
+            REGION_RETRY_MAX_DELAY_MS,
+            REGION_RETRY_ATTEMPTS,
+        );
+
+        let backoff = Arc::new(NoJitterBackoff {
```
sticnarf: It looks a bit strange that we need to create that backoff even if there isn't any error...

sticnarf: If you think a backoff is too large to copy, I think the `Copy` and `Clone` trait on `Backoff` is unnecessary.

src/request.rs
```diff
@@ -78,12 +104,25 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
...
         self,
-        _region_error: Error,
+        region_error: Error,
         pd_client: Arc<impl PdClient>,
+        backoff: Arc<impl Backoff>,
     ) -> BoxStream<'static, Result<Self::RpcResponse>> {
-        // Retry on region errors by default
-        // TODO: Add backoff and retry limit
-        self.response_stream(pd_client)
+        let mut backoff = *backoff;
```
sticnarf: Here you do a memcpy of the backoff...
We can use `Box` instead of `Arc`, then `backoff` can be mutable.

src/request.rs
silathdiir: @sticnarf Hi, I fix to remove `Arc` and wrap `impl` Backoff` with Option. So it could be assigned to `None::<NoJitterBackoff>` for the first time. Please help review the update. Thanks.

src/backoff.rs
```diff
@@ -0,0 +1,262 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
+
+// https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+
+use rand::{thread_rng, Rng};
+
+pub trait Backoff: Clone + Copy + Send + 'static {
+    fn next_delay_ms(&mut self) -> Option<u64>;
```
sticnarf: Could you give some comments describing what `None` means?

sticnarf: And could you use `Duration` instead of `_ms`?

src/request.rs
```diff
@@ -16,6 +17,9 @@ use std::sync::Arc;
 use std::time::Duration;
 
 const LOCK_RETRY_DELAY_MS: u64 = 10;
+const REGION_RETRY_BASE_DELAY_MS: u64 = 2;
+const REGION_RETRY_MAX_DELAY_MS: u64 = 500;
+const REGION_RETRY_ATTEMPTS: u32 = 10;
```
sticnarf: I think we can just declare a const `NoJitterBackoff` for it.

src/backoff.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+#[derive(Clone, Copy)]
```
sticnarf: I don't think we need to mark it `Copy`. Explicit clones will show how the states are changed more clearly.

src/request.rs
silathdiir: Hi @sticnarf , I replace these constants to one `const NoJitterBackoff`, and fix initializing this struct directly in other places (the compiler prompted the `new` function must be `const`, it looks a little strange, so I remove the `new` function).

And I update other places based on your code review. Please help review. Thanks.

sticnarf: I think the right approach is just declare the `new` function as a `const fn`. `const fn`s are used to return a const value.

silathdiir: Hi @sticnarf , I rollbacks the `new` function of struct NoJitterBackoff, and adds `const`. Please help review. Thanks.

sticnarf:
PTAL again @nrc

sticnarf:
@silathdiir Would you like to also solve this issue? https://github.com/tikv/client-rust/issues/135

silathdiir:
@sticnarf Yes, I pickup this issue #135 . Thanks.

silathdiir:
Hi @sticnarf @nrc , I am sorry that I found two CI compile errors on my last commit.

1. [CI error URL link 1](https://travis-ci.org/github/tikv/client-rust/jobs/663407420?utm_medium=notification&utm_source=github_status)
1. [CI error URL link 2](https://travis-ci.org/github/tikv/client-rust/jobs/663565564?utm_medium=notification&utm_source=github_status)

I tries to update `Cargo.toml` according to the one in project [kvproto](https://github.com/pingcap/kvproto/blob/master/Cargo.toml).
It seems that the first CI error could be fixed after update, and the CI `osx` build could work.
But the second one always occurred on `linux`:

```
--- stderr
third_party/boringssl/crypto/CMakeFiles/crypto.dir/build.make:83: recipe for target 'third_party/boringssl/crypto/err_data.c' failed

CMakeFiles/Makefile2:1269: recipe for target 'third_party/boringssl/crypto/CMakeFiles/crypto.dir/all' failed

CMakeFiles/Makefile2:244: recipe for target 'CMakeFiles/grpc.dir/rule' failed

Makefile:216: recipe for target 'grpc' failed

--- stderr

CMake Warning at cmake/protobuf.cmake:47 (message):

  gRPC_PROTOBUF_PROVIDER is "module" but PROTOBUF_ROOT_DIR is wrong

Call Stack (most recent call first):

  CMakeLists.txt:151 (include)

libunwind not found. Disabling unwind tests.

go build command-line-arguments: module requires Go 1.13

make[3]: *** [third_party/boringssl/crypto/err_data.c] Error 1

make[3]: *** Deleting file 'third_party/boringssl/crypto/err_data.c'

make[2]: *** [third_party/boringssl/crypto/CMakeFiles/crypto.dir/all] Error 2

make[1]: *** [CMakeFiles/grpc.dir/rule] Error 2

make: *** [grpc] Error 2

thread 'main' panicked at '

command did not execute successfully, got: exit code: 2
```

Is it an issue on feature `openssl` or another? Please give some suggestions on how to fix it if you have time. Thanks.

src/backoff.rs
```diff
@@ -0,0 +1,294 @@
...
+
+        self.current_attempts += 1;
+        self.current_delay_ms <<= 1;
+
+        Some(Duration::from_millis(delay_ms))
+    }
+}
+
+#[derive(Clone)]
+pub struct FullJitterBackoff {
```
nrc: Could you comment on these structs to describe the differences please?

src/request.rs
```diff
@@ -41,8 +43,16 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
     }
 
     fn response_stream(
+        self,
+        pd_client: Arc<impl PdClient>,
+    ) -> BoxStream<'static, Result<Self::RpcResponse>> {
+        self.retry_response_stream(pd_client, None::<NoJitterBackoff>)
```
nrc: Use `DEFAULT_REGION_BACKOFF` here rather than None

src/request.rs
```diff
@@ -41,8 +43,16 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
...
+    }
+
+    fn retry_response_stream(
         mut self,
         pd_client: Arc<impl PdClient>,
+        backoff: Option<impl Backoff>,
```
nrc: remove `Option`

src/backoff.rs
silathdiir: I add comment and formula to each struct Backoff, and remove Option from the parameter backoff you mentioned below. Thanks.

sticnarf:
@silathdiir Some components may be unavailable in nightly Rust. Would you please modify `.travis.yml` so `rustfmt` and `clippy` are only installed and used in stable?

silathdiir:
@sticnarf OK, I will update `travis.yml` for only running `rustfmt` and `clippy` on Rust stable now.

silathdiir:
@sticnarf I updated `.travis.yml` as you mentioned. It works. Thanks a lot. Please help review.

sticnarf:
Thanks for your contribution!

Issue #130, opened 2020-02-28
UCP: Implement RawBatchScan


sticnarf:
## Description

Implement the `RawBatchScan` interface. The protocol is:

```
message RawBatchScanRequest {
    Context context = 1;
    repeated KeyRange ranges = 2; // scanning range
    uint32 each_limit = 3; // max number of returning kv pairs for each scanning range
    bool key_only = 4;
    string cf = 5;
    bool reverse = 6;
}

message RawBatchScanResponse {
    errorpb.Error region_error = 1;
    repeated KvPair kvs = 2;
}
```

## Difficulty

- Easy

## Score
- 50

## Mentor(s)
- @sticnarf 
- @nrc 

## Recommended Skills
- Rust

PR #134, opened 2020-03-15
raw: Implement `RawBatchScan`

you06:
Signed-off-by: you06 <you1474600@gmail.com>

UCP: #130 

Currently all requests should be processed in a stream, this PR wrap the range batch in a stream, though there is only one item in it.


nrc:
You need to run Rustfmt to pass CI. There is also a failing test, I'm not sure if it is a genuine failure or not.

Cargo.toml
```diff
@@ -46,5 +46,5 @@ fail = { version = "0.3", features = [ "failpoints" ] }
 
 [replace]
 "protobuf:2.8.0" = { git = "https://github.com/nrc/rust-protobuf", rev="4df576feca3b10c01d55b0e7c634bfab30982087" }
-"protobuf-build:0.10.0" = { git = "https://github.com/tikv/protobuf-build", rev="42e52b9311f7fb31bbbe089fef5a24ec0392f9ce" }
+"protobuf-build:0.10.0" = { git = "https://github.com/tikv/protobuf-build", rev="1fe5d77689d9691b7917215f3a1d32968115cc02" }
```
nrc: Is this change needed? Is it an update?

examples/raw.rs
```diff
@@ -94,10 +94,29 @@ async fn main() -> Result<()> {
...
-        &[Key::from("k1".to_owned()), Key::from("k2".to_owned())]
+        &[Key::from("k1".to_owned()), Key::from("k2".to_owned()), Key::from("k3".to_owned())]
     );
     println!("Scaning from {:?} to {:?} gives: {:?}", start, end, keys);
 
+    let k1 = "k1";
+    let k2 = "k2";
+    let k3 = "k3";
+    let batch_scan_keys = vec![
+        (k1.clone().to_owned()..=k2.clone().to_owned()).to_owned(),
```
nrc: You don't need to clone here, `to_owned` always clones.

examples/raw.rs
```diff
@@ -94,10 +94,29 @@ async fn main() -> Result<()> {
...
+    ];
+    let kv_pairs = client
+        .batch_scan(batch_scan_keys.to_owned(), 10)
+        .await
+        .expect("Could not batch scan");
+    let vals: Vec<_> = kv_pairs.into_iter().map(|p| p.value().clone()).collect();
+    assert_eq!(&vals,
+        &[Value::from("v1".to_owned()), Value::from("v2".to_owned()),
+            Value::from("v2".to_owned()), Value::from("v3".to_owned()),
+            Value::from("v1".to_owned()), Value::from("v2".to_owned()), Value::from("v3".to_owned())]);
```
nrc: Can you use `.into` rather than `Value::from`?

examples/raw.rs
```diff
@@ -94,10 +94,29 @@ async fn main() -> Result<()> {
...
+    let batch_scan_keys = vec![
+        (k1.clone().to_owned()..=k2.clone().to_owned()).to_owned(),
+        (k2.to_owned()..=k3.clone().to_owned()).to_owned(),
+        (k1.to_owned()..=k3.to_owned()).to_owned(),
+    ];
+    let kv_pairs = client
+        .batch_scan(batch_scan_keys.to_owned(), 10)
+        .await
+        .expect("Could not batch scan");
+    let vals: Vec<_> = kv_pairs.into_iter().map(|p| p.value().clone()).collect();
```
nrc: I think you use `p.Value` rather than `p.value().clone()`

src/pd/client.rs
```diff
@@ -108,6 +108,18 @@ pub trait PdClient: Send + Sync + 'static {
         })
         .boxed()
     }
+
+    // Wrap ranges in a Stream.
+    fn stores_for_ranges(
+        self: Arc<Self>,
+        ranges: Vec<BoundRange>,
+    ) -> BoxStream<'static, Result<(Vec<BoundRange>, Store<Self::KvClient>)>> {
+        self.region_for_key(&Key::from(Vec::new()))
```
nrc: What is happening here? It seems a bit odd to find the region for the empty key, rather than something in `ranges`.

src/raw/requests.rs
```diff
@@ -432,9 +432,11 @@ impl KvRequest for kvrpcpb::RawBatchScanRequest {
 
     fn store_stream<PdC: PdClient>(
         &mut self,
-        _pd_client: Arc<PdC>,
+        pd_client: Arc<PdC>,
     ) -> BoxStream<'static, Result<(Self::KeyData, Store<PdC::KvClient>)>> {
-        future::err(Error::unimplemented()).into_stream().boxed()
+        let ranges = mem::replace(&mut self.ranges, Default::default())
+            .into_iter().map(BoundRange::from).collect();
```
nrc: Prefer using `into` rather than `from`

src/request.rs
```diff
@@ -153,6 +153,16 @@ pub fn store_stream_for_range<PdC: PdClient>(
         .boxed()
 }
 
+pub fn store_stream_for_ranges<PdC: PdClient>(
```
nrc: Why is this function needed, rather than just using `stores_for_ranges`?

src/request.rs
```diff
@@ -153,6 +153,16 @@ pub fn store_stream_for_range<PdC: PdClient>(
...
+    ranges: Vec<BoundRange>,
+    pd_client: Arc<PdC>,
+) -> BoxStream<'static, Result<(Vec<BoundRange>, Store<PdC::KvClient>)>> {
+    pd_client
+        .stores_for_ranges(ranges)
+        .into_stream()
```
nrc: `stores_for_ranges` returns a boxed stream, so why do we need this?

src/pd/client.rs
you06: Yeah, I do think it's odd. I just want to get store here.

sticnarf: I don't think it's correct here. You should find the stores and regions by the ranges.

you06: Oh I understand, if the ranges are in several stores, than I should make multiple requests right?

sticnarf: Yes. Every request can only match one region.

you06: Thanks, I'll update it!

you06:
Aha, build after rebase master keep failed until I remove my `Cargo.lock`. Why do we ignore this file?

you06:
Sorry for the late update, I've updated the function which find store for ranges and added a test. @nrc @sticnarf  PTAL again.

src/pd/client.rs
```diff
@@ -108,6 +108,58 @@ pub trait PdClient: Send + Sync + 'static {
         })
         .boxed()
     }
+
+    // Returns a Steam which iterates over the contexts for ranges in the same region.
+    fn group_ranges_by_region(
+        self: Arc<Self>,
+        mut ranges: Vec<BoundRange>,
+    ) -> BoxStream<'static, Result<(RegionId, Vec<BoundRange>)>> {
+        ranges.reverse();
```
nrc, line 117: I think there is another 'group by regions' function, can it share code with this new one?

you06: Do you mean group `group_keys_by_region` and `group_ranges_by_region` together into `group_by_region`?

nrc: Yeah, if it is possible and not too ugly.

you06: I tried this by implementing a `IntoRange` trait for key data types.
```rust
pub trait IntoRange {
    fn into_keys(self) -> (Key, Option<Key>);
    fn set_start_key(&mut self, k: Key);
    fn set_end_key(&mut self, k: Key);
}
```
After implementing this trait for `Key` and `BoundRange` I found some key data types are imported from other packages like `kvproto::protos::kvrpcpb::KvPair` and `kvproto::protos::kvrpcpb::Mutation` which makes the this work more strange. Besides, how we group ranges is more complex than group keys, because we need to split ranges and push the right part back into the range vector in this process, the logic can be redundant for key like data.
I don't have good ideas about how to doing this, what your opinion?

nrc: Yeah, seems like that is more work that it is worth. Maybe it is possible to factor out a helper function or two, or a macro? Otherwise don't worry about it.

you06: Let me try out macro~

you06:
I digged into macro usage, seems it's not able to match types inside a macro. Thus I use the basic match arm of macro, still needs to specify the data type, but use only one macro function now. @nrc PTAL.

nrc:
@sticnarf could you take a look please?

src/pd/client.rs
```diff
@@ -25,6 +25,16 @@ use crate::{
 const CQ_COUNT: usize = 1;
 const CLIENT_PREFIX: &str = "tikv-client";
 
+#[macro_export]
+macro_rules! group_by_range {
```
sticnarf, line 29: I doubt whether this macro is useful. It looks no different from just calling `group_ranges_by_region` or `group_keys_by_region`.
I guess @nrc hopes to make `group_ranges_by_region` and `group_keys_by_region` share some code?

you06: I've made some efforts on this, but it seems not able to merge the group function of key-like data and range-like data together.

sticnarf: Ok, that can be a bit difficult.
But then I think this macro is unnecessary. What do you think @nrc ?

nrc: Yeah, I'm happy either way (thanks for trying @you06 !)

you06: The macro has been removed.

PR #137, opened 2020-03-22
raw: Fix scan range in raw scan request

tz70s:
Prior scan range is ignored before passing into requests.
Propagate out each region range and pass them into requests.

Signed-off-by: Tzu-Chiao Yeh <su3g4284zo6y7@gmail.com>


ekexium:
Fixed in #164

Issue #95, opened 2019-08-06
UCP: Check if the PD server responds with less timestamps


sticnarf:
## Description
As discussed in https://github.com/tikv/client-rust/pull/92#discussion_r310870663, when the PD server returns less timestamps than requested, there will be more and more pending requests and the timestamp oracle will finally block forever.

For a robust client, we do need extra checks to detect it.

## Difficulty
- Easy

## Score
- 50

## Mentor(s)
- @sticnarf 
- @nrc 

## Recommended Skills
- Rust

Renkai:
pick up, unformally

PR #138, opened 2020-03-25
UCP: Check if the PD server responds with less timestamps

Renkai:
resolved https://github.com/tikv/client-rust/issues/95.

### DIFF
--- a/src/pd/timestamp.rs
+++ b/src/pd/timestamp.rs
@@ -136,26 +136,24 @@ impl<'a> Stream for TsoRequestStream<'a> {
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
         let pending_requests = self.pending_requests.clone();
         let mut pending_requests = pending_requests.borrow_mut();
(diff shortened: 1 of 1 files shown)


src/pd/timestamp.rs
```diff
@@ -132,10 +137,12 @@ impl<'a> Stream for TsoRequestStream<'a> {
         let pending_requests = self.pending_requests.clone();
         let mut pending_requests = pending_requests.borrow_mut();
         let mut count = 0;
```
sticnarf: It looks that `count` is redundant as we can rely on `requests.len()`.

Renkai: > It looks that `count` is redundant as we can rely on `requests.len()`.

updated

sticnarf: Please fix clippy

sticnarf:
Thanks for your contribution!

you06:
@Renkai the points is rewarded. FYI, all points in tikv org is shared, which means you need to pick up some medium/hard tasks next time :)

PR #139, opened 2020-03-31
[WIP] UCP 114: Add region cache for PD 

SSebo:
UCP #114 

Signed-off-by: ssebo <ssebo.zzz@gmail.com>

### What problem does this PR solve?

It costs too much if we query PD for region information in every request. We can save region information in cache and query if necessary (on region error or the region for the key is unknown)

### What is changed and how it works?

What's Changed:

continue nrc's work, right now just merge the latest code.

- [ ] In a single request, if multiple keys are in the same region, every key invalidates the cache even if it's refreshed by another key in the same request.

- [ ] Skip cache is too coarse grained. A region error can tell us which region's information is stale, we don't need to skip cache for all keys.

- [ ] When the info of one region changes, all requests related to it will get a region error and they all retry with "skip cache" option. This enlarges the above two problems.

- [ ]  have a mechanism to invalidate a specific region's info.

How it Works:

### Check List 

Tests 
- Integration test

###  Refer to a related PR or issue link (optional)
https://github.com/tikv/client-rust/issues/114
https://github.com/tikv/client-rust/pull/121
 
### Side effects

- Performance regression

- Breaking backward compatibility

### Release note


SSebo:
@sticnarf mentor, 

>  In a single request, if multiple keys are in the same region, every key invalidates the cache even if it's refreshed by another key in the same request.
I don't fully understand this bad case. Could you explain more detail to it ? 
I already picked this issue, Could you reopen this PR?

sticnarf:
> * I don't fully understand this bad case. Could you explain more detail to it ?

That's a problem in the detailed implementation of that PR. For example see this FIXME here: https://github.com/tikv/client-rust/pull/121/files#diff-49f989cfce4dadfa41d7d1bf2548edd7R40

SSebo:
@sticnarf hi, mentor, maybe we can maintain region cache status like `updating`, `updated` for reducing duplicated cache updates:
- solution 1: requests when region error, check region cache status, if status is `updating`, just backoff to next round until status is updated.
- solution 2: requests when region error, use future `poll` to check region cache status, if status is `updating` return `not ready` until status is updated.

sticnarf:
@SSebo Is it a per-region status or a global status? It can't be a global status because we should be able to query new region info concurrently. But a per-region status also has a small problem. If a region splits into halves, two keys which were in the same region are now in two regions. It's not optimal if one blocks the other but is acceptable. It looks both okay to just backoff or use futures.

You can use pd-ctl to transfer the leader of a region or split a region. This will cause region errors. Reference: https://pingcap.com/docs/stable/reference/tools/pd-control/#operator-show--add--remove

SSebo:
sorry for the delay, I 'll continue to work on this no matter ucp is finished.

SSebo:
@sticnarf  hi mentor, I got some ideas:

- When `get_region_by_id/key`, we should compare `error region` and `cached region` instead of just accept `is_retry`. So we need `error_region` pass around. 
- If `error region` is the same as `cached region` and `cached region` is `refreshing`, request should back off. So we need `region.time_stamp` and `region.status` to compare.
- If `error region` is older than `cached region` and `cached region` is `refreshed`, just return `cached region`.
- Otherwise,  we should set `cached region` to `refreshing` status, and fetch a new `region`.

sticnarf:
Sorry I didn't notice your updates.

I think it's unnecessary that you add an extra `ts` to `Region` for comparing versions. `metapb::Region` contains a `version` field. It can be used for comparison.

Passing the error region looks fine to me. An alternative in my mind is to provide an interface to just invalidate the error region.

Issue #132, opened 2020-02-28
UCP: reconnect to PD asynchronously


sticnarf:
## Description

Currently reconnecting to PD will block the event loop. It's better to do it asynchronously.

Fix https://github.com/tikv/client-rust/blob/master/src/pd/cluster.rs#L205

## Difficulty
- Medium

## Score
- 300

## Mentor(s)
- @sticnarf 
- @nrc 

## Recommended Skills
- Async Rust programming

SSebo:
@sticnarf  Could you change the difficulty to Easy? I count the un-picked total easy issues counts, it is less than 200. If this issue is ok to change to easy, I would like to fix it.

you06:
@SSebo The problem has been fixed, but you still need to get 200 points before picking up this. Don't worry about this task being picked up by other, there are not many teams achieved 200 points in tikv org yet.

SSebo:
thanks!

Renkai:
pick up informally

SSebo:
@Renkai you can formally pick this one:smile:

Renkai:
> @Renkai you can formally pick this one😄

@SSebo give up another task is inconvenient😄

you06:
@Renkai `/give-up-challenge` still works

Renkai:
@you06 The description says it is medium😂

PR #141, opened 2020-04-07
Async reconnect

Renkai:
UCP https://github.com/tikv/client-rust/issues/132


src/pd/client.rs
```diff
@@ -152,26 +152,28 @@ impl<KvC: KvConnect + Send + Sync + 'static> PdClient for PdRpcClient<KvC> {
...
                 RetryClient::connect(env, &config.pd_endpoints, security_mgr, config.timeout)
             },
         )
+        .await
     }
 }
 
 impl<KvC: KvConnect + Send + Sync + 'static, Cl> PdRpcClient<KvC, Cl> {
-    pub fn new<MakeKvC, MakePd>(
+    pub async fn new<Fu, MakeKvC, MakePd>(
```
sticnarf, line 168: Could you replace `Fu` with something more meaningful, for example, `PdFut`?

src/pd/retry.rs
```diff
@@ -109,20 +109,28 @@ impl fmt::Debug for RetryClient {
...
 }
 
+#[async_trait]
 impl Reconnect for RetryClient<Cluster> {
     type Cl = Cluster;
 
-    fn reconnect(&self, interval: u64) -> Result<()> {
-        if let Some(cluster) =
+    async fn reconnect(&self, interval: u64) -> Result<()> {
+        if let Some(cluster) = {
```
sticnarf, line 124: I realise that it could make some mess if two `reconnect` happens at the same time. I think it's better to add a gate to make sure that if there is an on-going reconnecting progress, we don't trigger another one.

src/lib.rs
```diff
@@ -100,6 +100,8 @@ extern crate lazy_static;
 extern crate log;
 #[macro_use]
 extern crate prometheus;
+#[macro_use]
+extern crate async_trait;
```
sticnarf, line 104: Prefer to use `use` in where the macros are used rather than `#[macro_use] extern crate` because the macros are not used too often.

src/pd/retry.rs
```diff
@@ -108,24 +115,81 @@ impl fmt::Debug for RetryClient {
     }
 }
 
+struct ShouldReconnect<'a> {
```
sticnarf, line 118: Could you please add some comments to it?

src/pd/retry.rs
```diff
@@ -108,24 +115,81 @@ impl fmt::Debug for RetryClient {
...
-    fn reconnect(&self, interval: u64) -> Result<()> {
-        if let Some(cluster) =
-            self.connection
-                .reconnect(&self.cluster.read().unwrap(), interval, self.timeout)?
-        {
-            *self.cluster.write().unwrap() = cluster;
+    async fn reconnect(&self, interval: u64) -> Result<()> {
+        let connected: MutexLockFuture<bool> = self.connected.lock();
+        let (mut connected, should_connect) = ShouldReconnect::new(connected).await;
+        //if try to reconnect while a previous reconnect is running,
```
sticnarf, line 172:
```suggestion
        // if try to reconnect while a previous reconnect is running,
```

src/pd/retry.rs
sticnarf: Please add a space after `//`.
I actually expect you can document what's the return value of this future...

src/pd/retry.rs
```diff
@@ -108,24 +115,84 @@ impl fmt::Debug for RetryClient {
...
-                .reconnect(&self.cluster.read().unwrap(), interval, self.timeout)?
-        {
-            *self.cluster.write().unwrap() = cluster;
+    async fn reconnect(&self, interval: u64) -> Result<()> {
+        let connected: MutexLockFuture<bool> = self.connected.lock();
+        let (mut connected, should_connect) = ShouldReconnect::new(connected).await;
+        // if try to reconnect while a previous reconnect is running,
+        // the second reconnect will succeed in direct if the first one is succeed
+        if should_connect {
+            *connected = false;
```
nrc, line 178: I'm a bit confused about this, since we hold the lock on `self.connected`, does that mean that this not observable to anyone else? I.e., there is no point in setting this to false then true?

Renkai:
> This is a partial review, because I had some questions but don't completely understand what is going on.
> 
> To clarify, the point of the `ShouldReconnect` future and associated code is so that if a cluster is asked to reconnect twice, then we only connect once? And all subsequent reconnects will block on the first?

Yes

.travis.yml
```diff
@@ -14,6 +14,10 @@ env:
     - RUST_BACKTRACE=1
     - RUSTFLAGS="-D warnings"
 
+matrix:
+  allow_failures:
+    - rust: nightly
+
```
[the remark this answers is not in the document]
Renkai: Nightly CI broken because of https://github.com/AltSysrq/proptest/issues/186 ,can be removed after it is fixed.

src/pd/cluster.rs
```diff
@@ -191,9 +195,10 @@ impl Connection {
     }
 
     // Re-establish connection with PD leader in synchronized fashion.
-    pub fn reconnect(
+    pub async fn reconnect(
         &self,
-        old_cluster: &Cluster,
+        cluster_id: u64,
+        previous_members: &pdpb::GetMembersResponse,
```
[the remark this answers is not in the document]
Renkai: > Why make this change? It seems to expose more of `Cluster`'s implementation.

In previous version, `old_cluster` is actually  a `RwLockReadGuard<Cluster>`,which can not pass through async barrier, and now this method is a async method.

src/pd/retry.rs
Renkai: > I'm a bit confused about this, since we hold the lock on `self.connected`, does that mean that this not observable to anyone else? I.e., there is no point in setting this to false then true?

Process may fail at Line 179 ~ Line 190,then value of connected will stay false, next request can start reconnect instantly.

src/lib.rs
```diff
@@ -100,6 +100,7 @@ extern crate lazy_static;
 extern crate log;
 #[macro_use]
 extern crate prometheus;
+extern crate async_trait;
```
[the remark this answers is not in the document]
Renkai: > Why do we need it? I think it is better to import macros explicitly

@nrc I originally add this line to conform the code style to legacy code, it is actually no need,removed now.

.travis.yml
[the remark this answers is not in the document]
Renkai: > What is this for? Could you add a comment for when it can be removed please?

@nrc That patch is merged now,so these lines can be removed.

src/pd/retry.rs
```diff
@@ -108,24 +115,84 @@ impl fmt::Debug for RetryClient {
...
+                self.connection
+                    .reconnect(id, &members, interval, self.timeout)
+                    .await?
+            } {
+                *self.cluster.write().unwrap() = cluster;
+            }
+            *connected = true;
+            Ok(())
+        } else {
+            Ok(())
```
nrc, line 194: I may be very wrong, but rather than having this logic plus a custom future, could you take the write lock on `self.cluster` on entry to this method and then check it to see if connected? Then reconnect if necessary. You are guaranteed only one task will try to reconnect at a time by the RwLock, and when that task does complete, the other task can just return.

src/pd/cluster.rs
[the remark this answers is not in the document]
Renkai: > I don't understand why we can pass `previous_members`, but not `old_cluster`, istm they are both references with the same lifetimes?

Maybe a move here is better than reference, I changed the parameter to move, `previous_members` is cloned from a field of `old_cluster`, which is just a list of data, `old_cluster` is a more complex struct, refactory it to make it possible to clone will be tricky.I must clone because because `old_cluster` is protected by a `RwLockReadGuard` which cannot pass through a sync barrier(which is `async fn` here )

src/pd/retry.rs
Renkai: > I may be very wrong, but rather than having this logic plus a custom future, could you take the write lock on `self.cluster` on entry to this method and then check it to see if connected? Then reconnect if necessary. You are guaranteed only one task will try to reconnect at a time by the RwLock, and when that task does complete, the other task can just return.

I think @nrc is right,maybe we over engineered in the current version, what's you opinion? @sticnarf

src/pd/retry.rs
```diff
@@ -49,23 +49,25 @@ impl<Cl> RetryClient<Cl> {
             cluster: RwLock::new(cluster),
             connection,
             timeout,
+            last_connected: RwLock::new(Instant::now()),
```
sticnarf, line 52: Could we put the last connected instant inside cluster to save a lock?

Renkai: > Could we put the last connected instant inside cluster to save a lock?

Updated

src/pd/cluster.rs
```diff
@@ -204,22 +211,23 @@ impl Connection {
...
         let cluster = Cluster {
-            id: old_cluster.id,
+            id: cluster_id,
+            last_update,
             client,
             members,
             tso,
         };
-        *self.last_update.write().unwrap() = Instant::now();
+        *self.last_update.write().unwrap() = last_update;
```
sticnarf, line 224: Can we just remove the `last_update` in `Connection`? It looks always identical to the one in `Cluster`.

src/pd/retry.rs
```diff
@@ -102,55 +101,70 @@ impl RetryClient<Cluster> {
...
+        let should_connect = reconnect_begin > write_lock.last_update;
+        if should_connect {
+            if let Some(cluster) = {
+                let (id, members) = {
+                    let read_guard = self.cluster.read().await;
+                    (read_guard.id, read_guard.members.clone())
+                };
+
+                self.connection
+                    .reconnect(id, members, interval, self.timeout)
```
sticnarf, line 134: I think now we can just pass the lock guard in.

src/pd/cluster.rs
```diff
@@ -191,35 +192,38 @@ impl Connection {
     }
 
     // Re-establish connection with PD leader in synchronized fashion.
-    pub fn reconnect(
+    pub async fn reconnect(
         &self,
-        old_cluster: &Cluster,
+        mut cluster_guard: RwLockWriteGuard<'_, Cluster>,
+        cluster_id: u64,
+        previous_members: pdpb::GetMembersResponse,
```
sticnarf, line 199: Once we have the cluster, these two seems unnecessary.

src/pd/cluster.rs
```diff
@@ -315,9 +322,9 @@ impl Connection {
         }
     }
 
-    fn try_connect_leader(
+    async fn try_connect_leader(
         &self,
-        previous: &pdpb::GetMembersResponse,
+        previous: pdpb::GetMembersResponse,
```
sticnarf, line 327: Why not use a reference? It can save us a clone.

Cargo.toml
```diff
@@ -30,6 +30,8 @@ regex = "1"
 serde = "1.0"
 serde_derive = "1.0"
 futures-timer = "3.0"
+async-trait = "0.1.27"
+tokio = { version = "0.2", features = ["sync" ] }
```
sticnarf, line 34:
```suggestion
tokio = { version = "0.2", features = ["sync"] }
```

examples/raw.rs
```diff
@@ -87,7 +87,7 @@ async fn main() -> Result<()> {
     let end = "k2";
     let pairs = client
         .with_key_only(true)
-        .scan((start..=end).to_owned(), 10)
+        .scan((start..=end).to_owned(), 10 as u32)
```
sticnarf, line 90: Why change it?

src/pd/cluster.rs
```diff
@@ -35,8 +35,9 @@ macro_rules! pd_request {
 /// A PD cluster.
 pub struct Cluster {
     pub id: u64,
+    pub last_connected: Instant,
     pub(super) client: pdpb::PdClient,
-    members: pdpb::GetMembersResponse,
+    pub members: pdpb::GetMembersResponse,
```
sticnarf, line 40: Why add it?

src/pd/cluster.rs
```diff
@@ -191,35 +192,39 @@ impl Connection {
     }
 
     // Re-establish connection with PD leader in synchronized fashion.
```
sticnarf, line 194: The comment can be updated. Now it's reconnected asynchronously.

sticnarf: And another FIXME comment in the call site can be updated.

src/pd/cluster.rs
```diff
@@ -191,35 +192,39 @@ impl Connection {
...
+        if cluster_guard.last_connected.elapsed() < Duration::from_secs(interval) {
             // Avoid unnecessary updating.
-            return Ok(None);
+            return Ok(());
         }
 
+        let cluster_id = cluster_guard.id;
+        let previous_members = &cluster_guard.members;
+
         warn!("updating pd client, blocking the tokio core");
```
sticnarf, line 209: should be deleted

src/pd/cluster.rs
```diff
@@ -191,35 +192,39 @@ impl Connection {
...
         interval: u64,
         timeout: Duration,
-    ) -> Result<Option<Cluster>> {
-        if self.last_update.read().unwrap().elapsed() < Duration::from_secs(interval) {
+    ) -> Result<()> {
+        if cluster_guard.last_connected.elapsed() < Duration::from_secs(interval) {
             // Avoid unnecessary updating.
-            return Ok(None);
+            return Ok(());
         }
```
sticnarf, line 204: I don't think we need the interval and the check here. If it's a repeated reconnecting, it will be rejected by checking the last updated time. What do you think?

examples/raw.rs
Renkai: > Why change it?

May be a previous in progress need change it,can restore to the origin version now

src/pd/cluster.rs
Renkai: > I don't think we need the interval and the check here. If it's a repeated reconnecting, it will be rejected by checking the last updated time. What do you think?

Yes, I think this logic is out of date, removed now.

src/pd/retry.rs
```diff
@@ -110,20 +111,23 @@ impl fmt::Debug for RetryClient {
...
         let reconnect_begin = Instant::now();
         let write_lock = self.cluster.write().await;
-        // If last_update is larger or equal than reconnect_begin,
+        // If `last_connected + interval_sec` is larger or equal than reconnect_begin,
         // a concurrent reconnect is just succeed when this thread trying to get write lock
-        let should_connect = reconnect_begin > write_lock.last_connected;
+        let should_connect = reconnect_begin
+            > write_lock
+                .last_connected
+                .add(Duration::from_secs(interval_sec));
```
sticnarf, line 130: You can simply use `+`

src/pd/retry.rs
```diff
@@ -102,55 +101,61 @@ impl RetryClient<Cluster> {
...
+    async fn reconnect(&self, interval_sec: u64) -> Result<()> {
+        let reconnect_begin = Instant::now();
+        let write_lock = self.cluster.write().await;
+        // If `last_connected + interval_sec` is larger or equal than reconnect_begin,
+        // a concurrent reconnect is just succeed when this thread trying to get write lock
+        let should_connect =
+            reconnect_begin > write_lock.last_connected + Duration::from_secs(interval_sec);
+        if should_connect {
+            self.connection.reconnect(write_lock, self.timeout).await?;
+            Ok(())
```
sticnarf, line 130: What about writing `Ok(())` only once out of the if branch?

sticnarf:
@nrc PTAL again

PR #143, opened 2020-05-04
Feat/txn scan iter

weihanglo:
UCP #131

## What?

Implement transactional scan request for Transaction::scan.

## How

See https://github.com/tikv/client-rust/pull/133 and https://github.com/tikv/client-rust/pull/133#discussion_r397444905.

### DIFF
--- a/examples/transaction.rs
+++ b/examples/transaction.rs
@@ -21,7 +21,7 @@ async fn get(client: &Client, key: Key) -> Option<Value> {
 
 async fn scan(client: &Client, range: impl Into<BoundRange>, limit: u32) {
     let mut txn = client.begin().await.expect("Could not begin a transaction");
-    txn.scan(range, limit)
+    txn.scan(range, limit, false)
         .await
         .expect("Could not scan key-value pairs in rnage")
         .for_each(|pair| println!("{:?}", pair));
--- a/examples/transaction.rs
+++ b/examples/transaction.rs
@@ -3,9 +3,7 @@
 mod common;
 
 use crate::common::parse_args;
-use futures::prelude::*;
-use std::ops::RangeBounds;
-use tikv_client::{Config, Key, KvPair, TransactionClient as Client, Value};
+use tikv_client::{BoundRange, Config, Key, KvPair, TransactionClient as Client, Value};
 
 async fn puts(client: &Client, pairs: impl IntoIterator<Item = impl Into<KvPair>>) {
     let mut txn = client.begin().await.expect("Could not begin a transaction");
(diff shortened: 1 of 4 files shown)


weihanglo:
@nrc @sticnarf 
This PR implements as https://github.com/tikv/client-rust/pull/133#discussion_r397444905 described. Need some review. Thanks!

examples/transaction.rs
```diff
@@ -21,23 +19,13 @@ async fn get(client: &Client, key: Key) -> Option<Value> {
     txn.get(key).await.expect("Could not get value")
 }
 
-// Ignore a spurious warning from rustc (https://github.com/rust-lang/rust/issues/60566).
-#[allow(unused_mut)]
-async fn scan(client: &Client, range: impl RangeBounds<Key>, mut limit: usize) {
+async fn scan(client: &Client, range: impl Into<BoundRange>, mut _limit: usize) {
```
nrc, line 22: Why cut out the limit functionality here?

examples/transaction.rs
```diff
@@ -21,23 +19,13 @@ async fn get(client: &Client, key: Key) -> Option<Value> {
...
-                false
-            } else {
-                limit -= 1;
-                true
-            })
-        })
-        .for_each(|pair| future::ready(println!("{:?}", pair)))
-        .await;
+        .await
+        .unwrap()
```
nrc, line 26: Could you use `expect` instead of `unwrap` please?

weihanglo: Fixed 35eeceb

examples/transaction.rs
weihanglo: The `limit` parameter has implemented at 963e3b3.

Here we've got some questions:

1. What's the maximum limit of a transactional scan? Should it be checked as what raw client does?

https://github.com/tikv/client-rust/blob/65602f7a45c9807a4119fc28488519a4ac9b8e4c/src/raw/client.rs#L249-L252

2. How should the caller pass `key_only` to client? Seems that `raw` client has a field to indicate this matter.

https://github.com/tikv/client-rust/blob/65602f7a45c9807a4119fc28488519a4ac9b8e4c/src/raw/client.rs#L15-L19

nrc: 1: Yes.

2: yes, transactional scan also has that field: https://github.com/pingcap/kvproto/blob/master/proto/kvrpcpb.proto#L50

weihanglo: 1. If I understand it correctly, just declare a constant `const MAX_TXN_KV_SCAN_LIMIT: u32 = 10240;` under `src/transaction/transaction.rs`and check the limit.
2. Yes, I know there is a field on Protobuf definition, but I am not quite sure how to implement it within transactional client. Should I add a field `key_only` on txn client, or `Transaction` itself?

weihanglo:
@nrc 
Thanks for you review! `key_only` parameter for scan has be implemented on 731cb10.

weihanglo:
@nrc 
Thanks for the review! `key_only` parameter for scan has be implemented on 731cb10.

examples/transaction.rs
```diff
@@ -21,23 +19,13 @@ async fn get(client: &Client, key: Key) -> Option<Value> {
...
-            } else {
-                limit -= 1;
-                true
-            })
-        })
-        .for_each(|pair| future::ready(println!("{:?}", pair)))
-        .await;
+    txn.scan(range, limit, false)
+        .await
+        .expect("Could not scan key-value pairs in rnage")
```
nrc, line 26:
```suggestion
        .expect("Could not scan key-value pairs in range")
```

weihanglo:
> For the limit, this request will get sent to n regions and so the real limit will end up being n*limit. So, you must do something like we do for raw scan and count the returned results and stop when we hit limit.

I am a bit stuck here.  I am thinking of two approaches:

1.  Check the limit before making any unnecessary gRPC request. Not sure would it be possible with `DispatchHook` or other smiliar mechanism.
2. Just send all `n * limit` request, and then use `Iter::take` to return the actual amount of pairs. This one is quite intuitive but more wasteful on resources.

Besides, I still cannot find the reference implementation of raw scan on this feature. Could you shed some light on any details? Thanks @nrc !

nrc:
> Besides, I still cannot find the reference implementation of raw scan on this feature

Hmm, it seems not to exist :-) Either it has been removed or I am mis-remembering and it is in TiDB or elsewhere, not in this client. In any case, I think it means we don't need it here!

nrc:
@sticnarf could you take a look please?

Issue #111, opened 2019-08-28
Add resolve lock lite


sticnarf:
Normal resolve lock scans the whole region for locks. With resolve lock lite, we resolve only the specified locks. This can reduce TiKV's load when there are a lot of conflicts

PR #145, opened 2020-05-24
Lite resolve lock

Renkai:
UCP https://github.com/tikv/client-rust/issues/111


src/transaction/lock.rs
```diff
@@ -21,67 +21,79 @@ pub async fn resolve_locks(
...
+        })
+        .into_iter()
+        .map(|(k, v)| (k, v.collect()))
+        .collect();
+
+    for ((_region_ver_id, lock_version), locks) in grouped {
+        let is_large_txn = locks.iter().any(|lock| lock.txn_size >= 16);
+
+        let commit_version =
+            requests::new_cleanup_request(locks.get(0).unwrap().primary_key.clone(), lock_version)
```
sticnarf, line 66: Why is only the first lock in a region cleaned up?

Renkai: > Why is only the first lock in a region cleaned up?

https://github.com/Renkai/client-rust/blob/3ab4252e1dce5d1a32042ca7fb028329a5a1426e/src/transaction/lock.rs#L49

In the original version, every lock_version will do a region clean up only once. Here the locks are already grouped by lock version, so only the first grouped lock do a region clean up behave exactly the same as original.

sticnarf: Ah, I got what you mean. The idea looks good. Yet there is something to refine.
For every lock with the same `(region, lock_version)`, the `txn_size` is the same. So actually we don't need to save the keys of the large transactions.
And there seems to be lots of clones, could you try to reduce clones?

src/transaction/lock.rs
```diff
@@ -21,67 +21,77 @@ pub async fn resolve_locks(
...
+
+    struct RichLock {
+        primary_key: Key,
+        region: Region,
+        lock_version: u64,
+        txn_size: u64,
+    }
+
+    let mut expired_locks = vec![];
+    for lock in locks {
```
sticnarf, line 34: I think we can iterate only once. For every expired lock, check if it's big enough to use the lite way, store them in a map with `(Region, lock_version)` as the key. The value is only the primary key for the normal way, and primary key plus all the locks in the region for the lite way. Then use different ways to handle them.

First collecting expired and then using `group by` look too heavy for this requirement.

src/transaction/lock.rs
```diff
@@ -21,67 +22,76 @@ pub async fn resolve_locks(
...
-        if !expired {
+        if expired {
+            let primary_key: Key = lock.primary_lock.into();
+            let region: Region = pd_client.region_for_key(&primary_key).await?;
+
+            grouped
+                .entry((region.ver_id(), lock.lock_version))
+                .or_insert(ResolveInfo {
+                    txn_size: lock.txn_size,
+                    region: region.clone(),
```
sticnarf, line 47: This clone looks unnecessary

src/transaction/lock.rs
```diff
@@ -21,67 +22,76 @@ pub async fn resolve_locks(
...
 ) -> Result<bool> {
     let ts = pd_client.clone().get_timestamp().await?;
+    type LockVersion = u64;
     let mut has_live_locks = false;
-    let expired_locks = locks.into_iter().filter(|lock| {
+
+    struct ResolveInfo {
+        txn_size: u64,
+        region: Region,
+        keys: Vec<Key>,
```
sticnarf, line 31: I think we can save primary key only once in `ResolveInfo`. And the `keys` are used to save the lock keys if the transaction is small enough.

src/transaction/lock.rs
```diff
@@ -21,67 +22,76 @@ pub async fn resolve_locks(
...
+    let mut grouped: HashMap<(RegionVerId, LockVersion), ResolveInfo> = HashMap::new();
+
+    for lock in locks {
         let expired = ts.physical - Timestamp::from_version(lock.lock_version).physical
             >= lock.lock_ttl as i64;
-        if !expired {
+        if expired {
+            let primary_key: Key = lock.primary_lock.into();
+            let region: Region = pd_client.region_for_key(&primary_key).await?;
+
```
sticnarf, line 42: I recommend checking `txn_size` here.

src/transaction/lock.rs
```diff
@@ -21,67 +22,74 @@ pub async fn resolve_locks(
...
-                let commit_version = requests::new_cleanup_request(primary_key, lock.lock_version)
-                    .execute(pd_client.clone())
-                    .await?;
-                commit_versions.insert(lock.lock_version, commit_version);
-                commit_version
-            }
-        };
+    for ((_region_ver_id, lock_version), resolve_info) in grouped {
+        let commit_version =
+            requests::new_cleanup_request(resolve_info.primary_key.clone(), lock_version)
```
sticnarf, line 65: You can replace this key out instead of `clone`.

src/transaction/lock.rs
```diff
@@ -21,67 +22,74 @@ pub async fn resolve_locks(
...
+                .execute(pd_client.clone())
+                .await?;
 
-        let cleaned_region = resolve_lock_with_retry(
-            lock.key.into(),
-            lock.lock_version,
+        let _cleaned_region = resolve_lock_with_retry(
+            &resolve_info.region,
+            &resolve_info.keys,
```
sticnarf, line 71: You can change it to take the ownership and move it in to avoid cloning in the later processes.

Renkai: > You can change it to take the ownership and move it in to avoid cloning in the later processes.

cloning in the later process seems can't be reduced since clone happens in a loop.

sticnarf: If you just take a `Vec<Key>`, I think we can write `req.set_keys(keys.into_iter().map(Into::into).collect());` [here](https://github.com/tikv/client-rust/pull/145/files#diff-d01850033b0d2fc084df1829a7789724R264)

Renkai: > If you just take a `Vec<Key>`, I think we can write `req.set_keys(keys.into_iter().map(Into::into).collect());` [here](https://github.com/tikv/client-rust/pull/145/files#diff-d01850033b0d2fc084df1829a7789724R264)

If `new_resolve_lock_request` take a `Vec<Key>`, `value moved here, in previous iteration of loop` will be generated at https://github.com/tikv/client-rust/pull/145/files#diff-49f989cfce4dadfa41d7d1bf2548edd7R99

sticnarf: Ah sorry, I got what's the limitation here.

Cargo.toml
```diff
@@ -30,6 +30,7 @@ regex = "1"
 serde = "1.0"
 serde_derive = "1.0"
 futures-timer = "3.0"
+itertools = "0.9.0"
```
sticnarf, line 33: This is now not used.

src/transaction/lock.rs
```diff
@@ -1,14 +1,15 @@
...
 use crate::request::KvRequest;
 use crate::{ErrorKind, Key, Result, Timestamp};
 
 use kvproto::kvrpcpb;
-use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 const RESOLVE_LOCK_RETRY_LIMIT: usize = 10;
 
+use std::collections::HashMap;
```
sticnarf, line 11: Please combine this `use` with other std uses.

src/transaction/lock.rs
```diff
@@ -21,67 +22,72 @@ pub async fn resolve_locks(
...
-        };
+    for ((_region_ver_id, lock_version), resolve_info) in grouped {
+        let commit_version = requests::new_cleanup_request(resolve_info.primary_key, lock_version)
+            .execute(pd_client.clone())
+            .await?;
 
-        let cleaned_region = resolve_lock_with_retry(
-            lock.key.into(),
-            lock.lock_version,
+        let _cleaned_region = resolve_lock_with_retry(
```
sticnarf, line 67: Now cleaned region has no effect I think.

src/transaction/lock.rs
```diff
@@ -21,67 +22,72 @@ pub async fn resolve_locks(
...
 }
 
 async fn resolve_lock_with_retry(
-    key: Key,
+    region: &Region,
+    keys: &Vec<Key>,
     start_version: u64,
     commit_version: u64,
     pd_client: Arc<impl PdClient>,
 ) -> Result<RegionVerId> {
```
sticnarf, line 86: I think now we needn't return `RegionVerId`.

src/transaction/lock.rs
```diff
@@ -21,67 +22,72 @@ pub async fn resolve_locks(
...
     // TODO: Add backoff
```
sticnarf, line 87: Would you mind add basic backoff support?

Renkai: > Would you mind add basic backoff support?

Updated

src/transaction/lock.rs
```diff
@@ -21,81 +21,92 @@ pub async fn resolve_locks(
...
     start_version: u64,
     commit_version: u64,
     pd_client: Arc<impl PdClient>,
-) -> Result<RegionVerId> {
-    // TODO: Add backoff
+) -> Result<()> {
     let mut error = None;
+    let mut back_off = NoJitterBackoff::new(10, 1000, RESOLVE_LOCK_RETRY_LIMIT as u32);
+
     for _ in 0..RESOLVE_LOCK_RETRY_LIMIT {
```
sticnarf, line 89: WIth the backoff providing retry limit (`next_delay_duration` returns `None` if the limit is exceeded), I think maybe we can use `loop` and `break` to avoid repeated reference to RESOLVE_LOCK_RETRY_LIMIT

src/transaction/lock.rs
```diff
@@ -21,81 +21,92 @@ pub async fn resolve_locks(
...
+    region: &Region,
+    keys: &Vec<Key>,
     start_version: u64,
     commit_version: u64,
     pd_client: Arc<impl PdClient>,
-) -> Result<RegionVerId> {
-    // TODO: Add backoff
+) -> Result<()> {
     let mut error = None;
+    let mut back_off = NoJitterBackoff::new(10, 1000, RESOLVE_LOCK_RETRY_LIMIT as u32);
```
sticnarf, line 87: We treat "backoff" as a single word, do not add underline between them.
And I recommend making `NoJitterBackoff::new(10, 1000, RESOLVE_LOCK_RETRY_LIMIT as u32)` a const.

src/transaction/lock.rs
```diff
@@ -21,81 +21,92 @@ pub async fn resolve_locks(
...
+    let mut grouped: HashMap<(RegionVerId, LockVersion), ResolveInfo> = HashMap::new();
+
+    for lock in locks {
         let expired = ts.physical - Timestamp::from_version(lock.lock_version).physical
             >= lock.lock_ttl as i64;
-        if !expired {
+        if expired {
+            let primary_key: Key = lock.primary_lock.into();
+            let region: Region = pd_client.region_for_key(&primary_key).await?;
+            let is_small_txn = lock.txn_size < 16;
```
sticnarf, line 41: Could you change the hardcoded `16` to a const?

src/transaction/lock.rs
Renkai: > We treat "backoff" as a single word, do not add underline between them.
> And I recommend making `NoJitterBackoff::new(10, 1000, RESOLVE_LOCK_RETRY_LIMIT as u32)` a const.

`NoJitterBackoff::next_delay_duration` is a `&mut self` method, maybe const is not suitable.

sticnarf: That doesn't matter. Actually we can define `const RESOLVE_LOCK_BACKOFF: NoJitterBackoff =
    NoJitterBackoff::new(10, 1000, RESOLVE_LOCK_RETRY_LIMIT as u32);` and use `let mut backoff = RESOLVE_LOCK_BACKOFF;`

Renkai:
block on https://github.com/tikv/client-rust/issues/114

paused

PR #149, opened 2020-06-11
Remove compat

### LINKED ISSUE #148: Upgrade gRPC and remove futures compat code
GRPC-rs now supports futures 0.3 so if we upgrade then we shouldn’t need the compatibility shims to go from 0.1 to 0.3.

longfangsong:
Signed-off-by: longfangsong longfangsong@icloud.com

Upgrade grpcio and remove the compaction code.

This should fix #148.

~~Note because of some dependency problems, we have to use my own fork of kvproto. And because of [cargo #8258](https://github.com/rust-lang/cargo/issues/8258), we cannot use   1.43+ version of rust to build the project.~~
These limitations have been removed after these crates and tools are updated into new version.


sticnarf:
Looks great.

@nrc 
If TiKV changes to use futures 0.3, it's natural that kvproto upgrades its dependency. It's nice to have but it's not an urgent work. I don't expect it to happen very soon.

Anyway, instead of using @longfangsong's personal fork, can we add a branch using futures 0.3 to tikv/kvproto before landing this PR?

longfangsong:
> Anyway, instead of using @longfangsong's personal fork, can we add a branch using futures 0.3 to tikv/kvproto before landing this PR?

I prefer this solution too.

nrc:
@longfangsong could you rebase this PR please? Do we have a branch on kvproto we can use?

longfangsong:
> Do we have a branch on kvproto we can use?

I have a pr here: [kvprotp/633](https://github.com/pingcap/kvproto/pull/633). But it hasn't get merged yet.

longfangsong:
Now we are using a branch on pingcap/kvproto.
So I think this pr is ready for merge.
@nrc @sticnarf

src/kv_client/mod.rs
```diff
@@ -0,0 +1,127 @@
+// Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
```
nrc, line 1: I think maybe this file is a rebasing error? I think it has moved to one of the crates

Issue #151, opened 2020-07-06
Refactoring: remove Value type


nrc:
Either just use `byte[]` or create an alias `type Value = byte[];` since we do not enforce any additional invariants for `Value` over `byte[]` and to simplify the code. This will also require changes to the `KeyValue` types, etc.

cc #75 #99

PR #152, opened 2020-07-07
set Value type to an alias of Vec<u8>

ekexium:
set Value type to an alias of Vec<u8> since the Original Value implementation does not do more than Vec<u8>. (#151 )

Signed-off-by: ekexium <ekexium@gmail.com>

### DIFF
--- a/src/proptests/raw.rs
+++ b/src/proptests/raw.rs
@@ -23,7 +23,7 @@ proptest! {
         ).unwrap();
 
         match out_value {
-            None =>assert!(pair.value().is_empty()),
+            None => assert!(pair.value().is_empty()),
             Some(out) => assert_eq!(Value::from(pair.value().clone()), out)
         }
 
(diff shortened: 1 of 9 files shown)


nrc:
> the Original Value implementation does not do more than Vec

FTR, my understanding is that the type `Value` offers no additional/different invariants to the type `Vec<u8>` and thus should not be a separate type.

src/proptests/raw.rs
```diff
@@ -21,7 +21,11 @@ proptest! {
         let out_value = block_on(
             client.get(pair.key().clone())
         ).unwrap();
-        assert_eq!(Some(Value::from(pair.value().clone())), out_value);
+
+        match out_value {
+            None =>assert!(pair.value().is_empty()),
```
sticnarf, line 26: It seems that the project is not fully reformatted?

ekexium: This piece of code inside a macro is not automatically formatted by rustfmt.

Issue #153, opened 2020-07-08
Refactoring: replace ColumnFamily type with an enum


nrc:
Only a few distinguished names are accepted as column families, so we should make those an enum: Default, Lock, Write, Version ("versioncf").

sticnarf:
Closed by #154.

PR #154, opened 2020-07-08
change ColumnFamily to enum

ekexium:
#153 

Signed-off-by: ekexium <ekexium@gmail.com>

### DIFF
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -43,11 +43,11 @@ impl Client {
     /// supplied column family constraint. The original `Client` can still be used.
     ///
     /// ```rust,no_run
-    /// # use tikv_client::{Config, RawClient};
+    /// # use tikv_client::{Config, RawClient, ColumnFamily};
     /// # use futures::prelude::*;
     /// # use std::convert::TryInto;
     /// # futures::executor::block_on(async {
-    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf("write".try_into().unwrap());
+    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf(ColumnFamily::Write);
     /// let get_request = client.get("foo".to_owned());
     /// # });
     /// ```
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -6,7 +6,6 @@ use crate::{
     Result, Value,
 };
 
-use std::convert::TryInto;
 use std::{sync::Arc, u32};
 
 const MAX_RAW_KV_SCAN_LIMIT: u32 = 10240;
@@ -46,18 +45,18 @@ impl Client {
     /// ```rust,no_run
     /// # use tikv_client::{Config, RawClient};
     /// # use futures::prelude::*;
+    /// # use std::convert::TryInto;
     /// # futures::executor::block_on(async {
-    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf("write").unwrap();
+    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf("write".try_into().unwrap());
     /// let get_request = client.get("foo".to_owned());
     /// # });
     /// ```
-    pub fn with_cf(&self, cf: &str) -> Result<Client> {
-        let cf = cf.try_into()?;
-        Ok(Client {
+    pub fn with_cf(&self, cf: ColumnFamily) -> Client {
+        Client {
             rpc: self.rpc.clone(),
             cf: Some(cf),
             key_only: self.key_only,
-        })
+        }
     }
 
     /// Set the `key_only` option of requests.
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -6,6 +6,7 @@ use crate::{
     Result, Value,
 };
 
+use std::convert::TryInto;
 use std::{sync::Arc, u32};
 
 const MAX_RAW_KV_SCAN_LIMIT: u32 = 10240;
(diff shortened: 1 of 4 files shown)


src/raw/mod.rs
```diff
@@ -42,18 +40,34 @@ mod requests;
 /// **But, you should not need to worry about all this:** Many functions which accept a
 /// `ColumnFamily` accept an `Into<ColumnFamily>`, which means all of the above types can be passed
 /// directly to those functions.
-#[derive(Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
-pub struct ColumnFamily(String);
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```
nrc, line 43: I don't think we should implement Ord and PartialOrd, unless it is really necessary.

src/raw/mod.rs
```diff
@@ -42,18 +40,34 @@ mod requests;
...
+}
 
 impl<T: Into<String>> From<T> for ColumnFamily {
     fn from(i: T) -> ColumnFamily {
-        ColumnFamily(i.into())
+        match i.into().as_str() {
+            "default" => ColumnFamily::Default,
+            "lock" => ColumnFamily::Lock,
+            "write" => ColumnFamily::Write,
+            "versioncf" => ColumnFamily::Version,
```
nrc, line 57: the string representation for the Version CF is `"ver_default"` (sorry, that was my mistake). The variant should probably be named `VersionDefault`.

src/raw/mod.rs
```diff
@@ -42,18 +40,34 @@ mod requests;
...
+            _ => unreachable!(),
+        }
     }
 }
 
-impl fmt::Display for ColumnFamily {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        self.0.fmt(f)
+impl ToString for ColumnFamily {
```
nrc, line 63: You should always implement Display rather than ToString - ToString is automatically derived for all types which implement Display, but not vice-versa.

src/raw/mod.rs
```diff
@@ -42,18 +40,34 @@ mod requests;
...
-pub struct ColumnFamily(String);
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum ColumnFamily {
+    Default,
+    Lock,
+    Write,
+    Version,
+}
 
 impl<T: Into<String>> From<T> for ColumnFamily {
```
nrc, line 51: I would implement `TryFrom` rather than `From` for `String` and `&str` (not `T: Into<String>`). Then we can return an error rather than panicking if the string is an incorrect name,

src/raw/client.rs
```diff
@@ -46,16 +47,17 @@ impl Client {
...
     /// # use futures::prelude::*;
     /// # futures::executor::block_on(async {
-    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf("write");
+    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf("write").unwrap();
     /// let get_request = client.get("foo".to_owned());
     /// # });
     /// ```
-    pub fn with_cf(&self, cf: impl Into<ColumnFamily>) -> Client {
-        Client {
+    pub fn with_cf(&self, cf: &str) -> Result<Client> {
```
nrc, line 54: Lets change `with_cf` to take `ColumnFamily` and let the caller handle the conversion. The example will then be:
` let client = RawClient::new(Config::default()).await.unwrap().with_cf(ColumnFamily::Write);`

src/raw/mod.rs
```diff
@@ -31,43 +35,57 @@ mod requests;
...
-            ColumnFamily::Default => "default".to_owned(),
-            ColumnFamily::Lock => "lock".to_owned(),
-            ColumnFamily::Write => "write".to_owned(),
-            ColumnFamily::Version => "versioncf".to_owned(),
-        }
+impl TryFrom<String> for ColumnFamily {
+    type Error = Error;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        TryFrom::try_from(value.as_str())
```
nrc, line 73:
```suggestion
        TryFrom::try_from(&*value)
```

src/raw/mod.rs
```diff
@@ -31,43 +35,57 @@ mod requests;
...
+        write!(
+            f,
+            "{}",
+            match self {
+                ColumnFamily::Default => "default",
+                ColumnFamily::Lock => "lock",
+                ColumnFamily::Write => "write",
+                ColumnFamily::VersionDefault => "ver_default",
+            }
+        )
```
nrc, line 88: style nit: I would put the `match` on the 'outside' rather than inside the `write!`, you can use `f.write_str("foo")` for each case, which is marginally simpler than using `write!`

src/raw/client.rs
```diff
@@ -45,15 +45,16 @@ impl Client {
     /// ```rust,no_run
     /// # use tikv_client::{Config, RawClient};
     /// # use futures::prelude::*;
+    /// # use std::convert::TryInto;
     /// # futures::executor::block_on(async {
-    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf("write");
+    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf("write".try_into().unwrap());
```
nrc, line 50:
```suggestion
    /// let client = RawClient::new(Config::default()).await.unwrap().with_cf(ColumnFamily::Write);
```

nrc:
PTAL @sticnarf

PR #158, opened 2020-07-23
Refactor PD crate to make it more independent

nrc:
This PR refactors the PD crate and pd/retry to make the PD crate lower level and more independent. That in turn let me move some stuff out of the common crate. The tricky bit was refactoring retry to work with async/await. I have converted a bunch of code from futures to async/await along the way.

PTAL @sticnarf @ekexium


ekexium:
Looks good.

PR #159, opened 2020-09-07
Add a mock tikv and a mock pd for easier integration tests (first step)

ekexium:
Changes:

- A mock TiKV that implments a series of raw operations: get, put, delete and batched {get, put, delete}.
- A mock PD.

Problems:
- TimestampOracle was not designed to stop. It might be the reason to prevent pd_server from gracefully shutting down.


src/mock/mock_kv.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```
nrc, line 1: Can you remove the changes to src/mock? I think they are orthogonal to the mock-tikv changes.

mock-tikv/src/server.rs
```diff
@@ -0,0 +1,508 @@
...
+        sink: grpcio::UnarySink<kvproto::kvrpcpb::RawBatchGetResponse>,
+    ) {
+        let mut resp = kvproto::kvrpcpb::RawBatchGetResponse::default();
+        resp.set_pairs(self.inner.raw_batch_get(req.get_keys()));
+        let f = sink
+            .success(resp)
+            .map_err(move |e| panic!("failed to reply {:?}: {:?}", req, e))
+            .map(|_| ());
+        ctx.spawn(f)
+    }
```
nrc, line 201: Can you use a macro for these implementations?

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    pub fn raw_get(&self, key: &[u8]) -> Vec<u8> {
+        let data = self.data.read().unwrap();
+        data.get(key).unwrap_or(&vec![]).to_vec()
+    }
+
+    pub fn raw_batch_get(&self, keys: &[Vec<u8>]) -> Vec<KvPair> {
+        let data = self.data.read().unwrap();
+        let mut pairs = vec![];
+        for key in keys {
```
nrc, line 33: Use `keys.iter().map(...).collect()` rather than a `for` loop here.

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,76 @@
...
+        data.remove(key).ok_or(())
+    }
+
+    // if any of the key does not exist, return non-existent keys
+    pub fn raw_batch_delete<'a>(&self, keys: &'a [Vec<u8>]) -> Result<(), Vec<&'a str>> {
+        let mut data = self.data.write().unwrap();
+        let mut non_exist_keys = vec![];
+        keys.iter()
+            .filter(|&key| !data.contains_key(key))
+            .for_each(|key| non_exist_keys.push(std::str::from_utf8(key).unwrap()));
```
nrc, line 66: I think you can use map and collect rather than for_each and push

src/raw/mod.rs
```diff
@@ -16,7 +16,7 @@ use std::{convert::TryFrom, fmt};
 use tikv_client_common::errors::ErrorKind::ColumnFamilyError;
 
 mod client;
-mod requests;
+pub(crate) mod requests;
```
nrc, line 19: Is this change still needed?

mock-tikv/src/pd.rs
```diff
@@ -0,0 +1,313 @@
...
+impl Pd for MockPd {
+    fn get_members(
+        &mut self,
+        ctx: ::grpcio::RpcContext,
+        req: GetMembersRequest,
+        sink: ::grpcio::UnarySink<GetMembersResponse>,
+    ) {
+        let mut resp = GetMembersResponse::default();
+        resp.set_header(ResponseHeader::default());
+        let mut member = Member::default();
```
nrc, line 62: Are the fields public? I think it is easier to use a struct literal rather than default and setters if they are. You can use the following pattern if you only want to initialize some fields:
```
let f = Foo {
    a: whatever,
    b: something,
    ..Foo::default()
};
```

mock-tikv/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
+mod pd;
+mod server;
+mod store;
+
+pub use pd::{start_mock_pd_server, MockPd, MOCK_PD_PORT};
+pub use server::{start_mock_tikv_server, MockTikv, MOCK_TIKV_PORT};
+pub use store::KvStore;
+
+#[macro_export]
```
nrc, line 10: Could you add a comment describing what the macro does please?

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub fn raw_batch_get(&self, keys: &[Vec<u8>]) -> Vec<KvPair> {
+        let data = self.data.read().unwrap();
+        keys.iter()
+            .map(|key| {
+                let mut pair = KvPair::default();
+                pair.set_value(data.get(key).unwrap_or(&vec![]).to_vec());
+                pair.set_key(key.to_vec());
+                pair
+            })
+            .collect::<Vec<_>>()
```
nrc, line 41: Shouldn't need the type annotation here

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+impl KvStore {
+    pub fn new() -> KvStore {
+        KvStore {
+            data: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    pub fn raw_get(&self, key: &[u8]) -> Vec<u8> {
+        let data = self.data.read().unwrap();
+        data.get(key).unwrap_or(&vec![]).to_vec()
```
nrc, line 29:
```suggestion
        data.get(key).map(|s| s.to_vec()).unwrap_or_else(Vec::new)
```

Similarly elsewhere

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    pub fn raw_batch_get(&self, keys: &[Vec<u8>]) -> Vec<KvPair> {
```
nrc, line 32: I think you can take `keys: Vec<Vec<u8>>` then use `into_iter` and `key` instead of `key.to_vec()`.

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+            .map(|key| {
+                let mut pair = KvPair::default();
+                pair.set_value(data.get(key).unwrap_or(&vec![]).to_vec());
+                pair.set_key(key.to_vec());
+                pair
+            })
+            .collect::<Vec<_>>()
+    }
+
+    pub fn raw_put(&self, key: &[u8], value: &[u8]) {
```
nrc, line 44: I think you take key and value by value

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+        let mut data = self.data.write().unwrap();
+        *data.entry(key.to_vec()).or_default() = value.to_vec();
```
nrc, line 46: can you use `data.insert` here and below?

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    pub fn raw_batch_put(&self, pairs: &[KvPair]) {
+        let mut data = self.data.write().unwrap();
+        for pair in pairs {
+            *data.entry(pair.get_key().to_vec()).or_default() = pair.get_value().to_vec();
+        }
+    }
+
+    // if success, return the key deleted
+    pub fn raw_delete(&self, key: &[u8]) -> Result<Vec<u8>, ()> {
```
nrc, line 57: Return an Option rather than a Result

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+        data.remove(key).ok_or(())
+    }
+
+    // if any of the key does not exist, return non-existent keys
+    pub fn raw_batch_delete<'a>(&self, keys: &'a [Vec<u8>]) -> Result<(), Vec<&'a str>> {
+        let mut data = self.data.write().unwrap();
+        let non_exist_keys = keys
+            .iter()
+            .filter(|&key| !data.contains_key(key))
+            .map(|key| std::str::from_utf8(key).unwrap())
```
nrc, line 68: Why use strings here rather than `Vec<u8>`?

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+            .collect::<Vec<_>>();
+        if !non_exist_keys.is_empty() {
+            Err(non_exist_keys)
+        } else {
+            keys.iter().for_each(|key| {
```
nrc, line 73: Can you rewrite the function so that you only iterate over `keys` once?

mock-tikv/src/store.rs
ekexium: Emm makes sense

mock-tikv/src/store.rs
ekexium: I don't come up with an elegant way to do this.

nrc: It's a bit of a novelty question really. My attempt is:

```
        let mut data = self.data.write().unwrap();
        let non_exist: Vec<_> = keys.into_iter().filter_map(|k| data.remove(&k).xor(Some(k))).collect();
        if non_exist.is_empty() {
            Ok(())
        } else {
            Err(non_exist)
        }
```

Note that it has slightly different semantics - it removes any keys which are in the batch, whereas yours doesn't delete anything if any key is missing. I'm not actually sure what TiKV does, but I suspect it will not matter in practice (given how MySQL `DELETE` works).

ekexium: Yeah if we don't want to gurantee the atomicity of this operation, we can do it.

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    pub fn raw_put(&self, key: Vec<u8>, value: Vec<u8>) {
+        let mut data = self.data.write().unwrap();
+        data.insert(key, value);
+    }
+
+    pub fn raw_batch_put(&self, pairs: &[KvPair]) {
+        let mut data = self.data.write().unwrap();
+        for pair in pairs {
```
nrc, line 55: Could use `data.extend(pairs.iter().map(...))`. Could also take `pairs` by value to avoid the clone.

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,79 @@
...
+        let mut data = self.data.write().unwrap();
+        data.remove(key)
+    }
+
+    // if any of the key does not exist, return non-existent keys; delete other keys
+    pub fn raw_batch_delete(&self, keys: &[Vec<u8>]) -> Result<(), Vec<Vec<u8>>> {
+        let mut data = self.data.write().unwrap();
+        let non_exist: Vec<_> = keys
+            .iter()
+            .filter_map(|k| data.remove(k).xor(Some(k.to_vec())))
```
nrc, line 71: Can you take keys by value and avoid the clone?

mock-tikv/src/server.rs
```diff
@@ -0,0 +1,494 @@
...
+        let keys: &[Vec<u8>] = req.get_keys();
+        let res = self.inner.raw_batch_delete(keys);
+        let mut resp = RawBatchDeleteResponse::default();
+        if res.is_err() {
+            resp.set_error(format!(
+                "Non-existent keys:[{}]",
+                res.err()
+                    .unwrap()
+                    .iter()
+                    .map(|k| std::str::from_utf8(k).unwrap())
```
nrc, line 254: You can't convert the bytes into  string since the bytes might not be valid unicode. You could use the `debug` method or write a custom function to get a string representation.

Did you check TiKV's behaviour when trying batch delete non-existent keys? Does it give an error?

ekexium: I asked Yilin. Deleting non-existent keys should not return error.

mock-tikv/src/store.rs
```diff
@@ -0,0 +1,75 @@
...
+        );
+    }
+
+    // if success, return the key deleted
+    pub fn raw_delete(&self, key: &[u8]) {
+        let mut data = self.data.write().unwrap();
+        data.remove(key);
+    }
+
+    // if any of the key does not exist, return non-existent keys; delete other keys
```
nrc, line 68: This comment is out of date

nrc (approved):
Just one comment needs updating, otherwise lgtm

PR #161, opened 2020-09-11
Optimize condition in stores_for_range

ekexium:
In the condition, both upper bounds are exclusive. If `x=region_end` , we don’t need a next region.


nrc (approved):
lgtm

PR #162, opened 2020-09-13
[WIP] Add codec for encoding region for transaction client

georgeteo:
In this [github issue](https://github.com/tikv/tikv/issues/8626), it appears that transaction client does not encode/decode keys when querying PD.

This PR:
- duplicates bytes encoder/decoder from [tikv_utils](https://github.com/tikv/tikv/blob/e18d5a5f4ebab046de22ba9073040f773dd32460/components/tikv_util/src/codec/bytes.rs)
- Adds a method to encode Key
- Adds `PdCodecClient` that wraps/decorates `PdClient` encodes `key` and decodes `start_key` and `end_key` in region meta before delegating to underlying `PdClient`.
- Swap the underlying `PdClient` in `TransactionClient` with the one that encodes/decodes keys

Reference[ Go client.](https://github.com/tikv/client-go/blob/95774393107b33a79e97c7acd231ff26ed7deede/locate/codec.go#L32)

Caveat: I'm new to Rust so general Rust tips would be appreciated!

Signed-off-by: George Teo <george.c.teo@gmail.com>


src/pd/client.rs
```diff
@@ -266,6 +263,55 @@ impl<KvC: KvConnect + Send + Sync + 'static, Cl> PdRpcClient<KvC, Cl> {
     }
 }
 
+/// This client wraps a PdClient and encodes/decodes keys and regions according to
+/// TiKV transaction codec.
+pub struct PdCodecClient<PdClient = PdRpcClient> {
+    pd: Arc<PdClient>
+}
+
+impl<PdC: PdClient + Send + Sync + 'static> PdClient for PdCodecClient<PdC> {
```
georgeteo, line 272: TODO: write tests

ekexium:
You can check the commands and outputs of CI. Note you need to use latest nightly channel to build. 
Seems to me the only error was an unresolved import and some format issues.
```Rust
error[E0432]: unresolved import `crate::kv::codec::bytes`

   --> tikv-client-common/src/kv/codec.rs:134:9

    |

134 |     use crate::kv::codec::bytes;
```
Removing that line and `cargo fmt` should fix the build.

Also you need to [sign-off](https://github.com/tikv/tikv/blob/master/CONTRIBUTING.md#signing-off-the-commit) your commit.

ekexium:
PTAL @nrc @sticnarf

src/pd/client.rs
```diff
@@ -266,6 +267,58 @@ impl<KvC: KvConnect + Send + Sync + 'static, Cl> PdRpcClient<KvC, Cl> {
     }
 }
 
+/// This client wraps a PdClient and encodes/decodes keys and regions according to
+/// TiKV transaction codec.
+pub struct PdCodecClient<PdClient = PdRpcClient> {
```
nrc, line 272: I don't think we need a separate client for this, you can just add the encoding to the existing client.

tikv-client-common/src/kv/codec.rs
```diff
@@ -0,0 +1,211 @@
+use crate::{errors::Result, Error};
+use std::{io::Write, ptr};
+
+const ENC_GROUP_SIZE: usize = 8;
+const ENC_MARKER: u8 = b'\xff';
```
nrc, line 5:
```suggestion
const ENC_MARKER: u8 = 0xff';
```

tikv-client-common/src/kv/key.rs
```diff
@@ -101,6 +102,14 @@ impl Key {
             Bound::Excluded(self)
         }
     }
+
+    #[inline]
+    pub fn as_encoded(&self) -> Key {
```
nrc, line 107: I think we should use the type system to differentiate between an encoded and unencoded key by adding an `EncodedKey` type

tikv-client-common/src/kv/key.rs
```diff
@@ -101,6 +102,14 @@ impl Key {
...
+        let len = codec::max_encoded_bytes_size(self.0.len());
+        let mut encoded = Vec::with_capacity(len);
+        encoded.encode_bytes(self.into(), false).unwrap();
```
nrc, line 110:
```suggestion
        encoded.encode_bytes(self.0, false).unwrap();
```

nrc (commented):
Thanks for the PR! The rest of it looks great.

georgeteo:
@nrc: thanks for the quick review! Just to clarify the overarching point so I don't go down the wrong rabbit hole:

- Have a single `PdRpcClient`
- Handle `RawKey` and `EncodedKey` as two structs that impl a `Key` trait and wrap that concern in Key. 
- Use `EncodedKey` in `TransactionClient` and `RawKey` in `RawClient`?

Thoughts on naming `EncodedKey` as `TransactionKey` or `TransactionalKey` for `TransactionClient` to mirror `RawKey/RawClient`?

nrc:
> Handle RawKey and EncodedKey as two structs that impl a Key trait and wrap that concern in Key.

 I don't think you need a Key trait - the idea is to prevent using an encoded key where a raw key is required and vice versa, a Key trait would circumvent that.

> Use EncodedKey in TransactionClient and RawKey in RawClient?

AIUI, encoded keys are only used with PD and we use raw keys with TiKV.

> Thoughts on naming EncodedKey as TransactionKey or TransactionalKey for TransactionClient to mirror RawKey <-> RawClient?

I don't think it makes sense, whether a key is encoded is orthogonal to which API it is used with.

ekexium:
I made some changes:
1. Use a single PD client so that we don't need to write duplicated code. All encoding stuff is hiden in the client. Lower-level PD functions and higher-level clients do not care about encoding.
2. Fix a few bugs related to encoding
3. Find and fix a few bugs not related to encoding.
4. Construct a simple integration test to test with splitted regions.

src/pd/client.rs
```diff
@@ -92,12 +98,15 @@ pub trait PdClient: Send + Sync + 'static {
...
             };
             let end_key = end_key.clone();
-
             let this = self.clone();
             Either::Left(self.region_for_key(&start_key).and_then(move |region| {
                 let region_end = region.end_key();
                 this.map_region_to_store(region).map_ok(move |store| {
-                    if end_key.map(|x| x <= region_end).unwrap_or(false) || region_end.is_empty() {
+                    if end_key
+                        .map(|x| x <= region_end && !x.is_empty())
```
ekexium, line 105: Not sure whether the condition `!x.is_empty()` is necessary

ekexium (commented):
`None` and `Some(empty)` can both represent upper unbounded. I didn't comprehensively consider this change. Related to #174

ekexium:
@sticnarf PTAL

src/pd/client.rs
sticnarf: Yes I think it's necessary

tests/integration_tests.rs
```diff
@@ -283,8 +339,8 @@ async fn raw() -> Fallible<()> {
     let res = client
         .batch_get(vec!["k4".to_owned(), "k3".to_owned()])
         .await?;
-    assert_eq!(res[0].1, "v4".as_bytes());
-    assert_eq!(res[1].1, "v3".as_bytes());
+    assert_eq!(res[0].1, "v3".as_bytes());
```
sticnarf, line 342: Because the result is not the same order with the input, it's better to also check the key

sticnarf (approved):
LGTM

ekexium:
Hi @georgeteo, we need you to sign off your commits to merge the PR. About the DCO check: https://github.com/probot/dco#how-it-works

PR #164, opened 2020-09-14
Bound region range by query range

ekexium:
In `store_stream_for_range`, we passed the `region_range` to following steps and use it in request, which is incorrect if `region_range` is not a subset of origianl query range. We should take the common part of the two.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -8,7 +8,11 @@ use crate::{
 use futures::{future::BoxFuture, prelude::*, stream::BoxStream};
 use grpcio::CallOption;
 use kvproto::kvrpcpb;
-use std::{sync::Arc, time::Duration};
+use std::{
+    cmp::{max, min},
+    sync::Arc,
+    time::Duration,
+};
 use tikv_client_common::{BoundRange, Error, Key, Result};
 use tikv_client_store::{HasError, HasRegionError, KvClient, RpcFnType, Store};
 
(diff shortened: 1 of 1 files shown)


src/request.rs
```diff
@@ -188,16 +188,30 @@ pub fn store_stream_for_range<PdC: PdClient>(
...
-            (range, store)
+            let region_range = store.region.range();
+            (bound_range(region_range, range.clone()), store)
         })
         .into_stream()
         .boxed()
 }
 
+/// The range used for request should be the common range of `region_range` and `range`.
+fn bound_range(region_range: (Key, Key), range: BoundRange) -> (Key, Key) {
```
nrc, line 201: A better name might be `intersection`.

src/request.rs
```diff
@@ -188,16 +188,30 @@ pub fn store_stream_for_range<PdC: PdClient>(
...
+    let (mut lower, mut upper) = region_range;
+    let (lower_bound, upper_bound) = range.into_keys();
+    if lower_bound > lower {
+        lower = lower_bound;
+    }
+    if let Some(up) = upper_bound {
+        if up < upper || upper.is_empty() {
```
nrc, line 208: Why do we need the `is_empty` special case?

src/request.rs
```diff
@@ -188,16 +188,30 @@ pub fn store_stream_for_range<PdC: PdClient>(
...
+            upper = up;
+        }
+    }
+    (lower, upper)
```
nrc, line 212: I think you can use min, max, and unwrap_or, to avoid the mutation and conditionals

src/request.rs
ekexium: If `upper` is empty, which means it's upper unbounded, `up < upper` does not hold. We still need to bound it.

src/request.rs
```diff
@@ -188,16 +192,28 @@ pub fn store_stream_for_range<PdC: PdClient>(
...
         })
         .into_stream()
         .boxed()
 }
 
+/// The range used for request should be the intersection of `region_range` and `range`.
+fn bound_range(region_range: (Key, Key), range: BoundRange) -> (Key, Key) {
+    let (lower, upper) = region_range;
+    let (lower_bound, upper_bound) = range.into_keys();
+    let upper_bound = upper_bound.unwrap_or_default();
```
nrc, line 208: I think you can match on upper_bound, without doing this unwrap

nrc (commented):
looks much better!

nrc (approved):
lgtm

PR #165, opened 2020-09-15
propagate error in map_region_to_store

ekexium:
Propagate error in `map_region_to_store` instead of `unwrap`


nrc (approved):
lgtm, thanks!

PR #166, opened 2020-09-15
Handle None when converting BoundRange to KeyRange

ekexium:
Convert `None` to empty Vec instead of `unwrap`

### DIFF
--- a/tikv-client-common/src/kv/bound_range.rs
+++ b/tikv-client-common/src/kv/bound_range.rs
@@ -189,11 +189,7 @@ impl Into<kvrpcpb::KeyRange> for BoundRange {
         let (start, end) = self.into_keys();
         let mut range = kvrpcpb::KeyRange::default();
         range.set_start_key(start.into());
-        let end_key = match end {
-            Some(k) => k.into(),
-            None => vec![],
-        };
-        range.set_end_key(end_key);
+        range.set_end_key(end.unwrap_or_default().into());
         range
     }
 }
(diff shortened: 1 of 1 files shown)


tikv-client-common/src/kv/bound_range.rs
```diff
@@ -189,8 +189,11 @@ impl Into<kvrpcpb::KeyRange> for BoundRange {
         let (start, end) = self.into_keys();
         let mut range = kvrpcpb::KeyRange::default();
         range.set_start_key(start.into());
-        // FIXME handle end = None rather than unwrapping
-        end.map(|k| range.set_end_key(k.into())).unwrap();
+        let end_key = match end {
+            Some(k) => k.into(),
+            None => vec![],
+        };
```
nrc, line 195: You could use map and unwrap_or_default, rather than a match statement.

Where is the bound range used? Did you check that an empty end key has these semantics in TiKV?

ekexium: It's used in RawBatchScanRequest.
https://github.com/tikv/client-rust/blob/a3a0d9fe6503aed6469bc9b44130bb1db318c838/src/raw/requests.rs#L423

In TiKV, the semantics is upper unbounded.
https://github.com/tikv/tikv/blob/fbf0936bed8a2424347140115c99f070cf34ab9e/src/storage/mod.rs#L1335-L1343

nrc (approved):
lgtm

PR #168, opened 2020-09-17
Fix txn/scan: combines local buffer and fetched result

ekexium:
Previous implementatoin of txn/scan does not make use of local buffer, which causes erroneous results.

### DIFF
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -1,8 +1,12 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
 use kvproto::kvrpcpb;
-use std::{collections::BTreeMap, future::Future, sync::Mutex};
-use tikv_client_common::{Key, KvPair, Result, Value};
+use std::{
+    collections::{BTreeMap, HashMap},
+    future::Future,
+    sync::Mutex,
+};
+use tikv_client_common::{BoundRange, Key, KvPair, Result, Value};
 
 /// A caching layer which buffers reads and writes in a transaction.
 #[derive(Default)]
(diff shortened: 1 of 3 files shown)


ekexium:
Do you think we should add results of scan to local buffer? @nrc

src/transaction/buffer.rs
```diff
@@ -70,6 +74,56 @@ impl Buffer {
         Ok(results)
     }
 
+    pub async fn scan_and_fetch<F, Fut>(
```
nrc, line 77: Could you document this function please

src/transaction/buffer.rs
```diff
@@ -70,6 +74,56 @@ impl Buffer {
...
+                .filter(|(_, m)| matches!(m, Mutation::Del))
+                .count() as u32;
+
+        let mut results = f(range, redundant_limit)
+            .await?
+            .into_iter()
+            .map(|pair| pair.into())
+            .collect::<HashMap<Key, Value>>();
+
+        // override using local data
```
nrc, line 105: For the results which are not overridden, we should insert them into the buffer as `Mutation::Cached`

nrc (commented):
Looks good, just missing caching the results

src/transaction/buffer.rs
```diff
@@ -70,6 +74,62 @@ impl Buffer {
...
+                }
+                Mutation::Del => {
+                    results.remove(k);
+                }
+                _ => {}
+            }
+        }
+
+        // update local buffer
+        for (k, v) in results.iter() {
```
nrc, line 120:
```suggestion
        for (k, v) in &results {
```

Is more idiomatic

nrc (approved):
lgtm

PR #172, opened 2020-09-22
Make raw and transactional clients have similar API

ekexium:
Previous:
- raw client has a `with_key_only` parameter and use `put`
- txn client has `key_only` parameters in its `scan` method and use `set`

Changes:
- Rename `txn.set` to `txn.put`
- Move `key_only` parameter from `txn.scan` to txn client


nrc (approved):
lgtm

PR #173, opened 2020-09-23
Fix txn.batch_get

ekexium:
Previously `txn.batch_get` returns an iterator of `(Key, Option<Value>)`. However, it does not always return all non-existent entries, which is confusing. 

Changes:
Change the signature and behavior to be same as `raw.batch_get`: it returns an iterator of `Kvpair` and skips non-existent entries.

### DIFF
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -103,7 +103,7 @@ impl Client {
     /// Create a new 'batch get' request.
     ///
     /// Once resolved this request will result in the fetching of the values associated with the
-    /// given keys.
+    /// given keys, skipping non-existent entris.
     ///
     /// ```rust,no_run
     /// # use tikv_client::{KvPair, Config, RawClient};
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -103,7 +103,7 @@ impl Client {
     /// Create a new 'batch get' request.
     ///
     /// Once resolved this request will result in the fetching of the values associated with the
-    /// given keys, skipping non-existent entris.
+    /// given keys, skipping non-existent entries.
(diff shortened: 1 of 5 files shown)


src/raw/client.rs
```diff
@@ -103,7 +103,7 @@ impl Client {
     /// Create a new 'batch get' request.
     ///
     /// Once resolved this request will result in the fetching of the values associated with the
-    /// given keys.
+    /// given keys, skipping non-existent entris.
```
nrc, line 106:
```suggestion
    /// given keys, skipping non-existent entries.
```

src/transaction/buffer.rs
```diff
@@ -58,7 +61,12 @@ impl Buffer {
                 })
                 .partition(|(_, v)| *v == MutationValue::Undetermined);
 
-            let cached_results = cached_results.into_iter().map(|(k, v)| (k, v.unwrap()));
+            let cached_results = cached_results
+                .into_iter()
+                .filter_map(|(k, v)| match v.unwrap() {
```
nrc, line 66: could be v.map(...) rather than matching

src/transaction/transaction.rs
```diff
@@ -75,7 +75,7 @@ impl Transaction {
             .await
     }
 
-    /// Gets the values associated with the given keys.
+    /// Gets the values associated with the given keys, skipping non-existent entris.
```
nrc, line 78:
```suggestion
    /// Gets the values associated with the given keys, skipping non-existent entries.
```

nrc (approved):
lgtm, some minor comments

PR #174, opened 2020-09-24
Fill a missing situation in bound_range

ekexium:
`upper_bound` can be `Some(empty)`.


sticnarf (approved):
I approve this as a temporary solution. But it is more like a design mistake if `Some(empty)` and `None` both represent unbounded

ekexium:
Close this PR as the change is already merged to master branch in #162 .

PR #175, opened 2020-09-24
Use delete_range to clear TiKV in integration tests

ekexium:
We can use `delete_range` on {Default, Lock, Write} CFs to clear TiKV.

Needs #176 to work correcly.


nrc (approved):
lgtm

PR #176, opened 2020-09-25
Fix some misuse of into_future()

ekexium:
In previous implementation, prewrite/commit/batch_rollback does not poll the entire result stream.

Changes:
Fold the stream instead of `into_future`


ekexium:
@sticnarf PTAL

sticnarf (approved):
LGTM

Issue #171, opened 2020-09-21
Plan for releasing 0.1 version


ekexium:
We're going to make the project ready for a 0.1 release in the following weeks. Please leave a comment if you have suggestions!

We expect it to be fully functional in most cases, and enable Rust users to freely play with TiKV. Becaues of the publish blocker #32 , we will probably not publish it on crate.io. 

Supported functionalities in 0.1:
- Raw API
  - `put`, `get`, `delete`, `scan`, `batch_put`, `batch_get`, `batch_delete`, `batch_scan`, `delete_range`
- Transactional API (optimistic) 
  - `begin`, `commit`, `rollback`, `put`, `get`, `delete`, `scan`, `batch_get`, `lock`

Checklist:
- Functionality
  - [x] ~Whether to remove `delete_range`~ `delete_range` is currently supported by TiKV.
  - [x] ~GC will not be provided in 0.1~ #180
- Bugfix and test
  - [x] Encoding problem #162 
  - [x] Parameters of `group_keys_by_region` must be sorted #162 
  - [x] Some requests only sends to the first region because of misuse of `stream.into_future` #176 
  - [x] For end_key, `None` and `Some(empty)` should both mean unbounded, while in some palces we miss the second situation. We cannot only use `None` to represent unbounded because parameters is first converted into kvproto request parameters (which is Vec<u8>). Related PR: #162, #174, #179 . There are probably more.
  - [x] `limit` only applies to each region for raw_scan and raw_batch_scan #179 
  - [x] Integration tests
    - [x] raw
    - [x] transactional
- Usability
  - [x] Consistent naming: `put`/`set`, `key_only`/`with_key_only`, etc.
  - [x] `batch_get` returns `Iter<Option<KvPair>>`, while may not contain all non-existent entries. #173 
- Robustness
  - [x] Do not panic when error happens
  - [ ] ~~[Jepsen Test](https://github.com/tikv/tikv/issues/9588)~~
- Documentation
  - [x] Tutorial for user #170 
- Examples and use cases

nrc:
I'm thinking of the 0.1 release again. I think we should aim for the end of March. I've added some issues which I think should be blocking to the 0.1 milestone.

I don't think we need to block on the Jepsen test, though we should probably block a 1.0 on that.

@ekexium are you aware of anywhere where we panic where we should create an error?

ekexium:
> I don't think we need to block on the Jepsen test, though we should probably block a 1.0 on that.

Agree.

[...]

PR #170, opened 2020-09-21
Update readme

ekexium:
Update dependency, limit and add simple code examples.

PR #179, opened 2020-09-29
Fix limit problem in raw_scan and unbouned problem in batch_scan

ekexium:
Changes:
-  `raw_scan` limit the final results.
- `end_key=Some(empty)` is treated as upper unbouned. The problem is also found in other places #171 . 

Remaining problem:
- The parameter `each_limit` of `raw_batch_scan` is not what we expect. It works on each region of each range, but not each range. It's not convenient to fix this problem so I suggest we keep this request but mark it as experimental for users #170 .


nrc (approved):
lgtm

PR #182, opened 2020-10-12
GC: initial implementation

ekexium:
An attempt to implement GC.

- The implementation does not contain the second step "delete ranges" described in the [doc](https://docs.pingcap.com/tidb/stable/garbage-collection-overview).

### DIFF
--- a/src/transaction/client.rs
+++ b/src/transaction/client.rs
@@ -8,7 +8,7 @@ use crate::{
 };
 use futures::executor::ThreadPool;
 use kvproto::kvrpcpb;
-use std::sync::Arc;
+use std::{mem, sync::Arc};
 use tikv_client_common::{Config, Result, Timestamp, TimestampExt};
 
 const SCAN_LOCK_BATCH_SIZE: u32 = 1024; // TODO: cargo-culted value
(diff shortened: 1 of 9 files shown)


ekexium:
@sticnarf Could you take a look at if it is correct?

src/transaction/client.rs
```diff
@@ -88,6 +91,37 @@ impl Client {
         self.pd.clone().get_timestamp().await
     }
 
+    pub async fn gc(&self, safepoint: Timestamp) -> Result<bool> {
```
sticnarf, line 94: Could you please add some necessary documentation to this method?

src/transaction/client.rs
```diff
@@ -88,6 +91,37 @@ impl Client {
...
+        // scan all locks with ts <= safepoint
+        let mut locks: Vec<kvrpcpb::LockInfo> = vec![];
+        let mut start_key = vec![];
+        loop {
+            let req =
+                new_scan_lock_request(start_key.clone(), safepoint.clone(), SCAN_LOCK_BATCH_SIZE);
```
sticnarf, line 100: I think we can use `mem::take(&mut start_key)` to reduce a clone.

sticnarf (approved):
LGTM

src/transaction/client.rs
```diff
@@ -88,6 +91,47 @@ impl Client {
         self.pd.clone().get_timestamp().await
     }
 
+    /// Cleans stale MVCC records in TiKV.
+    ///
+    /// It is done by:
+    /// 1. resolve all locks with ts <= safepoint
+    /// 2. update safepoint to PD
+    ///
+    /// This is a simplified version of [GC in TiDB](https://docs.pingcap.com/tidb/stable/garbage-collection-overview).
```
nrc, line 100: Could you state how it is simplified please?

nrc (approved):
Mostly lgtm, I wouldn't mind a bit more in one comment

ti-srebot (approved):

PR #184, opened 2020-10-16
Basic support of pessimistic transaction

### LINKED ISSUE #155: Feature Request: Pessimistic transaction
I found nothing about pessimistic transaction in this repo, maybe we should support this.

longfangsong:
Currently client-rust doesn't support pessimistic transaction, but pessimistic transaction is not only useful but also more and more important in TiKV. The client may support it, so it can at least work as a simpler demo of using the pessimistic transaction system in TiKV than the whole TiDB.
So this pr adds basic support of pessimistic transaction into the client.
Close #155.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -13,10 +13,11 @@ use std::{
     sync::Arc,
     time::Duration,
 };
-use tikv_client_common::{BoundRange, Error, Key, Result};
+use tikv_client_common::{BoundRange, Error, ErrorKind, Key, Result};
 use tikv_client_store::{HasError, HasRegionError, KvClient, RpcFnType, Store};
 
 const LOCK_RETRY_DELAY_MS: u64 = 10;
+const LOCK_RETRY_MAX_TIMES: usize = 10;
 const DEFAULT_REGION_BACKOFF: NoJitterBackoff = NoJitterBackoff::new(2, 500, 10);
 
 pub trait KvRequest: Sync + Send + 'static + Sized {
@@ -32,7 +33,7 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
 
     fn execute(self, pd_client: Arc<impl PdClient>) -> BoxFuture<'static, Result<Self::Result>> {
         Self::reduce(
-            self.response_stream(pd_client)
+            self.response_stream(pd_client, LOCK_RETRY_MAX_TIMES)
                 .and_then(|mut response| match response.error() {
                     Some(e) => future::err(e),
                     None => future::ok(response),
@@ -45,14 +46,16 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
     fn response_stream(
         self,
         pd_client: Arc<impl PdClient>,
+        remaining_retry_count: usize
     ) -> BoxStream<'static, Result<Self::RpcResponse>> {
-        self.retry_response_stream(pd_client, DEFAULT_REGION_BACKOFF)
+        self.retry_response_stream(pd_client, DEFAULT_REGION_BACKOFF, remaining_retry_count)
     }
 
     fn retry_response_stream(
         mut self,
         pd_client: Arc<impl PdClient>,
         backoff: impl Backoff,
+        remaining_retry_count: usize
     ) -> BoxStream<'static, Result<Self::RpcResponse>> {
         let stores = self.store_stream(pd_client.clone());
         stores
@@ -82,6 +85,7 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
                 // Resolve locks
                 let locks = response.take_locks();
                 if !locks.is_empty() {
+                    // todo: exit directly when found PessimisticLock
                     let pd_client = pd_client.clone();
                     return resolve_locks(locks, pd_client.clone())
                         .map_ok(|resolved| {
@@ -89,8 +93,18 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
                             let delay_ms = if resolved { 0 } else { LOCK_RETRY_DELAY_MS };
                             futures_timer::Delay::new(Duration::from_millis(delay_ms))
                         })
-                        .map_ok(move |_| request.response_stream(pd_client))
-                        .try_flatten_stream()
+                        .map_ok(move |_| {
+                            if remaining_retry_count != 0 {
+                                request.response_stream(pd_client, remaining_retry_count - 1)
+                            } else {
+                                // todo: when met up with a long-time lock, this might be called
+                                // recusively for many times and cause a stack overflow here
+                                // currently I use a temporaray solution of limiting retry times,
+                                // a better way to handle this situation needs to be discussed
+                                stream::once(future::err(ErrorKind::Unimplemented.into())).boxed()
+                            }
+                        }
+                        ).try_flatten_stream()
                         .boxed();
                 }
                 stream::once(future::ok(response)).boxed()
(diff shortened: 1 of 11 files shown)


longfangsong:
PTAL @nrc @ekexium @sticnarf

src/transaction/transaction.rs
```diff
@@ -232,6 +234,69 @@ impl Transaction {
         .commit()
         .await
     }
+
+    /// Pessimisticly lock the keys
```
ekexium, line 238:
```suggestion
    /// Pessimistically lock the keys
```

ekexium (commented):
Users may acquire pessimistic locks and then use optimistic commit protocol. I think it would be better if we prohibit this kind of behavior.

src/request.rs
```diff
@@ -82,14 +85,25 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
...
                     return resolve_locks(locks, pd_client.clone())
                         .map_ok(|resolved| {
                             // TODO: backoff
                             let delay_ms = if resolved { 0 } else { LOCK_RETRY_DELAY_MS };
                             futures_timer::Delay::new(Duration::from_millis(delay_ms))
                         })
-                        .map_ok(move |_| request.response_stream(pd_client))
+                        .map_ok(move |_| {
+                            if remaining_retry_count != 0 {
+                                // todo: when met up with a long-time lock, this might be called
```
nrc, line 98: Can you file an issue for this please?

src/transaction/requests.rs
```diff
@@ -475,6 +491,84 @@ pub fn new_batch_rollback_request(
...
+
+        req
+    }
+
+    fn map_result(_result: Self::RpcResponse) -> Self::Result {}
+
+    fn reduce(
+        results: BoxStream<'static, Result<Self::Result>>,
+    ) -> BoxFuture<'static, Result<Self::Result>> {
+        results.try_for_each(|_| future::ready(Ok(()))).boxed()
```
nrc, line 535:
```suggestion
        results.try_for_each_concurrent(|_| future::ready(Ok(()))).boxed()
```

src/transaction/transaction.rs
```diff
@@ -232,6 +234,69 @@ impl Transaction {
...
+    /// # use tikv_client_common::Key;
+    /// # futures::executor::block_on(async {
+    /// # let client = Client::new(Config::default()).await.unwrap();
+    /// let mut txn = client.begin().await.unwrap();
+    /// // ... Do some actions.
+    /// let key1: Key = b"key1".to_vec().into();
+    /// let result: () = txn.pessimistic_lock(vec![key1.clone()]).await.unwrap();
+    /// # });
+    /// ```
+    pub async fn pessimistic_lock(
```
nrc, line 252: I think this should not be part of the public API, instead we should set the transaction to be optimistic or pessimistic, and if pessimistic then we take the lock on each update/insert/delete, and we can add an option for reads to make locking reads.

src/transaction/transaction.rs
```diff
@@ -232,6 +234,69 @@ impl Transaction {
...
+    /// # use futures::prelude::*;
+    /// # futures::executor::block_on(async {
+    /// # let client = Client::new(Config::default()).await.unwrap();
+    /// let mut txn = client.begin().await.unwrap();
+    /// // ... Do some actions.
+    /// let req = txn.pessimistic_commit();
+    /// let result: () = req.await.unwrap();
+    /// # });
+    /// ```
+    pub async fn pessimistic_commit(&mut self) -> Result<()> {
```
nrc, line 290: Similarly, we shouldn't have separate pessimistic commit API, commit should do the right thing based on whether then transaction is optimistic or pessimistic

src/transaction/transaction.rs
```diff
@@ -278,6 +343,49 @@ impl TwoPhaseCommitter {
         }
     }
 
+    async fn pessimistic_commit(mut self, for_update_ts: u64) -> Result<()> {
```
nrc, line 346: If possible, it would be nice for these two methods to share code with the optimistic versions.

src/request.rs
longfangsong: Now in #185

src/request.rs
```diff
@@ -32,7 +33,7 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
 
     fn execute(self, pd_client: Arc<impl PdClient>) -> BoxFuture<'static, Result<Self::Result>> {
         Self::reduce(
-            self.response_stream(pd_client)
+            self.response_stream(pd_client, LOCK_RETRY_MAX_TIMES)
```
sticnarf, line 36: I'm thinking if we can combine `LOCK_RETRY_MAX_TIMES` into backoff because backoff has the function of retry limit. Maybe give PessimisticLock request a custom backoff?

src/request.rs
```diff
@@ -46,13 +45,14 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
         self,
         pd_client: Arc<impl PdClient>,
     ) -> BoxStream<'static, Result<Self::RpcResponse>> {
-        self.retry_response_stream(pd_client, DEFAULT_REGION_BACKOFF)
+        self.retry_response_stream(pd_client, DEFAULT_REGION_BACKOFF, DEFAULT_LOCK_BACKOFF)
```
sticnarf, line 48: Pessimistic lock and optimistic lock has different backoffs. When doing pessimistic locking, no extra waiting is needed at the client because it also waits at TiKV. But we need to wait at the client for optimistic locks.

src/transaction/transaction.rs
```diff
@@ -75,6 +80,34 @@ impl Transaction {
...
+    /// # let client = Client::new(Config::new(vec!["192.168.0.100", "192.168.0.101"])).await.unwrap();
+    /// let mut txn = client.begin_pessimistic().await.unwrap();
+    /// let key = "TiKV".to_owned();
+    /// let result: Option<Value> = txn.get_and_lock(key).await.unwrap();
+    /// // now the key "TiKV" is locked, other transactions cannot modify it
+    /// // Finish the transaction...
+    /// txn.commit().await.unwrap();
+    /// # });
+    /// ```
+    pub async fn get_and_lock(&mut self, key: impl Into<Key>) -> Result<Option<Value>> {
```
sticnarf, line 98: Is it more correct to call it "lock_and_get"?

longfangsong: Another name I've considered is `get_for_update`.

sticnarf: Sounds good

nrc (approved):
lgtm

nrc:
@ekexium @sticnarf PTAL

src/transaction/transaction.rs
```diff
@@ -70,7 +75,37 @@ impl Transaction {
...
+    /// let key = "TiKV".to_owned();
+    /// let result: Option<Value> = txn.get_for_update(key).await.unwrap();
+    /// // now the key "TiKV" is locked, other transactions cannot modify it
+    /// // Finish the transaction...
+    /// txn.commit().await.unwrap();
+    /// # });
+    /// ```
+    pub async fn get_for_update(&mut self, key: impl Into<Key>) -> Result<Option<Value>> {
+        if !self.is_pessimistic {
+            panic!("get_and_lock is not allowed in optimistic transaction!")
```
ekexium, line 101: We can return an error instead of panicking.

src/transaction/transaction.rs
```diff
@@ -105,11 +140,46 @@ impl Transaction {
...
+    /// // Finish the transaction...
+    /// txn.commit().await.unwrap();
+    /// # });
+    /// ```
+    pub async fn batch_get_for_update(
+        &mut self,
+        keys: impl IntoIterator<Item = impl Into<Key>>,
+    ) -> Result<impl Iterator<Item = KvPair>> {
+        if !self.is_pessimistic {
+            panic!("batch_get_and_lock is not allowed in optimistic transaction!")
```
ekexium, line 176: same here

ekexium (approved):
LGTM. Nit: I think users won't expect panicking in requests.

sticnarf (approved):

ti-srebot (approved):

PR #187, opened 2020-10-26
Propagates error instead of panicking when getting context returns error

ekexium:
Try not to panic, return error instead.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -64,26 +64,18 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
         stores
             .and_then(move |(key_data, store)| {
                 let request = self.make_rpc_request(key_data, &store);
-                let mut req = None;
-                let mut err = None;
-                match request {
-                    Ok(r) => req = Some(r),
-                    Err(e) => err = Some(e),
-                }
+                let (req, err) = match request {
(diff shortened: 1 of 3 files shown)


ekexium:
/cc @nrc @sticnarf 
Need to improve style in request.rs line 67-86. Do you have suggestions?

src/request.rs
```diff
@@ -64,20 +64,31 @@ pub trait KvRequest: Sync + Send + 'static + Sized {
         stores
             .and_then(move |(key_data, store)| {
                 let request = self.make_rpc_request(key_data, &store);
+                let mut req = None;
+                let mut err = None;
+                match request {
+                    Ok(r) => req = Some(r),
+                    Err(e) => err = Some(e),
+                }
```
nrc, line 72: So, I played around with this a lot, and I don't think it can be improved much - the borrowed/owned issue is (I think) impossible to work around. You can make the code a little bit more idiomatic by using `let (req, err) = match ...` here, rather than mutable variables, and matching on `req` on line 75, rather than using if/else,

ekexium: All right. Thanks.

nrc (approved):
lgtm

PR #190, opened 2020-10-28
Refactor and tidy up APIs in upstream crates

nrc:
The commit descriptions describe the changes.

PTAL @ekexium


tikv-client-store/src/lib.rs
```diff
@@ -1,32 +1,35 @@
...
 use std::{sync::Arc, time::Duration};
+use tikv_client_common::stats::tikv_stats;
 
 /// A trait for connecting to TiKV stores.
 pub trait KvConnect: Sized + Send + Sync + 'static {
     type KvClient: KvClient + Clone + Send + Sync + 'static;
 
     fn connect(&self, address: &str) -> Result<Self::KvClient>;
+
+    fn connect_to_store(&self, region: Region, address: String) -> Result<Store<Self::KvClient>> {
```
ekexium, line 28: What is this function used for?

ekexium (commented):
Others LGTM

tikv-client-store/src/lib.rs
```diff
@@ -1,30 +1,35 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
+#![feature(type_alias_impl_trait)]
```
ekexium, line 2: Will this change stop users using this crate in stable channel?

tikv-client-store/src/lib.rs
nrc: Connecting to a store :-) It's a convenience method for users who use this crate directly so that they don't need to deal with the KvClient trait.

tikv-client-store/src/lib.rs
nrc: Yes, I've undone it in my next PR.

PR #191, opened 2020-10-29
More polishing of the API and refactoring

nrc:
PTAL @ekexium The commit descriptions describe the refactorings.

### DIFF
--- a/src/raw/requests.rs
+++ b/src/raw/requests.rs
@@ -11,30 +11,24 @@ use crate::{
     BoundRange, ColumnFamily, Key, KvPair, Result, Value,
 };
 use futures::{future::BoxFuture, prelude::*, stream::BoxStream};
-use kvproto::{kvrpcpb, tikvpb::TikvClient};
+use kvproto::kvrpcpb;
 use std::{mem, sync::Arc};
-use tikv_client_store::{KvClient, RpcFnType, Store};
+use tikv_client_store::Store;
 
 impl KvRequest for kvrpcpb::RawGetRequest {
     type Result = Option<Value>;
     type RpcResponse = kvrpcpb::RawGetResponse;
     type KeyData = Key;
-    const REQUEST_NAME: &'static str = "raw_get";
-    const RPC_FN: RpcFnType<Self, Self::RpcResponse> = TikvClient::raw_get_async_opt;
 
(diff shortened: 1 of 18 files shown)


src/request.rs
```diff
@@ -62,7 +63,8 @@ pub trait KvRequest: Request + Clone + Sync + Send + 'static + Sized {
                 let request = self.make_rpc_request(key_data, &store);
                 async move {
                     let request = request?;
-                    let response = store.dispatch::<_, Self::RpcResponse>(&request).await?;
+                    let response = store.dispatch::<_, Self::RpcResponse>(&request).await;
+                    let response = tikv_stats(request.label()).done(response)?;
```
ekexium, line 70: Should we create the stats before dispatch? Otherwise, the duration stats don't make much sense.

src/raw/requests.rs
```diff
@@ -268,10 +265,8 @@ impl KvRequest for kvrpcpb::RawBatchDeleteRequest {
 
     fn map_result(_: Self::RpcResponse) -> Self::Result {}
 
-    fn reduce(
-        results: BoxStream<'static, Result<Self::Result>>,
-    ) -> BoxFuture<'static, Result<Self::Result>> {
-        results.try_collect().boxed()
+    async fn reduce(results: BoxStream<'static, Result<Self::Result>>) -> Result<Self::Result> {
+        results.try_collect().await
```
ekexium, line 269: Is it better to use `try_for_each_concurrent` instead of `try_collect`?

ekexium (commented):
LGTM. Amazing! It looks much better and cleaner. I just have two questions not quite related to this PR.

src/raw/requests.rs
nrc: `try_for_each_concurrent` only returns `()`, so I don't think we can use it instead of `try_collect`.

src/request.rs
nrc: Yes it should :-)

src/raw/requests.rs
ekexium: But the result types are all `()` where we use `try_collect`

nrc: Oh, in which case we probably can!

PR #192, opened 2020-10-30
Build our own protos

### LINKED ISSUE #32: Resolving blockers for publishing
This crate is intended to be published to the Crates.io registry to support our users easily (and correctly) depending on this client.

It will also enable future `cargo install tikv-client` uses for our command line tool (#17).

The current blockers:

* [x] CI is red (fixed by #22 and #30)
* [x] No README (fixed by #31)
* [ ] KVproto is a git based dependency (this is a **hard** blocker for publishing. https://github.com/pingcap/kvproto/issues/280)

These issues prevent us from reaching a published 0.0.1 release.

nrc:
I've copied protos from kvproto so we no longer need that dependency, then we build them ourselves, so we only need the Prost versions. That means we don't need any Git dependencies or need to replace any deps (and can publish the client crates).

Closes #32

PTAL @ekexium


ekexium:
LGTM. 
According to @sticnarf, TiKV is always backward compatible. Thus, we don't have to manually copy any future changes applied to kvproto. But we should clearly state the lowest required version of TiKV server.

nrc:
> According to @sticnarf, TiKV is always backward compatible. Thus, we don't have to manually copy any future changes applied to kvproto.

True, but we will need to copy them if we wish to support new features

> But we should clearly state the lowest required version of TiKV server.

Agree

PR #193, opened 2020-10-31
Refactor: move some modules to main crate

ekexium:
This PR is based on #192 which should be merged first.

Changes are described by commit messages.


src/lib.rs
```diff
@@ -92,15 +96,16 @@ mod proptests;
 #[macro_use]
 extern crate log;
 
+#[doc(inline)]
+pub use crate::kv::{codec, BoundRange, Key, KvPair, ToOwnedRange, Value};
```
nrc, line 100: Does codec need to be public?

src/store.rs
```diff
@@ -0,0 +1,29 @@
...
+        Ok(self
+            .client
+            .dispatch(request)
+            .await?
+            .downcast()
+            .expect("Downcast failed"))
+    }
+}
+
+pub trait KvConnectStore: KvConnect {
```
nrc, line 23: There should be an impl of this trait for `TikvConnect`

nrc (commented):
one question and one missing impl, otherwise lgtm

src/lib.rs
ekexium: For users using only this library, no. Is it possible that someone wants to manually encode keys to directly interact with PD?

nrc: I think that might be possible, but then it should be exported by the PD crate and not the main crate.

ekexium: Makes sense.

PR #196, opened 2020-11-06
Fix: rollback and commit in transaction

### LINKED ISSUE #35: Transaction Drop Behavior
As part of #15, one question to consider is how to handle it when a transaction is dropped?

Transaction *should* have `#[must_use]` on it to create a warning for users who might accidentally drop it, but in the case of something like `let _ = client.begin()` we should consider if it is appropriate to just drop the memory. Should we attempt to rollback?

ekexium:
We should not allow further operations in a transaction after it successfully commits or rolls back.
Close #35 .

I extended Fangsong's implementation at the drop of a `Transaction`: spawn a rollback work at the thread pool. 

Changes:

- Add a rollback function to Transaction
- Record the status of transaction, and check before every operation. 
- Use clone instead of `mem::take()` because commit or rollback can fail and the mutations are expected to still exist in the transaction.
- Automatically rollback transaction when `Transaction` gets dropped instead of when `TwoPhaseCommitter` gets dropped.

### DIFF
--- a/src/transaction/transaction.rs
+++ b/src/transaction/transaction.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use derive_new::new;
 use futures::{executor::ThreadPool, prelude::*, stream::BoxStream};
-use std::{iter, mem, ops::RangeBounds, sync::Arc};
+use std::{iter, ops::RangeBounds, sync::Arc};
 use tikv_client_proto::{kvrpcpb, pdpb::Timestamp};
 
 #[derive(PartialEq)]
@@ -519,27 +519,7 @@ impl TwoPhaseCommitter {
 impl Drop for Transaction {
     fn drop(&mut self) {
         if self.status == TransactionStatus::Active {
-            let buffer = mem::take(&mut self.buffer);
-            let bg_worker = self.bg_worker.clone();
-            let rpc = self.rpc.clone();
-            let version = self.timestamp.version();
-            let for_update_ts = self.for_update_ts;
-            self.bg_worker.spawn_ok(async move {
-                TwoPhaseCommitter::new(
-                    buffer.to_proto_mutations().await,
-                    version,
-                    bg_worker.clone(),
-                    rpc.clone(),
-                    for_update_ts,
-                )
-                .rollback()
-                .map(|res| {
-                    if let Err(e) = res {
-                        warn!("Failed to rollback at Drop: {}", e);
-                    }
-                })
-                .await
-            })
+            panic!("Dropping an active transaction. Consider commit or rollback it.")
         }
     }
 }
--- a/src/transaction/transaction.rs
+++ b/src/transaction/transaction.rs
@@ -14,8 +14,13 @@ use tikv_client_proto::{kvrpcpb, pdpb::Timestamp};
 
 #[derive(PartialEq)]
 enum TransactionStatus {
+    /// The transaction is read-only [`Snapshot`](super::Snapshot::Snapshot), no need to commit or rollback or panic on drop.
+    ReadOnly,
+    /// The transaction have not been committed or rolled back.
     Active,
+    /// The transaction has committed.
     Committed,
+    /// The transaction has rolled back.
     Rolledback,
 }
 
@@ -54,9 +59,15 @@ impl Transaction {
         rpc: Arc<PdRpcClient>,
         key_only: bool,
         is_pessimistic: bool,
+        read_only: bool,
     ) -> Transaction {
+        let status = if read_only {
+            TransactionStatus::ReadOnly
+        } else {
+            TransactionStatus::Active
+        };
         Transaction {
-            status: TransactionStatus::Active,
+            status,
             timestamp,
             buffer: Default::default(),
             bg_worker,
@@ -388,7 +399,7 @@ impl Transaction {
 
     fn check_status(&self) -> Result<()> {
         match self.status {
-            TransactionStatus::Active => Ok(()),
+            TransactionStatus::ReadOnly | TransactionStatus::Active => Ok(()),
             TransactionStatus::Committed => Err(ErrorKind::OperationAfterCommitError.into()),
             TransactionStatus::Rolledback => Err(ErrorKind::OperationAfterCommitError.into()),
         }
--- a/src/transaction/transaction.rs
+++ b/src/transaction/transaction.rs
@@ -9,8 +9,15 @@ use crate::{
 };
 use derive_new::new;
 use futures::{executor::ThreadPool, prelude::*, stream::BoxStream};
-use std::{iter, mem, ops::RangeBounds, sync::Arc};
+use std::{iter, ops::RangeBounds, sync::Arc};
 use tikv_client_proto::{kvrpcpb, pdpb::Timestamp};
+
+#[derive(PartialEq)]
+enum TransactionStatus {
+    Normal,
+    Committed,
+    Rolledback,
+}
 
 /// A undo-able set of actions on the dataset.
 ///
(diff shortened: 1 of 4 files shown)


ekexium:
\cc @sticnarf, @nrc, @longfangsong

src/transaction/transaction.rs
```diff
@@ -12,6 +12,13 @@ use futures::{executor::ThreadPool, prelude::*, stream::BoxStream};
 use std::{iter, mem, ops::RangeBounds, sync::Arc};
 use tikv_client_proto::{kvrpcpb, pdpb::Timestamp};
 
+#[derive(PartialEq)]
+enum TransactionStatus {
+    Normal,
```
nrc, line 17: Rather than normal, how about `Open` or `Pending` or `Active`?

src/transaction/transaction.rs
```diff
@@ -73,6 +82,7 @@ impl Transaction {
     /// # });
     /// ```
     pub async fn get(&self, key: impl Into<Key>) -> Result<Option<Value>> {
+        self.check_status()?;
```
nrc, line 85: We could check this only on commit/rollback, i.e., the semantics is that we only allow one of those actions to happen once. Then we can combine the check and the set, and it makes things a bit less noisy. On the other hand we don't catch bugs where the user commits once but calls `get` after the commit, etc.

src/transaction/transaction.rs
```diff
@@ -450,8 +498,8 @@ impl TwoPhaseCommitter {
             .await
     }
 
-    fn rollback(&mut self) -> impl Future<Output = Result<()>> + 'static {
-        let mutations = mem::take(&mut self.mutations);
+    fn rollback(&self) -> impl Future<Output = Result<()>> + 'static {
```
nrc, line 501: Could we take `self` by value and avoid the clone?

src/transaction/transaction.rs
```diff
@@ -466,14 +514,30 @@ impl TwoPhaseCommitter {
...
-impl Drop for TwoPhaseCommitter {
+impl Drop for Transaction {
     fn drop(&mut self) {
-        if !self.committed {
-            self.bg_worker.clone().spawn_ok(self.rollback().map(|res| {
-                if let Err(e) = res {
-                    warn!("Failed to rollback: {}", e);
-                }
-            }))
+        if self.status == TransactionStatus::Normal {
```
nrc, line 519: I think we should panic here, rather than rolling back - I think it is a bit bad to do serious work in a dtor.

src/transaction/transaction.rs
```diff
@@ -305,15 +322,42 @@ impl Transaction {
...
+            self.status = TransactionStatus::Committed;
+        }
+        res
+    }
+
+    /// Rollback the transaction.
+    ///
+    /// If it succeeds, all mutations made by this transaciton will not take effect.
+    pub async fn rollback(&mut self) -> Result<()> {
+        self.check_status()?;
```
nrc, line 346: Can we check to see if we need to rollback? If we haven't taken any locks or done select for update, then I think we can just drop the transaction

nrc (commented):
Looks good!

src/transaction/transaction.rs
ekexium: `Active` sounds good.

src/transaction/transaction.rs
ekexium: I don't think so. If the rollback request failed, we need to keep the data.

sticnarf: I don't think we need to keep the data on failure. Rollback failure rarely happens so it's ok to just leave these garbage on TiKV.

src/transaction/transaction.rs
ekexium: Read after commit is valid but not expected, I think.
I feel like allowing operations after commit is a weird behavior, especially if users don't know how we implement it. They don't know whether the data is buffered locally or sent to the server. If their writes succeed but they cannot commit or rollback, they're going to be confused. 
Besides, checking before every operation may help users detect their errors early.

src/transaction/transaction.rs
ekexium: My mistake, it's inside `TwoPhaseCommiter` which is not reused.

src/transaction/transaction.rs
ekexium: The check is in performed in `TwoPhaseCommitter`.

src/transaction/transaction.rs
nrc: Makes sense

src/transaction/transaction.rs
nrc: I think we might want to let the user retry a rollback, or at least we might want to let them retry a commit, and I think the commit and rollback APIs should be similar.

ekexium: So how about this:
Dropping `Trarnsaction` before any rollback or commit => panic,
Commit or rollback success => cannot do anything further
Commit or rollback failure => can only retry, will not panic on drop

nrc: This sounds good. In addition, I think `TwoPhaseCommitter` can consume `self` on rollback or commit?

sticnarf: :+1:

ekexium: Yeah. They already consumed `self`.

tikv-client-common/src/errors.rs
```diff
@@ -29,6 +29,11 @@ pub enum ErrorKind {
     /// Invalid key range to scan. Only left bounded intervals are supported.
     #[fail(display = "Only left bounded intervals are supported")]
     InvalidKeyRange,
+    /// It's not allowed to perform operations in a transaction after it has been committed or rolled back.
+    #[fail(
+        display = "Cannot read or write data after any attempt to commit or roll back the transaction"
```
nrc, line 34:
```suggestion
        display = "Cannot read or write data after transaction has been commited or rolled back"
```

src/transaction/transaction.rs
```diff
@@ -357,15 +389,56 @@ impl Transaction {
...
             self.timestamp.version(),
             self.bg_worker.clone(),
             self.rpc.clone(),
             self.for_update_ts,
         )
         .commit()
-        .await
+        .await;
+
+        if res.is_ok() {
```
nrc, line 410: I'm not sure about the logic here: if the user tries to commit and fails, why shouldn't they be able to modify the transaction before trying again to commit?

src/transaction/transaction.rs
```diff
@@ -9,9 +9,25 @@ use crate::{
...
 use derive_new::new;
 use futures::{executor::ThreadPool, prelude::*, stream::BoxStream};
-use std::{iter, mem, ops::RangeBounds, sync::Arc};
+use std::{iter, ops::RangeBounds, sync::Arc};
 use tikv_client_proto::{kvrpcpb, pdpb::Timestamp};
 
+#[derive(PartialEq)]
+enum TransactionStatus {
+    /// The transaction is read-only [`Snapshot`](super::Snapshot::Snapshot), no need to commit or rollback or panic on drop.
+    ReadOnly,
```
nrc, line 18: Do we need to know this explicitly? Could we instead check for an empty buffer?

src/transaction/transaction.rs
ekexium: I think there are 2 reasons:
1. Users seldom need to modify transactions after they attempted to commit or rollback. 
2. It's possible that commit or rollback succeed in the server side but return error in the client side. In such case performing more operations will lead to weird behavior.

tikv-client-common/src/errors.rs
ekexium: This is part of the logic discussion.

src/transaction/transaction.rs
nrc: I think 1 doesn't mean we should prevent it, but 2 is a good reason.

src/transaction/transaction.rs
ekexium: The buffer is protected by an async-aware mutex. So it brings difficulty in implementing `Drop` if we do this.

PR #197, opened 2020-11-09
Use async aware Mutex instead of std::mutex in async functions

ekexium:
We used `std::sync::Mutex` to protect `mutations` in `Buffer`. While the LockGuard is held across await points.

Changes:
Use `tokio::sync::Mutex` to replace it.


ekexium:
PTAL @nrc

nrc (approved):
lgtm

Issue #194, opened 2020-11-02
Tracking issue: documentation


nrc:
## For initial release

- [ ] Introduction (readme, include in rustdoc, check crate root docs)
- Public API of main crate
  - [ ] RawClient
  - [ ] ColumnFamily
  - [ ] TransactionClient
  - [ ] Snapshot
  - [ ] Transaction
  - [ ] Config
  - [ ] SecurityManager
  - [ ] BoundRange
  - [ ] Error
  - [ ] ErrorKind
  - [ ] Key
  - [ ] KvPair
  - [ ] Result
  - [ ] Timestamp
  - [ ] ToOwnedRange
  - [ ] Value
  - [ ] Region
  - [ ] RegionId
  - [ ] RegionVerId
  - [ ] StoreId
- [ ] examples (each example is introduced, well-commented, linked from docs)

## For later

- [ ] Public API of all crates

nrc:
There's still a lot of work to do here, but I believe that #269 gets us to a state where we can make a 0.1 release.

PR #198, opened 2020-11-10
Update inline doc for types

ekexium:
Added some inline doc to types.
Removed a few unused or unnecessary code.

Related issue: #194

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -10,13 +10,14 @@ pub use tikv_client_proto::pdpb::Timestamp;
 const PHYSICAL_SHIFT_BITS: i64 = 18;
 const LOGICAL_MASK: i64 = (1 << PHYSICAL_SHIFT_BITS) - 1;
 
-/// A helper trait to convert between [`Timestamp`](Timestamp) and an u64.
+/// A helper trait to convert a Timestamp to and from an u64.
 ///
-/// A `Timestamp` (64 bits) contains a physical part (first 46 bits) and a logical part (last 18 bits).
+/// Currently the only implmentation of this trait is [`Timestamp`](Timestamp) in TiKV.
+/// It contains a physical part (first 46 bits) and a logical part (last 18 bits).
 pub trait TimestampExt {
-    /// Combine physical and logical parts to get a single `Timestamp` (version).
+    /// Convert the timestamp to u64.
     fn version(&self) -> u64;
-    /// Decompose the `Timestamp` (version) into physical and logical parts.
+    /// Convert u64 to a timestamp.
     fn from_version(version: u64) -> Self;
 }
 
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -10,8 +10,13 @@ pub use tikv_client_proto::pdpb::Timestamp;
 const PHYSICAL_SHIFT_BITS: i64 = 18;
 const LOGICAL_MASK: i64 = (1 << PHYSICAL_SHIFT_BITS) - 1;
 
+/// A helper trait to convert between [`Timestamp`](Timestamp) and an u64.
+///
+/// A `Timestamp` (64 bits) contains a physical part (first 46 bits) and a logical part (last 18 bits).
 pub trait TimestampExt {
+    /// Combine physical and logical parts to get a single `Timestamp` (version).
     fn version(&self) -> u64;
+    /// Decompose the `Timestamp` (version) into physical and logical parts.
     fn from_version(version: u64) -> Self;
 }
 
--- a/src/transaction/client.rs
+++ b/src/transaction/client.rs
@@ -43,6 +43,7 @@ pub struct Client {
 impl Client {
     /// Creates a transactional [`Client`](Client).
     ///
+    /// # Examples
     /// ```rust,no_run
     /// use tikv_client::{Config, TransactionClient};
     /// use futures::prelude::*;
@@ -76,6 +77,7 @@ impl Client {
     ///
     /// For details, check our [SIG-transaction](https://github.com/tikv/sig-transaction/tree/master/doc/tikv#optimistic-and-pessimistic-transactions).
     ///
+    /// # Examples
     /// ```rust,no_run
     /// use tikv_client::{Config, TransactionClient};
     /// use futures::prelude::*;
@@ -97,6 +99,7 @@ impl Client {
     /// Write operations will lock the data until commit, thus commit requests should not suffer from write conflict.
     /// For details, check our [SIG-transaction](https://github.com/tikv/sig-transaction/tree/master/doc/tikv#optimistic-and-pessimistic-transactions).
     ///
+    /// # Examples
     /// ```rust,no_run
     /// use tikv_client::{Config, TransactionClient};
     /// use futures::prelude::*;
@@ -120,6 +123,7 @@ impl Client {
 
     /// Retrieves the current [`Timestamp`](Timestamp).
     ///
+    /// # Examples
     /// ```rust,no_run
     /// use tikv_client::{Config, TransactionClient};
     /// use futures::prelude::*;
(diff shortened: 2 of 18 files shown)


ekexium:
@nrc PTAL

tikv-client-common/src/errors.rs
```diff
@@ -189,7 +187,8 @@ impl From<tikv_client_proto::kvrpcpb::KeyError> for Error {
     }
 }
 
-/// A result holding an [`Error`](Error).
+/// The result type used in tikv-client.
+/// It holds an error type that contains varisous kinds of error that can happen in the crate.
```
nrc, line 191: I don't think we need either of these comments, they don't add much.

src/transaction/transaction.rs
```diff
@@ -346,6 +400,7 @@ impl Transaction {
 /// The default TTL of a lock in milliseconds
 const DEFAULT_LOCK_TTL: u64 = 3000;
 
+/// A struct wrapping the details of 2PC.
```
nrc, line 403: Could you expand this please?

src/transaction/transaction.rs
```diff
@@ -316,7 +368,9 @@ impl Transaction {
         .await
     }
 
-    /// Pessimistically lock the keys
+    /// Pessimistically lock the keys.
+    ///
+    /// Once resovled it acquires a mutex on the key in TiKV.
```
nrc, line 373: I think it is not really a mutex, just a logical lock

src/transaction/transaction.rs
```diff
@@ -271,8 +314,16 @@ impl Transaction {
         Ok(())
     }
 
-    /// Locks the given keys.
+    /// Lock the given keys without mutating value (at the time of commit).
     ///
+    /// In optimistic mode, write conflicts are not checked until commit.
+    /// So use this command to indicate that
+    /// "I do not want to commit if the value associated with this key has been modified".
+    /// It's useful to avoid *Write Skew* anomaly.
```
nrc, line 322:
```suggestion
    /// It's useful to avoid *write skew* anomaly.
```

src/transaction/transaction.rs
```diff
@@ -271,8 +314,16 @@ impl Transaction {
...
+    ///
+    /// In pessimistic mode, please use [`get_for_update`](Transaction::get_for_update) instead.
```
nrc, line 324: Do we use this command for pessimistic transactions in TiDB?

src/transaction/mod.rs
```diff
@@ -2,12 +2,11 @@
...
 //! Transactional related functionality.
 //!
-//! Using the [`TransactionClient`](TransactionClient) you can utilize TiKV's transactional interface.
+//! Using the [`TransactionClient`](client::Client) you can utilize TiKV's transactional interface.
 //!
 //! This interface offers SQL-like transactions on top of the raw interface.
 //!
-//! **Warning:** It is not advisable to use both raw and transactional functionality in the same keyspace.
-//!
+//! **Warning:** It is not advisable to use both raw and transactional functionalities in the same keyspace.
```
nrc, line 9: I think the original is correct

src/transaction/client.rs
```diff
@@ -15,7 +15,23 @@ use tikv_client_proto::{kvrpcpb, pdpb::Timestamp};
...
+/// and ends with a `rollback` or `commit` request.
+/// If a `Transaction` is dropped before it's rolled back or committed, it is automatically rolled back.
+///
+/// Transaction supports optimistic and pessimistic modes, for mroe deatils, check our
+/// [SIG-transaction](https://github.com/tikv/sig-transaction/tree/master/doc/tikv#optimistic-and-pessimistic-transactions).
+///
+/// Besides transaction, the client provides some utility methods:
+/// - `gc`: execute a GC process which clear stale data. It is not stablized yet.
+/// - `current_timestamp`: get the current `Timestamp`.
+/// - `snapshot`: get the [`Snapshot`](crate::transaction::Snapshot) of the database at a certain timestamp.
```
nrc, line 31: Are these public?

src/timestamp.rs
```diff
@@ -10,8 +10,13 @@ pub use tikv_client_proto::pdpb::Timestamp;
 const PHYSICAL_SHIFT_BITS: i64 = 18;
 const LOGICAL_MASK: i64 = (1 << PHYSICAL_SHIFT_BITS) - 1;
 
+/// A helper trait to convert between [`Timestamp`](Timestamp) and an u64.
+///
+/// A `Timestamp` (64 bits) contains a physical part (first 46 bits) and a logical part (last 18 bits).
 pub trait TimestampExt {
+    /// Combine physical and logical parts to get a single `Timestamp` (version).
     fn version(&self) -> u64;
+    /// Decompose the `Timestamp` (version) into physical and logical parts.
```
nrc, line 19: I think these two comments are incorrect - the functions convert to and from the u64 representation of a timestamp, to or from whatever implements this trait.

nrc (commented):
Looks good! A few comments/questions inline, nothing major.

src/transaction/transaction.rs
ekexium: Logically it's a mutually exclusive lock 😂. But yes it might be confusing.

src/transaction/transaction.rs
ekexium: I think we don't. `Lock` is used for `SELECT FOR UPDATE` in optimistic transactions.

src/transaction/client.rs
ekexium: They are

src/transaction/transaction.rs
nrc: I thought we could only select for update in pessimistic transactions, I'm probably wrong.

src/transaction/client.rs
nrc: Do they need to be?

ekexium: For now, I think they are intended to be public because they are not used by other parts.
`snapshot` is a wrapper of transaction to make it read-only. If it's not public we can remove it.
`gc` needs to be manually triggered by users and requires a `Timestamp`.
`current_timestamp` is the easiest way for users to get a timestamp. We don't provide convenience methods for users to calculate Timestamps. Maybe we should add these in the future.

src/transaction/transaction.rs
ekexium: Makes sense.

src/transaction/client.rs
nrc: Do users ever need to supply timestamps? I thought they were internal to the transaction? I guess we need to consider the lower level API.

Does Snapshot offer anything that Transaction doesn't?

src/transaction/transaction.rs
```diff
@@ -336,6 +337,9 @@ impl Transaction {
     /// # });
     /// ```
     pub async fn lock_keys(&self, keys: impl IntoIterator<Item = impl Into<Key>>) -> Result<()> {
+        if self.is_pessimistic {
+            return Err(ErrorKind::InvalidTransactionType.into());
+        }
```
nrc, line 342: This change should be in a separate PR, I think

src/transaction/client.rs
ekexium: Timestamps are needed for GC and snapshot. 

In GC users should supply a safe point. In a TiDB cluster there will be only one node that performs GC and it controls when to do it. In a TiKV client we cannot guarantee this, so I think it should be the user's choice.

`Snapshot` lets you read from any given timestamp, while `Transaction` cannot. Also the timestamp is related to GC safe point, which could be another supportive reason for users providing safe point for GC.

src/transaction/transaction.rs
ekexium: OK

src/transaction/client.rs
nrc: OK, thanks!

PR #201, opened 2020-11-23
Improve ergonomics

andylokandy:
This PR contains two changes:

1. move `key_only` from client to scan.
2. move `pd_endpoints` from config to client.

IMO, these two are a bit confusing and not so idiomatic to rust.

### DIFF
--- a/examples/pessimistic.rs
+++ b/examples/pessimistic.rs
@@ -14,13 +14,13 @@ async fn main() {
     // Create a configuration to use for the example.
     // Optionally encrypt the traffic.
     let config = if let (Some(ca), Some(cert), Some(key)) = (args.ca, args.cert, args.key) {
-        Config::new(args.pd).with_security(ca, cert, key)
+        Config::default().with_security(ca, cert, key)
     } else {
-        Config::new(args.pd)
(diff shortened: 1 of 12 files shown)


nrc (commented):
I like the change moving pd endpoints from config to client. I'm not sure about the change to key_only. The philosophy with the design is that options which are not frequently used (key_only, cf) should not have to be specified since that causes boilerplate on every call. What are your thoughts on moving it?

ekexium:
How about using two different methods `scan` and `scan_keys`?

andylokandy:
@nrc Not similar to the cf, the key_only only affects scans. @ekexium 's solution seems better.

Ps: the ultimate solution is named/default arguments... no idea why rust doesn't have.

andylokandy:
One more reason for the key_only is that we can't scan key_only and kvpair in one transaction.

andylokandy:
@ekexium PTAL

ekexium:
lgtm

nrc:
OK, I like the scan/scan_keys solution. I am a bit wary of having lots of subtly different functions, but this one seems fine.

PR #202, opened 2020-11-25
Allow unbounded start key

andylokandy:
Unbounded start key will be translated to empty key.


nrc:
PTAL @ekexium

nrc (approved):
lgtm

ekexium:
lgtm. Can also replace all usages of `vec![]..` in tests and docs with `..` now.

PR #203, opened 2020-12-04
Initial work on async commit

nrc:
This is mostly prewrite and commit, there is no recovery code. It is also missing some features: fallback, setting min and max commit ts, etc. but it is a start and it seems to work.

PTAL @andylokandy @ekexium

### DIFF
--- a/src/transaction/requests.rs
+++ b/src/transaction/requests.rs
@@ -289,9 +289,10 @@ pub fn new_cleanup_request(key: impl Into<Key>, start_version: u64) -> kvrpcpb::
 
 #[async_trait]
 impl KvRequest for kvrpcpb::PrewriteRequest {
-    type Result = kvrpcpb::PrewriteResponse;
+    type Result = Vec<kvrpcpb::PrewriteResponse>;
     type RpcResponse = kvrpcpb::PrewriteResponse;
     type KeyData = Vec<kvrpcpb::Mutation>;
+
     fn make_rpc_request(&self, mutations: Self::KeyData, store: &Store) -> Result<Self> {
         let mut req = self.request_from_store(store)?;
 
@@ -331,14 +332,12 @@ impl KvRequest for kvrpcpb::PrewriteRequest {
     }
 
     fn map_result(response: Self::RpcResponse) -> Self::Result {
-        response
+        // FIXME this is clown-shoes inefficient.
+        vec![response]
     }
 
     async fn reduce(results: BoxStream<'static, Result<Self::Result>>) -> Result<Self::Result> {
-        results
-            .into_future()
-            .map(|(f, _)| f.expect("no results should be impossible"))
-            .await
+        results.try_concat().await
     }
 }
 
--- a/src/transaction/requests.rs
+++ b/src/transaction/requests.rs
@@ -289,11 +289,25 @@ pub fn new_cleanup_request(key: impl Into<Key>, start_version: u64) -> kvrpcpb::
 
 #[async_trait]
 impl KvRequest for kvrpcpb::PrewriteRequest {
-    type Result = ();
+    type Result = kvrpcpb::PrewriteResponse;
     type RpcResponse = kvrpcpb::PrewriteResponse;
     type KeyData = Vec<kvrpcpb::Mutation>;
     fn make_rpc_request(&self, mutations: Self::KeyData, store: &Store) -> Result<Self> {
         let mut req = self.request_from_store(store)?;
+
+        if self.use_async_commit {
+            // Only need to set secondary keys if we're sending the primary key.
+            if mutations.iter().any(|m| m.key == self.primary_lock) {
+                req.set_secondaries(self.secondaries.clone());
+            }
+            req.set_use_async_commit(true);
+            req.set_max_commit_ts(self.max_commit_ts);
+        }
+        // Only if there is only one request to send
+        if mutations.len() == self.secondaries.len() + 1 {
+            req.set_try_one_pc(self.try_one_pc);
+        }
+
         req.set_mutations(mutations);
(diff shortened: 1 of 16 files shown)


src/transaction/requests.rs
```diff
@@ -315,11 +330,14 @@ impl KvRequest for kvrpcpb::PrewriteRequest {
...
 
-    fn map_result(_: Self::RpcResponse) -> Self::Result {}
+    fn map_result(response: Self::RpcResponse) -> Self::Result {
+        response
+    }
 
     async fn reduce(results: BoxStream<'static, Result<Self::Result>>) -> Result<Self::Result> {
         results
-            .try_for_each_concurrent(None, |_| future::ready(Ok(())))
+            .into_future()
```
ekexium, line 339: I don't get it why it's changed to `into_future`. We may have multiple prewrite requests.

src/transaction/transaction.rs
```diff
@@ -28,6 +28,55 @@ enum TransactionStatus {
...
+
+    pub fn async_commit(mut self) -> TransactionStyle {
+        self.async_commit = true;
+        self
+    }
+
+    fn push_for_update_ts(&mut self, for_update_ts: u64) {
+        match &mut self.kind {
+            TransactionKind::Optimistic => {
+                self.kind = TransactionKind::Pessimistic(for_update_ts);
```
ekexium, line 70: I think push update ts for optimistic transactions is invalid, and it won't happen in the current implementation. Why don't just panic?

src/transaction/transaction.rs
```diff
@@ -542,62 +597,95 @@ struct TwoPhaseCommitter {
...
-            Err(e) => {
-                if self.undetermined {
-                    Err(Error::undetermined_error(e))
-                } else {
-                    Err(e)
+
+        let min_commit_ts = self.prewrite().await?;
+
+        if self.style.try_one_pc {
+            return Ok(min_commit_ts);
```
ekexium, line 614: Can we just return here? I think `try_one_pc == true` doesn't mean we really use 1PC.

ekexium (commented):
I have a few questions. Others LGTM.

Why 1PC is configured for clients while async commit is for transaction?

Also, I feel that now the requests created in the client are different from those sent to regions. The logic exists in both places. It doesn't look so elegant to me, though I don't know if we can refactor it.

src/transaction/transaction.rs
```diff
@@ -542,62 +597,95 @@ struct TwoPhaseCommitter {
...
+            TransactionKind::Pessimistic(_) => {
+                request
+                    .execute(self.rpc.clone(), PESSIMISTIC_BACKOFF)
+                    .await?
+            }
+        };
+
+        if self.style.try_one_pc {
+            if response.one_pc_commit_ts == 0 {
+                return Err(ErrorKind::OnePcFailure.into());
```
andylokandy, line 682: Will `OnePcFailure` fallback to 2PC (in next PR)?

nrc:
"Why 1PC is configured for clients while async commit is for transaction?" - whether a transaction can use 1pc or not is not decided by the user, but by the client, so the configuration is whether to *try* 1pc or not. Whether to use async commit is up to the user - they might choose to do it for some transactions but not others based on the size of the transaction or load on the system, etc.

src/transaction/requests.rs
nrc: Whoops, yeah that's wrong

src/transaction/transaction.rs
nrc: Hmm, there is nothing actually stopping you call pessimistic_lock on an optimistic transaction, but I can add an assert

src/transaction/transaction.rs
nrc: I should have set it to false if it wasn't used.

src/transaction/transaction.rs
nrc: Yes, I think so, at least as an option

nrc:
"The logic exists in both places." - what does this mean? I think there is logic in both places, but it is not the same logic, there are different things to check.

nrc:
@ekexium @andylokandy comments addressed, please take another look

src/transaction/transaction.rs
andylokandy: May be better to rename it to one_pc after it's determined to use one pc or not.

ekexium:
> "The logic exists in both places." - what does this mean? I think there is logic in both places, but it is not the same logic, there are different things to check.

Yes. I just mean the logic is distributed in two places, and the requests sent to servers are different from those created in the `Transaction`, which seems a bit confusing.

ekexium:
CI uses `pingcap/latest`, its failure is expected. But the test `txn_bank_transfer` also failed in my local environment using `pingcap/nightly` image. Don't know why.

nrc:
> Yes. I just mean the logic is distributed in two places, and the requests sent to servers are different from those created in the Transaction, which seems a bit confusing.

Ah, I see. I think it has to be this way since in Transaction we only create a single request but lower down it is split into multiple requests, one per region (for all requests and prior to this PR). Now we do a bit more than just splitting by region, but with the current architecture I don't think we can do it differently.

We could refactor so that the transaction creates multiple requests (one per store), but I fear we might end up with a lot of duplicated code. I think we should handle responses better too - we get one response per store, but we are forced to allocate a Vec per store at the moment if we want to return all the responses.

`txn_bank_transfer` passes locally for me. I'm not sure why it would fail. I am building my own TiKV and PD from master.

ekexium:
All right. LGTM. Probably there are some issues in the docker image.

andylokandy:
LGTM

PR #210, opened 2020-12-11
fix using wrong primary key

### LINKED ISSUE #204: Requesting a pessimistic lock does not use the correct primary key
It uses the first key of the action, not the primary key of the transaction.

longfangsong:
Should fix #204
Signed-off-by: longfangsong <longfangsong@icloud.com>

### DIFF
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -17,14 +17,14 @@ struct Mutations {
 impl Mutations {
     fn insert(&mut self, key: impl Into<Key>, mutation: Mutation) {
         let key = key.into();
-        if self.primary_key.is_none() {
-            self.primary_key = Some(key.clone());
+        if !matches!(mutation, Mutation::Cached(_)) {
+            self.primary_key.get_or_insert_with(|| key.clone());
         }
         self.key_mutation_map.insert(key, mutation);
     }
 
-    pub fn get_primary_key_or(&mut self, key: Key) -> Key {
-        self.primary_key.get_or_insert(key).clone()
+    pub fn get_primary_key_or(&mut self, key: &Key) -> &Key {
+        self.primary_key.get_or_insert(key.clone())
     }
 }
 
@@ -40,8 +40,8 @@ impl Buffer {
         self.mutations.lock().await.primary_key.clone()
     }
     /// Get the primary key of the buffer, if not exists, use `key` as the primary key.
-    pub async fn get_primary_key_or(&self, key: Key) -> Key {
-        self.mutations.lock().await.get_primary_key_or(key)
+    pub async fn get_primary_key_or(&self, key: &Key) -> Key {
+        self.mutations.lock().await.get_primary_key_or(key).clone()
     }
 
     /// Get a value from the buffer. If the value is not present, run `f` to get
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -8,13 +8,38 @@ use std::{
 use tikv_client_proto::kvrpcpb;
 use tokio::sync::Mutex;
 
+#[derive(Default)]
+struct Mutations {
+    primary_key: Option<Key>,
+    key_mutation_map: BTreeMap<Key, Mutation>,
+}
+
+impl Mutations {
+    fn insert(&mut self, key: impl Into<Key>, mutation: Mutation) {
+        let key = key.into();
+        if self.primary_key.is_none() {
+            self.primary_key = Some(key.clone());
+        }
+        self.key_mutation_map.insert(key, mutation);
+    }
+
+    pub fn get_primary_key_or(&mut self, key: Key) -> Key {
+        self.primary_key.get_or_insert(key).clone()
+    }
+}
+
 /// A caching layer which buffers reads and writes in a transaction.
 #[derive(Default)]
 pub struct Buffer {
-    mutations: Mutex<BTreeMap<Key, Mutation>>,
+    mutations: Mutex<Mutations>,
 }
 
 impl Buffer {
+    /// Get the primary key of the buffer, if not exists, use `key` as the primary key.
+    pub async fn get_primary_key_or(&self, key: Key) -> Key {
+        self.mutations.lock().await.get_primary_key_or(key)
+    }
+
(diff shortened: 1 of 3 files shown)


src/transaction/transaction.rs
```diff
@@ -497,7 +499,10 @@ impl Transaction {
             .map(|it: Key| it.into())
             .collect();
         keys.sort();
-        let primary_lock = keys[0].clone();
+        let primary_lock = self
+            .primary_key
+            .get_or_insert_with(|| keys[0].clone())
+            .clone();
```
nrc, line 505: This is a bit odd because it is the only place we set the primary key, so optimistic transactions will not get a primary key, which they should. As I understand it, we already deal with primary keys in the buffer, so I think here we should use that primary key (or keys[0] if it is None).

longfangsong: > This is a bit odd because it is the only place we set the primary key, so optimistic transactions will not get a primary key, which they should. As I understand it, we already deal with primary keys in the buffer, so I think here we should use that primary key (or keys[0] if it is None).

After some thought, I'm afraid currently `Buffer` can't handle this right. 
`Buffer` use a `BTreeMap` to store mutations, and regard the minium key in it as the primary key (see `to_proto_mutations` and `commit_primary`), but for example:
- We lock ["b", "c"] first,  since buffer is empty now, we'll use "b" as primary key and lock them.
- Then we want to lock ["a"], this seems work well, but after lock and "a" put into the buffer, "a" become the primary key.
- Then other lock requests will regard "a" as primary key, which is different with the first lock request.
So it seems we'll still need a `primary_key` field, maybe I'll put it in the `Buffer` struct.

longfangsong:
@nrc PTAL

src/transaction/buffer.rs
```diff
@@ -163,10 +191,14 @@ impl Buffer {
...
+        let mutations = self.mutations.lock().await;
+        let (primary, other) = mutations
+            .key_mutation_map
             .iter()
+            .partition::<Vec<_>, _>(|(key, _)| *key == mutations.primary_key.as_ref().unwrap());
+        primary
+            .into_iter()
+            .chain(other.into_iter())
             .filter_map(|(key, mutation)| mutation.to_proto_with_key(key))
             .collect()
```
longfangsong, line 203: This is really ugly, do we really need these mutations sorted? Or we can replace the `BTreeMap` with a naive `Vec`.

nrc: I think we need to keep the BTreeMap because we look up writes in our cache before every read, which could be expensive if the data is not sorted.

I think we should either not promise that the primary key is the first element here, or we could have a more complex data structure, like using a Vec of mutations and a BTree index of the keys.

nrc:
@longfangsong is this ready for re-review?

longfangsong:
> @longfangsong is this ready for re-review?
Yes, though I'm still trying to find out what's wrong with the CI.

ekexium:
> > @longfangsong is this ready for re-review?
> 
> Yes, though I'm still trying to find out what's wrong with the CI.

It's probably because of #181. We can try to change them back to nightly

Cargo.toml
```diff
@@ -12,20 +12,21 @@ edition = "2018"
...
 # Enable integration tests with a running TiKV and PD instance.
 # Use $PD_ADDRS, comma separated, to set the addresses the tests use. 
-integration-tests = []
+integration-tests = ["failure"]
 
 [lib]
 name = "tikv_client"
 
 [dependencies]
+failure = { version = "0.1.8", optional = true }
```
nrc, line 21: We shouldn't need failure, we're using thiserror now

src/transaction/buffer.rs
```diff
@@ -8,13 +8,42 @@ use std::{
...
+struct Mutations {
+    primary_key: Option<Key>,
+    key_mutation_map: BTreeMap<Key, Mutation>,
+}
+
+impl Mutations {
+    fn insert(&mut self, key: impl Into<Key>, mutation: Mutation) {
+        let key = key.into();
+        if self.primary_key.is_none() {
+            self.primary_key = Some(key.clone());
```
nrc, line 21: Could use `get_or_insert_with` rather than `if` and an assignment

src/transaction/buffer.rs
```diff
@@ -8,13 +8,42 @@ use std::{
...
+
 /// A caching layer which buffers reads and writes in a transaction.
 #[derive(Default)]
 pub struct Buffer {
-    mutations: Mutex<BTreeMap<Key, Mutation>>,
+    mutations: Mutex<Mutations>,
 }
 
 impl Buffer {
+    /// Get the primary key of the buffer, if not exists, use `key` as the primary key.
```
nrc, line 38: Doc is incorect

src/transaction/buffer.rs
```diff
@@ -8,13 +8,42 @@ use std::{
...
+    pub async fn get_primary_key(&self) -> Option<Key> {
+        self.mutations.lock().await.primary_key.clone()
```
nrc, line 40: Can we return `Option<&Key>` and let the caller clone?

src/transaction/buffer.rs
```diff
@@ -8,13 +8,42 @@ use std::{
...
+    }
+    /// Get the primary key of the buffer, if not exists, use `key` as the primary key.
+    pub async fn get_primary_key_or(&self, key: Key) -> Key {
```
nrc, line 43: Can we take the key by reference rather than value so the caller doesn't need to clone, then only clone if there is no primary key?

nrc: Also return a reference so the caller can clone

nrc (commented):
Looks good, a few minor comments inline

src/transaction/buffer.rs
longfangsong: We do can pass the param by reference, but it's hard to return a reference here due to the mutex.

nrc (approved):
lgtm

longfangsong:
@ekexium PTAL

ekexium (commented):
LGTM

ti-srebot (approved):

PR #211, opened 2020-12-15
Use thiserror and rename Error to ClientError

andylokandy:


nrc (commented):
Mostly looks great. I think though that we should not rename Error to ClientError. Users of the crate can either use it prefixed (`client::Error`) or aliased (`use client::Error as ClientError`) there is no need for us to give it a long name

andylokandy:
> Mostly looks great. I think though that we should not rename Error to ClientError. Users of the crate can either use it prefixed (`client::Error`) or aliased (`use client::Error as ClientError`) there is no need for us to give it a long name

Good point! I've updated it, please take a look.

nrc (approved):
lgtm, thanks for the change

PR #214, opened 2020-12-24
 CI: use nightly TiKV and PD images and run examples

ekexium:
We started to use the `latest` channel in #181, now it's time to use `nightly` since we are implementing async commit.

We should also run examples in CI to make sure they are up-to-date when making breaking changes.


nrc (approved):
lgtm, thanks!

PR #219, opened 2021-01-08
Fix deadlock in pd cluster 

ekexium:
The `RWlock` in pd cluster can cause deadlock. 

Changes:
Drop the read lock guard before trying to acquire write lock in `reconnect`.

### DIFF
--- a/src/pd/retry.rs
+++ b/src/pd/retry.rs
@@ -54,16 +54,14 @@ macro_rules! retry {
         let stats = pd_stats($tag);
         let mut last_err = Ok(());
         for _ in 0..LEADER_CHANGE_RETRY {
-            let res = {
-                let $cluster = &$self.cluster.read().await.0;
-                let res = $call.await;
-                res
-            };
-
-            match stats.done(res) {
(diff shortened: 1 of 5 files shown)


src/pd/retry.rs
```diff
@@ -59,6 +59,8 @@ macro_rules! retry {
                 Ok(r) => return Ok(r),
                 Err(e) => last_err = Err(e),
             }
+            
+            drop($cluster);
```
nrc, line 63: I think this needs a comment because changing it will only cause an error at runtime, so it's not obvious why it is here. I think it might be cleared to use a block rather than an explicit `drop`, but there should be a comment either way

andylokandy: It was intended to resolve a lifetime problem on block style. And I've figured out a solution, so no drop will be needed.

andylokandy (approved):

nrc (approved):
lgtm

ti-srebot (approved):

nrc:
@andylokandy your commits need the DCO signoff

PR #222, opened 2021-01-11
Add an option to control the behavior when txns get dropped

ekexium:
Currently, when a transaction is dropped without an attempt to commit or rollback, it panics. It prevents users using the `some_request.await?` pattern. And sometimes we don't care if it's dropped gracefully, e.g. in tests.

Changes:
- Add an option to `TransactionOptions` to control this. We can choose to warn instead of panicking.
- Fix a test that may cause such behavior.


src/transaction/transaction.rs
```diff
@@ -547,6 +554,14 @@ pub struct TransactionOptions {
...
     /// How to retry in the event of certain errors.
     retry_options: RetryOptions,
+    /// What to do if the transaction is dropped without an attempt to commit or rollback
+    check_level: CheckLevel,
+}
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum CheckLevel {
+    Panic,
+    Warn,
```
nrc, line 564: We should have a `None` level, IMO

src/transaction/transaction.rs
```diff
@@ -614,6 +631,17 @@ impl TransactionOptions {
         self
     }
 
+    /// Whether to be strict when transaction is dropped without an attempt to commit or rollback,
+    /// `true` => panic, `false` => warning.
+    pub fn drop_check(mut self, be_strict: bool) -> TransactionOptions {
```
nrc, line 636: Make the enum pub and take it as an argument, rather than using a bool

andylokandy:
What about adding an `auto_rollback` instead?

nrc:
> Or make it always rollback on drop

We can't do async stuff on drop, and in general we should avoid doing significant work in drop - we can't propagate errors or panic, and it can be a bit surprising.

src/transaction/transaction.rs
```diff
@@ -620,6 +639,13 @@ impl TransactionOptions {
         self
     }
 
+    /// Whether to be strict when transaction is dropped without an attempt to commit or rollback,
+    /// `true` => panic, `false` => warning.
```
nrc, line 643: Out of date comment

nrc (approved):
lgtm with comment fixed.

PR #229, opened 2021-01-15
Add key_exists method

andylokandy:
`key_exists` is a sugar and it will check whether the key exists. It can also be used to resolve lock by TiFlash.


nrc (commented):
Looks good, could you add a usage to either the integration tests or examples please?

nrc (approved):

Issue #209, opened 2020-12-10
Use CheckTransactionStatus rather than Cleanup


nrc:
Cleanup is deprecated from v4.0 onwards.

I've started implementation (along with async commit lock resolution) but have got a bit bogged down with the retry logic).

eduralph:
Done by the #519 resolver rework — lock resolution now goes through `check_txn_status` (with `verify_is_primary` set). Can this be closed?

Issue #208, opened 2020-12-10
Give the user more control over lock resolution


nrc:
Currently, we automatically resolve locks and retry (without bumping the start ts which would be unsound). This is ok as a default, but it would be nice to also allow the user to have more control, i.e., returning an error with the locks and providing a resolve_locks function.

I have started implementing this, but I'm not sure what the API should look like - i.e., how does the user opt-in to auto-retry? There are also lots of details in the retry code to work out.

Issue #206, opened 2020-12-08
Refactor handling of response types


nrc:
The pipeline at the moment takes a response from each region, converts it to the API type, then combines it into a single response. This is sub-optimal if the user wants multiple responses (and I bet we're dropping some errors which we should pass to the user). The combination step has types which mean that if we want to combine individual responses into a `Vec`, we need to allocate a `Vec` for each response and then concatenate them.

PR #232, opened 2021-02-02
Refactor the command abstraction

### LINKED ISSUE #205: The API of transaction::requests is inconsistent
Mostly the arguments to the `new_...` functions: some take high level type (e.g., `Key`), some take low level ones (e.g., `Vec<u8>`) and some take converting types (e.g., `impl Into<Key>`). I think they should all take high-level types and do the conversions internally.

Some of the names have an unnecessary `mvcc` in them.

We should check that after those changes there is nothing more to do.

nrc:
This is a major refactoring to the client's internals. The main point of the refactoring is to replace the old KvRequest trait and associated code with an a cleaner abstraction. The new abstraction is a plan, which is a strategy for executing a request. Plans are composable and more self-contained. They are also more flexible and in some cases more efficient than the old system.

As well as the major refactoring (in the last commit) there are a bunch of smaller, tidying-up refactorings in earlier commits.

Closes #205, #206. It is a prerequisite for #111, #208, and #209.

PTAL @andylokandy @ekexium


src/request/plan_builder.rs
```diff
@@ -0,0 +1,178 @@
...
+                phantom: PhantomData,
+            },
+            phantom: PhantomData,
+        }
+    }
+
+    /// Apply a processing step to a response (usually only needed if the request is sent to a
+    /// single region because post-porcessing can be incorporated in the merge step for multi-region
+    /// requests).
+    pub fn post_process(self) -> PlanBuilder<PdC, ProcessResponse<P, P::Result>, Ph>
```
andylokandy, line 109: What does this mean?

src/request/plan_builder.rs
```diff
@@ -0,0 +1,178 @@
...
+    plan.kv_client = Some(store.client);
+    Ok(PlanBuilder {
+        plan,
+        pd_client,
+        phantom: PhantomData,
+    })
+}
+
+/// Indicates that a request operates on a single key.
+pub trait SingleKey {
```
andylokandy, line 175: Who impls this trait?

src/store.rs
```diff
@@ -77,36 +49,61 @@ where
...
+                store,
+            )
         })
         .into_stream()
         .boxed()
 }
 
 /// The range used for request should be the intersection of `region_range` and `range`.
-fn bound_range(region_range: (Key, Key), range: BoundRange) -> (Key, Key) {
+fn bound_range(region_range: (Key, Key), range: (Key, Key)) -> (Key, Key) {
```
andylokandy, line 91: Rename to `range_intersect`?

src/store.rs
```diff
@@ -4,69 +4,41 @@ use crate::{pd::PdClient, BoundRange, Key, Region, Result};
...
-        .store_for_key(key_data.as_ref().clone())
-        .map_ok(move |store| (key_data, store))
-        .into_stream()
-        .boxed()
-}
-
 /// Maps keys to a stream of stores. `key_data` must be sorted in increasing order
-pub fn store_stream_for_keys<KeyData, IntoKey, I, PdC>(
-    key_data: I,
+pub fn store_stream_for_keys<K, K2, PdC>(
```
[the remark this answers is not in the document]
nrc: So that you can, e.g., pass in a KvPair and get out just the key.

src/request/plan_builder.rs
nrc: Which part needs clarification? The use case is something like you pass a `RawGetResponse` in and get a `Value` out, i.e., it is discarding some unneeded information and transforming types to something a little bit higher level.

src/request/plan_builder.rs
nrc: It is impl'ed for requests, e.g., `RawGetRequest: SingleKey`, it probably never needs to be impl'ed by users of the client, only implementors (i.e., we would need to add it for newly supported requests), but users of the crate might need to use it as a bound if they are doing something fancy.

src/store.rs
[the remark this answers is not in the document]
andylokandy: I see, so could we rename it to `Ret` or `Res`?

src/request/plan_builder.rs
andylokandy: I mean, we can do this by default, as there will be no need to call this explicitly,

src/request/plan_builder.rs
andylokandy: I still can't find any impl for requests by text-search. What did I miss?

ekexium (commented):
I haven't comprehensively read all the changes. Mostly LGTM. Some of my first thoughts are:

In the new pattern, I think it's easy to miss some steps or put the steps in an incorrect order. Are there any methods that help avoid this?

I tried running integration tests. It failed with `thread 'crud' panicked at '`async fn` resumed after completion'` and other kinds of errors.

src/request/plan_builder.rs
ekexium: @andylokandy The impls are in raw/requests.rs and transaction/requests.rs. GitHub hides them 🤦🏻‍♂️

src/request/plan_builder.rs
nrc: I'm not sure what you mean, we do this by default, but for users who want to go a non-default path, they will need to call it.

nrc:
> In the new pattern, I think it's easy to miss some steps or put the steps in an incorrect order. Are there any methods that help avoid this?

I use phantom types to ensure that a request is always targetted and that is the only step which if missed will cause a programming error. Other missed/out of order steps might cause logic errors, but it is also possible that the user wants those errors (e.g., if writing a tool to chaos test TiKV). I also tried a few ways to ensure the correct order, but they were pretty ugly and it was difficult to make them optional. I think that it is hard to make an error if you use the transaction API, and if you use the Plan API, then you get more flexibility but also more scope for errors.

TszKitLo40:
Can you please resolve the conflict?
And it seems that there is also some mismatched types

nrc:
I think this should pass integration tests, however, it should not land until we've resolved #235.

PTAL @andylokandy @ekexium

src/lib.rs
```diff
@@ -6,6 +6,8 @@
 #![allow(clippy::type_complexity)]
 #![allow(incomplete_features)]
 #![cfg_attr(test, feature(specialization))]
```
ekexium, line 8: So we won't support build on the stable channel?

src/request/plan.rs
```diff
@@ -0,0 +1,236 @@
...
+
+/// A technique for merging responses into a single result (with type `Out`).
+pub trait Merge<In>: Sized + Clone + Send + Sync + 'static {
+    type Out: Send;
+
+    fn merge(&self, input: Vec<Result<In>>) -> Result<Self::Out>;
+}
+
+#[derive(Clone)]
+pub struct MergeResponse<P: Plan, In, M: Merge<In>> {
```
ekexium, line 95: Some plans are named after "xxxPlan", some are "xxxResponse". Could we use a similar pattern for the names?

ekexium (commented):
I tried 
1. `PD_ADDRS="127.0.0.1:2379" cargo test --package tikv-client --test integration_tests --features integration-tests -- --nocapture --ignored`, which writes a lot of data and forces region splitting.
2. and then `PD_ADDRS="127.0.0.1:2379" cargo test --package tikv-client --test integration_tests --features integration-tests -- --nocapture`

Then `txn_bank_transfer` failed. **I'm not sure** if it's a bug, because the ignored test is not supposed to run with transactional tests (may cause invalid region boundaries for transactional API). But in `master` it passed, so there might be some problem. 

Anyway I think we need to add some multi-region tests in CI.

The `merge` is composable while `post_process` is bound to the response type. Is it possible that we need different post_process for the same type of request in the future?

ekexium (commented):
I still get ```panicked at '`async fn`` resumed after completion', /home/eke/code/client-rust/src/pd/client.rs:104:24``` errors.
My steps to reproduce:
1. `tiup playground nightly --db 0 --tiflash 0 --monitor false`
2. The first few runs often failed with such errors.

src/transaction/requests.rs
```diff
@@ -586,46 +314,73 @@ pub fn new_pessimistic_lock_request(
...
+        let mut mutations = self.mutations.clone();
+        mutations.sort_by(|a, b| a.key.cmp(&b.key));
+        store_stream_for_keys(mutations.into_iter(), pd_client.clone())
     }
 
-    fn map_result(mut result: Self::RpcResponse) -> Self::Result {
-        result.take_locks()
+    fn apply_shard(&mut self, shard: Self::Shard, store: &Store) -> Result<()> {
+        self.set_context(store.region.context()?);
+        self.set_mutations(shard.into_iter().map(Into::into).collect());
```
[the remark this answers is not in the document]
nrc: good catch!

src/transaction/requests.rs
```diff
@@ -180,197 +130,67 @@ impl HasLocks for kvrpcpb::ScanResponse {
...
-    }
-}
-
 pub fn new_resolve_lock_request(
-    context: kvrpcpb::Context,
     start_version: u64,
     commit_version: u64,
 ) -> kvrpcpb::ResolveLockRequest {
     let mut req = kvrpcpb::ResolveLockRequest::default();
-    req.set_context(context);
```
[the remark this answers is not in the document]
nrc: We only set the context as a way to pass in which store to send to, in the new system that is done explicitly, so the context hack is no longer needed

src/store.rs
```diff
@@ -77,36 +49,61 @@ where
...
+                region_range,
+                (range.0.clone().into(), range.1.clone().into()),
+            );
+            ((result_range.0.into(), result_range.1.into()), store)
+        })
+        .into_stream()
+        .boxed()
+}
+
+pub fn store_stream_for_range_by_start_key<PdC: PdClient>(
```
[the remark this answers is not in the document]
nrc: it almost can, but making the return type correct is tricky. I'm pretty sure it can be done with some fancy generics and a couple of closures, but life is too short

src/transaction/requests.rs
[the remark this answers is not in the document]
ekexium: I get it

nrc:
One issue with the integration tests is that you must leave the servers some time to start up. If you immediately start the servers, then run integration tests, there are mysterious errors because the cluster does not pass validation. However, if you wait a while (about 15s seems to be enough on my machine), the tests complete without errors.

nrc:
> Anyway I think we need to add some multi-region tests in CI.

Agree.

> I noticed the merge is composable while post_process is bound to the response type. Is it possible that we may need different post_process for the same type of request in the future?

Still pondering this, otherwise, I've addressed all the feedback and tests are passing.

src/pd/client.rs
```diff
@@ -103,7 +103,9 @@ pub trait PdClient: Send + Sync + 'static {
             let this = self.clone();
             async move {
                 let start_key = match start_key {
-                    None => return Ok(None),
+                    None => {
```
ekexium, line 106: Can remove the curly brackets

ekexium (approved):
LGTM

src/request/plan.rs
```diff
@@ -0,0 +1,303 @@
...
+            if self.backoff.is_none() {
+                return Err(Error::ResolveLockError);
+            }
+
+            let pd_client = self.pd_client.clone();
+            if resolve_locks(locks, pd_client.clone()).await? {
+                result = self.inner.execute().await?;
+            } else {
+                let mut clone = self.clone();
+                match clone.backoff.next_delay_duration() {
```
ekexium, line 235: The `clone` is always cloned from `self`, thus the status of `backoff` is always at its initial state.

PR #233, opened 2021-02-05
Support mutation Insert and CheckNotExists

### LINKED ISSUE #217: Support Insert and CheckNotExist mutations
This issue proposes to add `insert(key, value)` and `check_not_exist()` to `Transaction`.

`Insert` and `CheckNotExist` are special mutations that will check that the key should not exist until commit. The difference between them is that `CheckNotExist` doesn't carry a value and only works in the optimistic transaction while `Insert` should work in both. 

learning materials: https://github.com/tikv/sig-transaction
gPRC definition: https://github.com/pingcap/kvproto/blob/f3dafca4c7fd5af12fd83b9cf8540b1a32f61c02/proto/kvrpcpb.proto#L928

ziyi-yan:
Closes #217

PS: This PR is submitted as my code task for applying LFX mentorship 2021 Spring.


andylokandy:
Reference https://github.com/tikv/client-rust/issues/217

andylokandy (approved):

src/transaction/buffer.rs
```diff
@@ -274,6 +298,10 @@ enum BufferEntry {
     Put(Value),
     // Value has been deleted.
     Del,
+    // Key should be check not exists before.
```
ekexium, line 301: Could you also update the doc of `BufferEntry`?

src/transaction/buffer.rs
```diff
@@ -191,10 +191,28 @@ impl Buffer {
             .await
             .insert(key, BufferEntry::Put(value));
     }
+    
+    /// Mark a value as Insert mutation into the buffer (does not write through).
+    pub async fn insert(&self, key: Key, value: Value) {
+        self.mutations
+            .lock()
+            .await
+            .insert(key, BufferEntry::Insert(value));
```
ekexium, line 200: What happens if there exists a mutation of the key when we insert? Should we return an error instead of overwriting it?

src/transaction/buffer.rs
```diff
@@ -274,6 +298,10 @@ enum BufferEntry {
     Put(Value),
     // Value has been deleted.
     Del,
+    // Key should be check not exists before.
+    Insert(Value),
+    // Key should be check not exists before.
+    CheckNotExist,
```
ekexium, line 304: I think we can also provide a `check_not_exist` method.

ekexium (commented):
Thanks for the PR! I have a few questions.

And make your code pass `cargo fmt --all` and `cargo clippy -- -D clippy::all` please.

src/transaction/buffer.rs
ziyi-yan: Done. I add a new pub method to buffer to make it possible for `Transaction` to check whether the key has existed and handle it returning a new kind of error.

src/transaction/buffer.rs
ziyi-yan: It's kind of unclear for me to see any possible use cases of a `check_not_exist()` method on a `Transaction`. So, for now, I use `CheckNotExist` mutation to handle "delete-your-writes" cases as [it was introduced](https://github.com/pingcap/tidb/pull/14968).
Do you have any idea about a possible use case of `check_not_exist` method?

ekexium: I'm not sure. @andylokandy Have you seen such requirements? IIRC there are people who want this.

andylokandy: It's actually an optimistic version of key_exists() right? I think it's rarely that such optimization is need to expose to api level.

src/transaction/buffer.rs
```diff
@@ -44,6 +44,23 @@ impl Buffer {
...
+    /// If the returned value is None, it means the key doesn't exist in buffer yet.
+    pub async fn get(&self, key: &Key) -> Option<Value> {
+        match self
+            .mutations
+            .lock()
+            .await
+            .entry_map
+            .get(key)
+            .map(BufferEntry::get_value)
+            .unwrap_or(MutationValue::Undetermined)
```
nrc, line 57: can use `get_from_mutations` here

nrc (commented):
One small comment, otherwise lgtm

src/transaction/buffer.rs
```diff
@@ -192,9 +209,24 @@ impl Buffer {
...
     pub async fn delete(&self, key: Key) {
-        self.mutations.lock().await.insert(key, BufferEntry::Del);
+        let mut mutations = self.mutations.lock().await;
+        let value = mutations.entry_map.entry(key).or_insert(BufferEntry::Del);
+
+        if let BufferEntry::Insert(_) = value {
+            *value = BufferEntry::CheckNotExist
+        } else {
+            *value = BufferEntry::Del
+        }
```
ekexium, line 229: `primary_key` is set in the `InnerBuffer::insert` method. We'd better use the `insert` method here

src/transaction/buffer.rs
ziyi-yan: Done

src/transaction/buffer.rs
ziyi-yan: Done

nrc (approved):
lgtm, thanks for the PR!

ti-srebot (approved):

PR #236, opened 2021-02-20
Automatic heartbeat

### LINKED ISSUE #189: Support large transaction

### LINKED ISSUE #227: Automatically send heartbeats
For long-running transactions, the client should send heartbeat messages to the server. This should be optional (similar to retrying) and on by default. It should still be possible to manually send heartbeats.

TszKitLo40:
Closes #227
Closes #189

PS: This PR is submitted as my code task for applying LFX mentorship 2021 Spring.


src/request/plan.rs
```diff
@@ -234,3 +231,25 @@ impl<P: Plan<Result: HasLocks>, PdC: PdClient> Plan for ResolveLockPlan<P, PdC>
...
+    pub status: Arc<RwLock<TransactionStatus>>,
+}
+
+#[async_trait]
+impl<Req: TxnHeartBeatRequest, P: Plan> Plan for HeartbeatPlan<Req, P> {
+    type Result = Req::Response;
+
+    async fn execute(&self) -> Result<Self::Result> {
+        loop {
+            tokio::sleep(tokio::time::Duration::from_millis(5000));
```
ekexium, line 247: We should use a constant (or configurable interval) here instead of an integer literal. I suggest 10 seconds (TiDB's default).

ekexium (commented):
Thanks for the PR! First eliminate build errors please.

src/request/plan.rs
TszKitLo40: I have made it as a constant which is 10 seconds.

ekexium: I mean a `const` :)

src/transaction/transaction.rs
```diff
@@ -691,41 +694,56 @@ impl Transaction {
...
+                        primary_key.clone(),
+                        DEFAULT_LOCK_TTL,
+                    );
+                    let plan = PlanBuilder::new(rpc.clone(), request)
                         .single_region()
-                        .await?
-                        .resolve_lock(self.options.retry_options.lock_backoff.clone())
-                        .retry_region(self.options.retry_options.region_backoff.clone())
+                        .await
+                        .expect("gg")
```
andylokandy, line 731: 23333333

`error!()` is preferred.

rustfmt.toml
```diff
@@ -1,2 +1,3 @@
 imports_granularity="Crate"
 format_code_in_doc_comments = true
+edition = "2018"
```
ekexium, line 3: Why do we need it?

Cargo.toml
```diff
@@ -57,3 +58,8 @@ members = [
     "tikv-client-store",
     "mock-tikv"
 ]
+
+[[test]]
+name = "failpoint_tests"
+path = "tests/failpoints.rs"
```
ekexium, line 64:
```suggestion
path = "tests/failpoints_tests.rs"
```
And I prefer `failpoint_tests` :)

src/transaction/transaction.rs
```diff
@@ -43,12 +45,13 @@ use tikv_client_proto::{kvrpcpb, pdpb::Timestamp};
...
 /// ```
 pub struct Transaction {
-    status: TransactionStatus,
+    status: Arc<RwLock<TransactionStatus>>,
     timestamp: Timestamp,
     buffer: Buffer,
     bg_worker: ThreadPool,
     rpc: Arc<PdRpcClient>,
     options: TransactionOptions,
+    spawn_heartbeat: bool,
```
ekexium, line 54: I think there exists a less ambiguous name. Like `started_heartbeat` or something else

src/transaction/transaction.rs
```diff
@@ -499,16 +503,28 @@ impl Transaction {
...
         let primary_key = self.buffer.get_primary_key().await;
         let mutations = self.buffer.to_proto_mutations().await;
+        if self.options.kind == TransactionKind::Optimistic
+            && self.options.heartbeat
+            && self.spawn_heartbeat
+        {
+            self.spawn_heartbeat = false;
+            if !mutations.is_empty() {
+                self.spawn_heartbeat().await;
+            }
```
ekexium, line 526: There is a same check in `commit()`. Maybe we can move this check out, and early returns if there is no need to commit/send heartbeat.

src/transaction/transaction.rs
```diff
@@ -670,11 +700,62 @@ impl Transaction {
     fn is_pessimistic(&self) -> bool {
         matches!(self.options.kind, TransactionKind::Pessimistic(_))
     }
+
+    async fn spawn_heartbeat(&mut self) {
+        let c_status = self.status.clone();
+        let primary_key = match self.buffer.get_primary_key().await {
```
ekexium, line 706:
```suggestion
        let primary_key = self.buffer.get_primary_key().await.expect("No primary key is impossible");
```

src/transaction/transaction.rs
```diff
@@ -670,11 +700,62 @@ impl Transaction {
...
+            Some(k) => k,
+            None => panic!("should have primary key"),
+        };
+        let start_ts = self.timestamp.clone();
+        let resolve_lock = self.options.retry_options.lock_backoff.clone();
+        let retry_region = self.options.retry_options.region_backoff.clone();
```
ekexium, line 712: I suggest renaming to `lock_backoff` and `region_backoff`

src/transaction/transaction.rs
```diff
@@ -670,11 +700,62 @@ impl Transaction {
     fn is_pessimistic(&self) -> bool {
         matches!(self.options.kind, TransactionKind::Pessimistic(_))
     }
+
+    async fn spawn_heartbeat(&mut self) {
+        let c_status = self.status.clone();
```
ekexium, line 705: Had better use a meaningful name

src/transaction/transaction.rs
```diff
@@ -670,11 +700,62 @@ impl Transaction {
...
+        let rpc = self.rpc.clone();
+
+        let task = async move {
+            loop {
+                tokio::time::sleep(DEFAULT_HEARTBEAT_INTERVAL).await;
+                {
+                    let status = c_status.read().await;
+                    if matches!(
+                        *status,
+                        TransactionStatus::Rolledback | TransactionStatus::Committed
```
ekexium, line 722: When a txn gets dropped, it will never become `Rolledback` or `Committed`. We should stop heartbeat when it happens.

src/transaction/transaction.rs
```diff
@@ -670,11 +700,62 @@ impl Transaction {
...
+                    start_ts.clone(),
+                    primary_key.clone(),
+                    (current_ts.physical - start_ts.physical) as u64 + DEFAULT_LOCK_TTL,
+                );
+                let plan = PlanBuilder::new(rpc.clone(), request)
+                    .single_region()
+                    .await?
+                    .resolve_lock(resolve_lock.clone())
+                    .retry_region(retry_region.clone())
+                    .post_process_default()
```
ekexium, line 738: I think `resolve_lock` and `post_process_default` is not necessary here.

src/transaction/transaction.rs
```diff
@@ -670,11 +700,62 @@ impl Transaction {
...
+                    .plan();
+                plan.execute().await?;
+            }
+            Ok::<(), Error>(())
+        };
+
+        // self.bg_worker.spawn_ok(async {
+        tokio::spawn(async {
+            match task.await {
```
ekexium, line 747:
```suggestion
            if let Err(err) =  task.await {
```

src/transaction/transaction.rs
```diff
@@ -804,10 +889,17 @@ impl TransactionOptions {
...
+    pub fn no_heart_beat(mut self) -> TransactionOptions {
+        self.heartbeat = false;
+        self
+    }
 }
 
 /// The default TTL of a lock in milliseconds
 const DEFAULT_LOCK_TTL: u64 = 3000;
+/// The default heartbeat interval in milliseconds
+const DEFAULT_HEARTBEAT_INTERVAL: Duration = tokio::time::Duration::from_millis(10000);
```
ekexium, line 902:
```suggestion
const DEFAULT_HEARTBEAT_INTERVAL: Duration = tokio::time::Duration::from_secs(1);
```
Considering `DEFAULT_LOCK_TTL` is 3s, and we don't use/calculate other lock TTLs like in TiDB, I think we can set the interval to 1 second or so.

tests/failpoints_tests.rs
```diff
@@ -0,0 +1,60 @@
+#![cfg(feature = "integration-tests")]
+
+mod common;
+use common::{clear_tikv, pd_addrs};
+use serial_test::serial;
+use tikv_client::{Result, TransactionClient, TransactionOptions};
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 6)]
+#[serial]
+async fn optimistic_heartbeat() -> Result<()> {
```
ekexium, line 10: Maybe add a few comments to explain the tests?

ekexium (commented):
I have a few questions and some suggestions on the style. Others LGTM.
@nrc @andylokandy

rustfmt.toml
andylokandy: This is for running `rustfmt` without `cargo`

src/transaction/transaction.rs
andylokandy: what about `is_heartbeat_started`?

rustfmt.toml
TszKitLo40: Without this, we can use `rustfmt`.

src/transaction/transaction.rs
```diff
@@ -644,6 +669,10 @@ impl Transaction {
             .retry_region(self.options.retry_options.region_backoff.clone())
             .merge(Collect)
             .plan();
+        if self.spawn_heartbeat && self.options.heartbeat {
+            self.spawn_heartbeat = false;
```
andylokandy, line 673: why set to `false`?

src/transaction/transaction.rs
```diff
@@ -804,10 +889,17 @@ impl TransactionOptions {
             }
         }
     }
+
+    pub fn no_heart_beat(mut self) -> TransactionOptions {
```
andylokandy, line 893: What about this:

```suggestion
    pub fn stop_heart_beat(mut self) -> TransactionOptions {
```

andylokandy: TBO, I think it may be useless. The one who wants to control the heartbeat process will not enable the automatic heartbeat.

src/transaction/transaction.rs
TszKitLo40: what about `heartbeat_started`?

src/transaction/transaction.rs
TszKitLo40: What about `start_heartbeat`?

src/transaction/transaction.rs
andylokandy: It's a convention to give the bool variable a `is` prefix.

src/transaction/transaction.rs
TszKitLo40: It is `true`, it means it that we should spawn a heartbeat thread, otherwise a heartbeat thread has been spawned,

src/transaction/transaction.rs
TszKitLo40: For rust convention?

src/transaction/transaction.rs
TszKitLo40: This function is used to disable auto heartbeat when we start a txn.

src/transaction/transaction.rs
```diff
@@ -657,34 +689,89 @@ impl Transaction {
...
+            | TransactionStatus::Dropped => Err(Error::OperationAfterCommitError),
         }
     }
 
     fn is_pessimistic(&self) -> bool {
         matches!(self.options.kind, TransactionKind::Pessimistic(_))
     }
+
+    async fn start_auto_heartbeat(&mut self) {
+        let c_status = self.status.clone();
```
andylokandy, line 709:
```suggestion
        let status = self.status.clone();
```

src/transaction/transaction.rs
```diff
@@ -499,16 +503,31 @@ impl Transaction {
...
+        if mutations.is_empty() {
+            assert!(primary_key.is_none());
+            return Ok(None);
+        }
+
+        if self.options.kind == TransactionKind::Optimistic
+            && self.options.auto_heartbeat
+            && !self.is_heartbeat_started
+        {
+            self.is_heartbeat_started = true;
```
andylokandy, line 528: What about moving this one into `start_auto_heartbeat()`?

src/transaction/transaction.rs
```diff
@@ -804,10 +895,17 @@ impl TransactionOptions {
             }
         }
     }
+
+    pub fn stop_heart_beat(mut self) -> TransactionOptions {
```
andylokandy, line 899: Oh! It's the option.  I wonder it was the txn. I suggest `no_auto_heartbeat()` for consistency.

src/transaction/transaction.rs
```diff
@@ -644,6 +672,10 @@ impl Transaction {
             .retry_region(self.options.retry_options.region_backoff.clone())
             .merge(Collect)
             .plan();
+        if !self.is_heartbeat_started && self.options.auto_heartbeat {
+            self.is_heartbeat_started = true;
+            self.start_auto_heartbeat().await;
+        }
```
nrc, line 678: I think we should execute the lock plan before we start the heartbeat - we don't want the heartbeat to arrive before the write.

src/transaction/transaction.rs
```diff
@@ -657,34 +689,89 @@ impl Transaction {
...
 
 impl Drop for Transaction {
     fn drop(&mut self) {
-        if !std::thread::panicking() && self.status == TransactionStatus::Active {
-            match self.options.check_level {
-                CheckLevel::Panic => {
-                    panic!("Dropping an active transaction. Consider commit or rollback it.")
+        {
+            let status = futures::executor::block_on(self.status.read());
+            if !std::thread::panicking() && *status == TransactionStatus::Active {
```
nrc, line 761: I think you should check if the thread is panicking first and early return in that case.

tests/failpoints_tests.rs
nrc: I think this test would be better off as a unit test - you could use a mock to check the heartbeat message is sent without having to use failpoints and rely on timing.

TszKitLo40:
Do you think `pessimistic_heartbeat` should be also as a unit test? @nrc

nrc:
> Do you think pessimistic_heartbeat should be also as a unit test? @nrc

It could be, but it is also nice to have a system test

TszKitLo40:
> > Do you think pessimistic_heartbeat should be also as a unit test? @nrc
> 
> It could be, but it is also nice to have a system test

So do you think these two tests should also have a unit test and integration test?

TszKitLo40:
For a unit test, I think we should start a txn and commit it. But how to use a mock to check the heartbeat message is sent without having to use failpoints and rely on timing? @nrc

src/transaction/transaction.rs
```diff
@@ -499,16 +503,30 @@ impl Transaction {
...
         let primary_key = self.buffer.get_primary_key().await;
         let mutations = self.buffer.to_proto_mutations().await;
+        if mutations.is_empty() {
+            assert!(primary_key.is_none());
+            return Ok(None);
+        }
+
+        if self.options.kind == TransactionKind::Optimistic
+            && self.options.auto_heartbeat
+            && !self.is_heartbeat_started
```
ekexium, line 526: The checks can all be moved to `start_auto_heartbeat`. Then maybe rename `start_auto_heartbeat` to reflect the change.

TszKitLo40: What about `may_start_auto_heartbeat`?

andylokandy: I suggest `ensure_start_auto_heartbeat`

TszKitLo40: Do you mean `ensure_start_auto_heartbeat` including the `heartbeat` logic or just the check?

andylokandy: hmmmm, after deeper thought, I think we may keep it `start_auto_heartbeat`

TszKitLo40: But after the checks have been moved to `start_auto_heartbeat`, `start_auto_heartbeat` may be called multiple times.

andylokandy: I think it should be fine.

TszKitLo40: I have moved these checks to `start_auto_heartbeat`, PTAL.

andylokandy:
The test failed:

```
thread 'backoff::test::test_equal_jitter_backoff_with_invalid_base_delay_ms' panicked at 'Both base_delay_ms and max_delay_ms must be greater than 1', src/backoff.rs:146:9
   0: std::panicking::begin_panic
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:519:12
   1: tikv_client::backoff::Backoff::decorrelated_jitter_backoff
             at ./src/backoff.rs:171:9
   2: tikv_client::backoff::test::test_decorrelated_jitter_backoff_with_invalid_base_delay_ms
             at ./src/backoff.rs:297:9
   3: tikv_client::backoff::test::test_decorrelated_jitter_backoff_with_invalid_base_delay_ms::{{closure}}
             at ./src/backoff.rs:296:5
   4: core::ops::function::FnOnce::call_once
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/core/src/ops/function.rs:227:5
   5: core::ops::function::FnOnce::call_once
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:519:12
   1: tikv_client::backoff::Backoff::equal_jitter_backoff
             at ./src/backoff.rs:146:9
   2: tikv_client::backoff::test::test_equal_jitter_backoff_with_invalid_base_delay_ms
             at ./src/backoff.rs:264:9
   3: tikv_client::backoff::test::test_equal_jitter_backoff_with_invalid_base_delay_ms::{{closure}}
             at ./src/backoff.rs:263:5
   4: core::ops::function::FnOnce::call_once
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/core/src/ops/function.rs:227:5
   5: core::ops::function::FnOnce::call_once
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/core/src/ops/function.rs:227:5
```

andylokandy:
I've restarted the test. Let's see what will happen.

nrc:
That test is super flakey (#240)

longfangsong:
> test_equal_jitter_backoff_with_invalid_base_delay_ms

I think that one is supposed to panic:

```rust
#[test]
#[should_panic(expected = "Both base_delay_ms and max_delay_ms must be greater than 1")]
fn test_equal_jitter_backoff_with_invalid_base_delay_ms() {
    Backoff::equal_jitter_backoff(1, 7, 3);
}
```

The real test caused the CI failure is `failpoints_tests::optimistic_heartbeat`

ekexium:
> > test_equal_jitter_backoff_with_invalid_base_delay_ms
> 
> I think that one is supposed to panic:
> 
> ```rust
> #[test]
> #[should_panic(expected = "Both base_delay_ms and max_delay_ms must be greater than 1")]
> fn test_equal_jitter_backoff_with_invalid_base_delay_ms() {
>     Backoff::equal_jitter_backoff(1, 7, 3);
> }
> ```
> 
> The real test caused the CI failure is `failpoints_tests::optimistic_heartbeat`

The failure of `optimistic_heartbeat` is caused by a known bug, which will be fixed in #245

tests/failpoints_tests.rs
```diff
@@ -0,0 +1,64 @@
+#![cfg(feature = "integration-tests")]
```
andylokandy, line 1: Why is it seperated from `tests/integration_tests.rs`?

src/transaction/transaction.rs
```diff
@@ -1031,4 +1130,100 @@ enum TransactionStatus {
...
+            atomic::{AtomicUsize, Ordering},
+            Arc,
+        },
+    };
+    use tikv_client_proto::{kvrpcpb, pdpb::Timestamp};
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 6)]
+    async fn test_optimistic_heartbeat() -> Result<(), io::Error> {
+        let scenario = FailScenario::setup();
+        fail::cfg("after-prewrite", "return(true)").unwrap();
```
andylokandy, line 1158: why not use `fail::cfg("after-prewrite", "sleep(10000)").unwrap();`?

src/transaction/transaction.rs
```diff
@@ -1031,4 +1130,100 @@ enum TransactionStatus {
...
+            Timestamp::default(),
+            bg_worker,
+            pd_client,
+            TransactionOptions::new_optimistic(),
+        );
+        heartbeat_txn.put(key1.clone(), "foo").await.unwrap();
+        let heartbeat_txn_handle = tokio::spawn(async move {
+            assert!(heartbeat_txn.commit().await.is_ok());
+        });
+        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
```
andylokandy, line 1184: Had better check `heartbeats` == 0 before sleeping.

src/transaction/transaction.rs
TszKitLo40: It seems that `sleep(10000)` will block the threads so that the test will fail.

tests/failpoints_tests.rs
TszKitLo40: This test have failpoints.

andylokandy: I suggest to put them together

src/transaction/transaction.rs
andylokandy: So why it sleeps here:
https://github.com/tikv/client-rust/blob/b635096bed1eba224817ac827a056f18fad17898/tests/failpoints_tests.rs#L14

tests/failpoints_tests.rs
ekexium: It follows the suggestions given by `fail-rs`.

> Carefully consider complex, concurrent, non-deterministic combinations of fail points. Put test cases exercising fail points into their own test crate.

src/transaction/transaction.rs
```diff
@@ -856,13 +944,11 @@ impl Committer {
...
             }
         };
-        self.bg_worker
-            .clone()
-            .spawn_ok(self.commit_secondary(commit_ts.clone()).map(|res| {
-                if let Err(e) = res {
-                    warn!("Failed to commit secondary keys: {}", e);
-                }
-            }));
+        tokio::spawn(self.commit_secondary(commit_ts.clone()).map(|res| {
```
ekexium, line 947: @andylokandy Then we will require the tokio runtime? Others LGTM.

andylokandy: Yes

TszKitLo40:
It seems that you need to fix the DCO error. @andylokandy

src/transaction/transaction.rs
ekexium: timer and rwlock can be replaced with runtime independent ones.

ti-srebot (approved):

Issue #242, opened 2021-02-25
Some errors are not propagated out


ekexium:
https://github.com/tikv/client-rust/blob/3fad149ff31448288093a54288dd0e7688c3ad89/tikv-client-store/src/errors.rs#L153-L157

The implementation ignores `Err`, which will cause problems, e.g. we turn `error` to `Err` in `MultiRegion` and following plans will not be able to detect the error using `HasError::error`.

Even if the errors get propagated, we sometimes ignore them. For example, if the response contains errors, following commands can still succeed.
https://github.com/tikv/client-rust/blob/3fad149ff31448288093a54288dd0e7688c3ad89/src/transaction/transaction.rs#L943-L952
and 
https://github.com/tikv/client-rust/blob/3fad149ff31448288093a54288dd0e7688c3ad89/src/transaction/transaction.rs#L1008

nrc:
cc #238

Issue #238, opened 2021-02-22
0.2: Improve error handling


nrc:
Debugging and handling errors is sub-optimal. Our errors are pretty good for the general case, and for just printing an error to a log or whatever, but if you're actually trying to fix bugs or handle the errors programatically, then they are not ideal.

A concrete example is that if an error is thrown while a transaction has not been committed/dropped, then we panic and the underlying error is lost. I think there are other places where an error causes another error and we lose underlying info.

PR #245, opened 2021-03-02
Fix HasError and HasLocks

ekexium:
Related issues: #242 #238 

The PR tries to fix the following issues:
1. `HasError for Result` ignores `Err`
2. `HasLocks` ignores the `error` field
3. If a request returns `Ok(response)` while the response contains a non-empty `error` field, it's often ignored.

### DIFF
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -261,6 +261,11 @@ impl<P: Plan> Clone for ExtractError<P> {
     }
 }
 
+/// When executed, the plan extracts errors from its inner plan, and
+/// returns an `Err` wrapping the error.
+///
+/// The errors come from two places: `Err` from inner plans, and `Ok(response)`
+/// where `response` contains unresolved errors.
 #[async_trait]
 impl<P: Plan> Plan for ExtractError<P>
 where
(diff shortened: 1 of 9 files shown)


src/transaction/requests.rs
```diff
@@ -547,13 +563,82 @@ pub struct SecondaryLocksStatus {
     pub commit_ts: Option<Timestamp>,
 }
 
-impl HasLocks for kvrpcpb::CommitResponse {}
+impl HasLocks for kvrpcpb::CommitResponse {
```
nrc, line 566: Could you use a macro for these impls, it's a lot of duplicated code.

src/request/plan.rs
```diff
@@ -249,6 +249,35 @@ where
     }
 }
 
+pub struct PropagateError<P: Plan> {
```
nrc, line 252: Could you document the plan please? I also think the name is not quite right, maybe `ExtractError` or something is better?

nrc (commented):
Could you add some tests please? Otherwise its looking pretty good.

src/request/plan.rs
ekexium: My first thought was `ExtractError` too, but there exists a similar name `extract_errors`.

src/transaction/requests.rs
ekexium: Sure

src/request/plan.rs
nrc: It's in a pretty different part of the system, so I think it should be ok

ekexium:
Comments addressed, PTAL @nrc

nrc (approved):
lgtm!

PR #247, opened 2021-03-07
Upgrade a bunch of deps

nrc:
This gets us on the latest versions for everything, removes some duplicate deps, and means we pass `cargo audit`.

PTAL @ekexium


ekexium:
LGTM

Issue #249, opened 2021-03-10
Calculate proper TTL with transaction size


andylokandy:
Current implementation sets the TTL as a constant value, and we'd like to calculate the TTL based on the transaction size so as to reduce heartbeat requests.

PR #250, opened 2021-03-11
Calculate cleverer TTL for transaction

TszKitLo40:
Closes [#249](https://github.com/tikv/client-rust/issues/249)

### DIFF
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -227,12 +227,10 @@ impl Buffer {
     pub fn get_write_size(&self) -> usize {
         self.entry_map
             .iter()
-            .map(|(k, v)| {
-                match v {
-                    BufferEntry::Put(val) | BufferEntry::Insert(val) => val.len() + k.len(),
-                    BufferEntry::Del => k.len(),
-                    _ => 0,
-                }
+            .map(|(k, v)| match v {
+                BufferEntry::Put(val) | BufferEntry::Insert(val) => val.len() + k.len(),
+                BufferEntry::Del => k.len(),
+                _ => 0,
             })
             .sum()
     }
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -229,9 +229,7 @@ impl Buffer {
             .iter()
             .map(|(k, v)| {
                 match v {
-                    BufferEntry::Put(val) | BufferEntry::Insert(val) => {
-                        val.len() + k.len()
-                    }
+                    BufferEntry::Put(val) | BufferEntry::Insert(val) => val.len() + k.len(),
                     BufferEntry::Del => k.len(),
                     _ => 0,
                 }
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -224,6 +224,23 @@ impl Buffer {
             .collect()
     }
 
+    pub fn get_write_size(&self) -> usize {
+        self.entry_map
+            .iter()
+            .filter_map(|(k, v)| {
+                let mut write_size = 0;
+                if let BufferEntry::Put(val) | BufferEntry::Insert(val) = v {
+                    write_size += val.len();
+                    write_size += k.len();
+                }
+                if let BufferEntry::Del = v {
+                    write_size += k.len();
+                }
+                Some(write_size)
+            })
+            .sum()
+    }
+
     fn get_from_mutations(&self, key: &Key) -> MutationValue {
         self.entry_map
             .get(&key)
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -281,13 +281,7 @@ impl Buffer {
         mutations
             .entry_map
             .iter()
-            .filter(|(_, v)| {
-                matches!(
-                    v,
-                    BufferEntry::Put(_) | BufferEntry::Insert(_) | BufferEntry::Del
-                )
-            })
-            .map(|(k, v)| {
+            .filter_map(|(k, v)| {
                 let mut write_size = 0;
                 if let BufferEntry::Put(val) | BufferEntry::Insert(val) = v {
                     write_size += val.len();
@@ -296,7 +290,7 @@ impl Buffer {
                 if let BufferEntry::Del = v {
                     write_size += k.len();
                 }
-                write_size
+                Some(write_size)
             })
             .sum()
     }
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -229,12 +229,13 @@ impl Buffer {
             .iter()
             .map(|(k, v)| {
                 let mut write_size = 0;
-                if let BufferEntry::Put(val) | BufferEntry::Insert(val) = v {
-                    write_size += val.len();
-                    write_size += k.len();
-                }
-                if let BufferEntry::Del = v {
-                    write_size += k.len();
+                match v {
+                    BufferEntry::Put(val) | BufferEntry::Insert(val) => {
+                        write_size += val.len();
+                        write_size += k.len();
+                    }
+                    BufferEntry::Del => write_size += k.len(),
+                    _ => (),
                 }
                 write_size
             })
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -275,6 +275,31 @@ impl Buffer {
             }
         }
     }
+
+    pub async fn get_write_size(&self) -> usize {
+        let mutations = self.mutations.lock().await;
+        mutations
+            .entry_map
+            .iter()
+            .filter(|(_, v)| {
+                matches!(
+                    v,
+                    BufferEntry::Put(_) | BufferEntry::Insert(_) | BufferEntry::Del
+                )
+            })
+            .map(|(k, v)| {
+                let mut write_size = 0;
+                if let BufferEntry::Put(val) | BufferEntry::Insert(val) = v {
+                    write_size += val.len();
+                    write_size += k.len();
+                }
+                if let BufferEntry::Del = v {
+                    write_size += k.len();
+                }
+                write_size
+            })
+            .sum()
+    }
 }
 
 // The state of a key-value pair in the buffer.
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -228,16 +228,13 @@ impl Buffer {
         self.entry_map
             .iter()
             .map(|(k, v)| {
-                let mut write_size = 0;
                 match v {
                     BufferEntry::Put(val) | BufferEntry::Insert(val) => {
-                        write_size += val.len();
-                        write_size += k.len();
+                        val.len() + k.len()
                     }
-                    BufferEntry::Del => write_size += k.len(),
-                    _ => (),
+                    BufferEntry::Del => k.len(),
+                    _ => 0,
                 }
-                write_size
             })
             .sum()
     }
--- a/src/transaction/buffer.rs
+++ b/src/transaction/buffer.rs
@@ -227,7 +227,7 @@ impl Buffer {
     pub fn get_write_size(&self) -> usize {
         self.entry_map
             .iter()
-            .filter_map(|(k, v)| {
+            .map(|(k, v)| {
                 let mut write_size = 0;
                 if let BufferEntry::Put(val) | BufferEntry::Insert(val) = v {
                     write_size += val.len();
@@ -236,7 +236,7 @@ impl Buffer {
                 if let BufferEntry::Del = v {
                     write_size += k.len();
                 }
-                Some(write_size)
+                write_size
             })
             .sum()
     }
--- a/src/transaction/transaction.rs
+++ b/src/transaction/transaction.rs
@@ -895,10 +895,17 @@ impl TransactionOptions {
     }
 }
 
+/// The default max TTL of a lock in milliseconds
+const MANAGED_TTL: u64 = 20000;
 /// The default TTL of a lock in milliseconds.
 const DEFAULT_LOCK_TTL: u64 = 3000;
-/// The default heartbeat interval.
-const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(1);
+/// The default heartbeat interval
+const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_millis(MANAGED_TTL / 2);
+/// TiKV recommends each RPC packet should be less than ~1MB. We keep each packet's
+/// Key+Value size below 16KB.
+const TXN_COMMIT_BATCH_SIZE: u64 = 16 * 1024;
+const BYTES_PER_MB: u64 = 1024 * 1024;
+const TTL_FACTOR: f64 = 6000_f64;
 
 /// A struct wrapping the details of two-phase commit protocol (2PC).
 ///
--- a/src/transaction/transaction.rs
+++ b/src/transaction/transaction.rs
@@ -525,6 +525,7 @@ impl<PdC: PdClient> Transaction<PdC> {
             self.timestamp.clone(),
             self.rpc.clone(),
             self.options.clone(),
+            self.buffer.get_write_size().await as u64,
         )
         .commit()
         .await;
@@ -563,6 +564,7 @@ impl<PdC: PdClient> Transaction<PdC> {
             self.timestamp.clone(),
             self.rpc.clone(),
             self.options.clone(),
+            self.buffer.get_write_size().await as u64,
         )
         .rollback()
         .await;
@@ -923,6 +925,7 @@ struct Committer<PdC: PdClient = PdRpcClient> {
     options: TransactionOptions,
     #[new(default)]
     undetermined: bool,
+    txn_size: u64,
 }
 
 impl<PdC: PdClient> Committer<PdC> {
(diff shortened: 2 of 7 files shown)


src/transaction/buffer.rs
```diff
@@ -275,6 +275,20 @@ impl Buffer {
             }
         }
     }
+
+    pub async fn get_size(&self) -> u64 {
```
andylokandy, line 279: @ekexium Is it closer to convention?

```suggestion
    pub async fn get_write_size(&self) -> u64 {
```

src/transaction/transaction.rs
```diff
@@ -1101,6 +1110,21 @@ impl<PdC: PdClient> Committer<PdC> {
         }
         Ok(())
     }
+
+    fn txn_lock_ttl(&mut self) -> u64 {
```
andylokandy, line 1114:
```suggestion
    fn calc_txn_lock_ttl(&mut self) -> u64 {
```

src/transaction/transaction.rs
```diff
@@ -1101,6 +1110,21 @@ impl<PdC: PdClient> Committer<PdC> {
...
+        let mut lock_ttl = DEFAULT_LOCK_TTL;
+        if self.txn_size > TXN_COMMIT_BATCH_SIZE {
+            let size_mb = self.txn_size / BYTES_PER_MB;
+            lock_ttl = ((TTL_FACTOR as f64) * ((size_mb) as f64).sqrt()) as u64;
+            if lock_ttl < DEFAULT_LOCK_TTL {
+                lock_ttl = DEFAULT_LOCK_TTL;
+            }
+            if lock_ttl > MANAGED_TTL {
+                lock_ttl = MANAGED_TTL;
+            }
```
andylokandy, line 1124:
```suggestion
            let lock_ttl = lock_ttl.min(MANAGED_TTL).max(DEFAULT_LOCK_TTL);
```

src/transaction/buffer.rs
ekexium: I prefer `get_write_size`. It's more accurate

TszKitLo40: I have changed it to `get_write_size`

src/transaction/transaction.rs
```diff
@@ -895,10 +901,16 @@ impl TransactionOptions {
...
 }
 
-/// The default TTL of a lock in milliseconds.
+const MANAGED_TTL: u64 = 20000;
+/// The default TTL of a lock in milliseconds
 const DEFAULT_LOCK_TTL: u64 = 3000;
-/// The default heartbeat interval.
-const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(1);
+/// The default heartbeat interval in milliseconds
+const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(MANAGED_TTL / 2);
```
ekexium, line 908:
```suggestion
const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_millis(MANAGED_TTL / 2);
```

ekexium (commented):
I think the new failpoint is unnecessary? Though it makes the test take a long time. 
We can make the TTL configurable, maybe.

src/transaction/transaction.rs
```diff
@@ -710,7 +712,11 @@ impl<PdC: PdClient> Transaction<PdC> {
 
         let heartbeat_task = async move {
             loop {
-                tokio::time::sleep(DEFAULT_HEARTBEAT_INTERVAL).await;
+                let heartbeat_interval = (|| {
+                    fail_point!("reset-heartbeat-interval", |_| Duration::from_millis(1000));
```
nrc, line 716: Why have this fail point?

src/transaction/transaction.rs
```diff
@@ -895,10 +901,16 @@ impl TransactionOptions {
...
-/// The default TTL of a lock in milliseconds.
+const MANAGED_TTL: u64 = 20000;
+/// The default TTL of a lock in milliseconds
 const DEFAULT_LOCK_TTL: u64 = 3000;
-/// The default heartbeat interval.
-const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(1);
+/// The default heartbeat interval in milliseconds
+const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(MANAGED_TTL / 2);
+/// TiKV recommends each RPC packet should be less than ~1MB. We keep each packet's
+/// Key+Value size below 16KB.
```
nrc, line 910: I don't understand this comment, why does packet <1mb => K+V <16KB ?

src/transaction/transaction.rs
```diff
@@ -895,10 +901,16 @@ impl TransactionOptions {
...
+const TXN_COMMIT_BATCH_SIZE: u64 = 16 * 1024;
+const BYTES_PER_MB: u64 = 1024 * 1024;
+const TTL_FACTOR: u64 = 6000;
```
nrc, line 913:
```suggestion
const TTL_FACTOR: f64 = 6000;
```

src/transaction/transaction.rs
```diff
@@ -1101,6 +1113,16 @@ impl<PdC: PdClient> Committer<PdC> {
         }
         Ok(())
     }
+
+    fn cal_txn_lock_ttl(&mut self) -> u64 {
```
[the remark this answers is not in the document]
nrc: Could you document this function please. In particular why use the calculation which is used?

src/transaction/buffer.rs
```diff
@@ -275,6 +275,20 @@ impl Buffer {
             }
         }
     }
+
+    pub async fn get_write_size(&self) -> u64 {
+        let mutations = self.mutations.lock().await;
+        let mut size = 0_u64;
```
nrc, line 281:
```suggestion
        let mut size = 0;
```

And remove the casts below

src/transaction/buffer.rs
```diff
@@ -275,6 +275,20 @@ impl Buffer {
             }
         }
     }
+
+    pub async fn get_write_size(&self) -> u64 {
```
nrc, line 279:
```suggestion
    pub async fn get_write_size(&self) -> usize {
```

src/transaction/buffer.rs
```diff
@@ -275,6 +275,20 @@ impl Buffer {
...
+        let mutations = self.mutations.lock().await;
+        let mut size = 0_u64;
+        for (k, v) in &mutations.entry_map {
```
nrc, line 282:
```suggestion
        mutations.entry_map.iter().sum(...)
```

src/transaction/buffer.rs
```diff
@@ -275,6 +275,20 @@ impl Buffer {
...
+            match v {
+                BufferEntry::Put(val) => size += val.len() as u64,
+                BufferEntry::Insert(val) => size += val.len() as u64,
```
nrc, line 285:
```suggestion
                BufferEntry::Put(val) | BufferEntry::Insert(val) => size += val.len(),
```

nrc: Also, could use `if let` instead of `match`

nrc (commented):
Could you rebase to tidy up the commits please?

TszKitLo40:
> Could you rebase to tidy up the commits please?

src/transaction/transaction.rs
TszKitLo40: This comment is copied from the tikv client of TiDB.

src/transaction/transaction.rs
TszKitLo40: Because the default heart-interval is 10s and the test will failed if the ttl is 3s. So maybe we need to have a fail point here to set the heartbeat to 1s or make it configurable.

src/transaction/buffer.rs
TszKitLo40: The function `calc_txn_lock_ttl` is the same as TiDB. But there is no document in TiDB too.

nrc:
> I have tried to rebase. But there will be many conflicts. So I give it up. @nrc

I think you can squash the review changes commits and merge commits into the last meaningful commit, then start a new branch and cherry-pick the commits which are actually in this PR, then you'll have a clean commit history and you can force push to this remote branch

src/transaction/buffer.rs
```diff
@@ -275,6 +275,31 @@ impl Buffer {
...
+        mutations
+            .entry_map
+            .iter()
+            .filter(|(_, v)| {
+                matches!(
+                    v,
+                    BufferEntry::Put(_) | BufferEntry::Insert(_) | BufferEntry::Del
+                )
+            })
+            .map(|(k, v)| {
```
nrc, line 290: You can use `filter_map` instead of `filter(...).map(...)`

src/transaction/transaction.rs
nrc: Could you just link to that comment rather than copying it please?

src/transaction/transaction.rs
```diff
@@ -895,10 +901,16 @@ impl TransactionOptions {
     }
 }
 
-/// The default TTL of a lock in milliseconds.
+const MANAGED_TTL: u64 = 20000;
```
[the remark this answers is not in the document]
nrc: What is "managed" about the TTL? I think either rename the const or explain that in the comment.

src/transaction/transaction.rs
nrc: Yeah, lets make it configurable, it seems like a weird use of a failpoint to me

src/transaction/buffer.rs
TszKitLo40: If we make the heartbeat interval configurable, do you think we should make it as an option in transaction option?

nrc: Yes, that sounds right to me

TszKitLo40: @ekexium @andylokandy what you think of this?

andylokandy:
@nrc Just for curious, in what case configurable heartbeat interval is needed?

nrc:
> @nrc Just for curious, in what case configurable heartbeat interval is needed?

I don't know. There is at least one case which is making the test work :-) In real life, maybe we want to do this for chaos=style testing of TiKV, or maybe a user has a particularly unstable network and wants more frequent heartbeats?

andylokandy:
For testing we have many other solutions, for example, failpoint or `cfg[test]`. In fact, I think the ttl can't be properly configurable because it's not a fixed value.

nrc:
>  I don't think the ttl can be properly configurable because it's not a fixed value.

I could imagine something like:

```
enum TtlConfig {
    FixedTime(u64),
    Default,
    ....
```

where `Default` uses the default algortithm and `FixedTime` sets it to a regular heartbeat. We could also specify the parameters to the TTL algorithm, etc

src/transaction/transaction.rs
```diff
@@ -916,6 +925,7 @@ struct Committer<PdC: PdClient = PdRpcClient> {
     options: TransactionOptions,
     #[new(default)]
     undetermined: bool,
+    txn_size: u64,
```
andylokandy, line 928:
```suggestion
    write_size: u64,
```

src/transaction/transaction.rs
```diff
@@ -955,19 +965,20 @@ impl<PdC: PdClient> Committer<PdC> {
     async fn prewrite(&mut self) -> Result<Option<Timestamp>> {
         let primary_lock = self.primary_key.clone().unwrap();
         // FIXME: calculate TTL for big transactions
-        let lock_ttl = DEFAULT_LOCK_TTL;
+        let current_ts = self.rpc.clone().get_timestamp().await?;
```
andylokandy, line 968: You should store the beginning time in the transaction rather than a new fetching timestamp in prewrite.

src/transaction/transaction.rs
```diff
@@ -895,10 +897,17 @@ impl TransactionOptions {
...
 /// The default TTL of a lock in milliseconds.
 const DEFAULT_LOCK_TTL: u64 = 3000;
-/// The default heartbeat interval.
-const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(1);
+/// The default heartbeat interval
+const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_millis(MANAGED_TTL / 2);
+/// TiKV recommends each RPC packet should be less than ~1MB. We keep each packet's
+/// Key+Value size below 16KB.
+const TXN_COMMIT_BATCH_SIZE: u64 = 16 * 1024;
+const BYTES_PER_MB: u64 = 1024 * 1024;
```
andylokandy, line 909: I think this is straightforward and no need for a constant.

andylokandy:
@nrc AFAICS, it'll be relatively difficult for the user the set the proper heartbeat interval that works fine with the calculated TTL for prewrite, which is dynamic and not configurable. And it's only useful in test. So I suggest making these changes to the `TransacitonOption`:

```rust
pub struct TransactionOptions {
    ...
    #[cfg(test)]
    heartbeat: HeartbeatOption,
}

#[cfg(test)]
pub enum HeartbeatOption {
    NoHeartbeat,
    FixedTime(Duration),
}
```

nrc:
@andylokandy  Do you mean to have a default or dynamic option in `HeartbeatOption`? 

I don't mind too much if it is cfg(test) or not. I guess if people want to use it for real, they can ask us to un-cfg it.

andylokandy:
@nrc The heartbeat timeout isn't dynamic. I believe it's hard-coded `10s` in TiDB. The dynamic one is the TTL calculated for prewrite, and it's not configurable.

src/transaction/transaction.rs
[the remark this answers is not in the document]
andylokandy: I prefer `MAX_TTL`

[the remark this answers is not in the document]
ekexium: I believe the name comes from the `ManagedLockTTL` in TiDB. But I don't know why it's called "managed" either.

andylokandy:
@TszKitLo40 I have changed `#[cfg(test)]` to `#[doc(hidden)]` to simplify the problem.

TszKitLo40:
> @TszKitLo40 I have changed `#[cfg(test)]` to `#[doc(hidden)]` to simplify the problem.

And I have fixed the tests

andylokandy:
@ekexium @nrc PTAL

src/transaction/transaction.rs
```diff
@@ -1251,35 +1290,33 @@ mod tests {
...
             pd_client,
-            TransactionOptions::new_pessimistic(),
+            TransactionOptions::new_pessimistic()
+                .heartbeat_option(HeartbeatOption::FixedTime(Duration::from_secs(1))),
         );
         heartbeat_txn.put(key1.clone(), "foo").await.unwrap();
         assert_eq!(heartbeats.load(Ordering::SeqCst), 0);
-        tokio::time::sleep(tokio::time::Duration::from_millis(5000)).await;
+        tokio::time::sleep(tokio::time::Duration::from_millis(1500)).await;
         let heartbeat_txn_handle = tokio::spawn(async move {
```
ekexium, line 1315: We can add an assertion immediately after the sleep.
```suggestion
        assert_eq!(heartbeats.load(Ordering::SeqCst), 1);
        let heartbeat_txn_handle = tokio::spawn(async move {
```

src/transaction/transaction.rs
```diff
@@ -1173,6 +1198,16 @@ impl<PdC: PdClient> Committer<PdC> {
         }
         Ok(())
     }
+
+    fn calc_txn_lock_ttl(&mut self) -> u64 {
+        let mut lock_ttl = DEFAULT_LOCK_TTL;
+        if self.txn_size > TXN_COMMIT_BATCH_SIZE {
+            let size_mb = self.txn_size / 1024 / 1024;
+            lock_ttl = ((TTL_FACTOR as f64) * ((size_mb) as f64).sqrt()) as u64;
+            lock_ttl = lock_ttl.min(MAX_TTL).max(DEFAULT_LOCK_TTL);
```
andylokandy, line 1207:
```suggestion
            let size_mb = self.txn_size as f64 / 1024.0 / 1024.0;
            lock_ttl = (TTL_FACTOR * size_m.sqrt()) as u64;
            lock_ttl = lock_ttl.min(MAX_TTL).max(DEFAULT_LOCK_TTL);
```

src/transaction/transaction.rs
andylokandy: @TszKitLo40 Is this patch working?

src/transaction/transaction.rs
andylokandy: @TszKitLo40 PTAL

src/transaction/transaction.rs
andylokandy: @TszKitLo40 PTAL

src/transaction/transaction.rs
```diff
@@ -1027,19 +1051,20 @@ impl<PdC: PdClient> Committer<PdC> {
     async fn prewrite(&mut self) -> Result<Option<Timestamp>> {
         let primary_lock = self.primary_key.clone().unwrap();
         // FIXME: calculate TTL for big transactions
```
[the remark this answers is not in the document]
andylokandy: @TszKitLo40 PTAL

src/transaction/transaction.rs
TszKitLo40: done, PTAL

src/transaction/transaction.rs
andylokandy: @TszKitLo40 PTAL

TszKitLo40: done

andylokandy (approved):

PR #252, opened 2021-03-15
Fix batch_get_for_update

ekexium:
The PR tries to fix the `batch_get_for_update` by introducing a new plan `PreserveKeys` and a merge strategy `MergeAndMatchKey`.

### DIFF
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -128,10 +128,6 @@ impl Key {
     }
 }
 
-pub trait HasKeys {
-    fn get_keys(&self) -> Vec<Key>;
-}
-
 impl From<Vec<u8>> for Key {
     fn from(v: Vec<u8>) -> Self {
         Key(v)
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -128,6 +128,10 @@ impl Key {
     }
 }
 
+pub trait HasKeys {
+    fn get_keys(&self) -> Vec<Key>;
+}
+
 impl From<Vec<u8>> for Key {
     fn from(v: Vec<u8>) -> Self {
         Key(v)
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -2,7 +2,6 @@
 
 use crate::{
     backoff::Backoff,
-    kv::HasKeys,
     pd::PdClient,
     request::{KvRequest, Shardable},
     stats::tikv_stats,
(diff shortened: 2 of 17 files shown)


src/request/plan_builder.rs
```diff
@@ -14,6 +15,8 @@ use crate::{
 use std::{marker::PhantomData, sync::Arc};
 use tikv_client_store::HasError;
 
+use super::plan::PreserveKey;
```
nrc, line 18: add this to the crate import

src/request/shard.rs
```diff
@@ -9,6 +10,8 @@ use crate::{
 use futures::stream::BoxStream;
 use std::sync::Arc;
 
+use super::plan::PreserveKey;
```
nrc, line 13: and here

src/request/shard.rs
```diff
@@ -51,6 +54,21 @@ impl<P: Plan + Shardable, PdC: PdClient> Shardable for ResolveLock<P, PdC> {
...
+        pd_client: &Arc<impl PdClient>,
+    ) -> BoxStream<'static, Result<(Self::Shard, Store)>> {
+        self.inner.shards(pd_client)
+    }
+
+    fn apply_shard(&mut self, shard: Self::Shard, store: &Store) -> Result<()> {
+        self.inner.apply_shard(shard, store)
+    }
+}
+
```
nrc, line 71: Could you use an impl to generate this impl and the RetryRegion one?

src/kv/key.rs
```diff
@@ -128,6 +128,10 @@ impl Key {
     }
 }
 
+pub trait HasKeys {
```
nrc, line 131: I think it is better for this trait to live in src/requests/plan.rs

src/request/plan.rs
```diff
@@ -292,6 +313,38 @@ where
...
+impl<P> Plan for PreserveKey<P>
+where
+    P: Plan + HasKeys,
+{
+    type Result = ResponseAndKeys<P::Result>;
+
+    async fn execute(&self) -> Result<Self::Result> {
+        let keys = self.inner.get_keys();
+        let res = self.inner.execute().await?;
+        // TODO: should we check they have the same length?
```
nrc, line 343: I don't think we can, since `res` may not be a list

src/request/plan.rs
```diff
@@ -349,3 +402,60 @@ mod test {
             .for_each(|r| assert!(r.is_err()));
     }
 }
+
```
nrc, line 405: These items should come before the tests in the file

src/request/plan.rs
```diff
@@ -349,3 +402,60 @@ mod test {
...
+// contains a response and the corresponding keys
+// currently only used for matching keys and values in pessimistic lock requests
+#[derive(Debug, Clone)]
+pub struct ResponseAndKeys<Resp>(Resp, Vec<Key>);
```
nrc, line 409: Could this just be a tuple?

nrc (commented):
Mostly looks good, a few comments but they're mostly about style.

src/request/plan.rs
ekexium: I doubt tuples can be ambiguous. It's not intuitive to tell what `(Response, Vec<Key>)` means, IMO.

src/request/shard.rs
nrc: whoops, _macro_ to generate the impl

src/request/plan.rs
nrc: IMO the reading of the tuple is exactly `Response` and `Key`s. To be a bit abstract, I believe that a tuple struct is equivalent to a newtype and similarly should be used if and only if the newtype introduces an invariant over the component types.

ekexium: Makes sense. 
I think one thing that introducing the new type benefits is that the traits are implemented only for this type but not for general `(Resnponse, Vec<Key>)`. So it won't interfere with other uses of `(Resnponse, Vec<Key>)` if we use it for other purposes, or even where we may need different implementations of these traits (though not very likely).

nrc: Are there other uses of (Response, Vec<Key>) where this impl does not make sense?

ekexium: I can't think of one...

nrc:
ping @ekexium what's the status of this PR?

ekexium:
> ping @ekexium what's the status of this PR?

I've been busy in other things. Will update it soon.

ekexium:
@nrc PTAL, I think it's almost done. Based on the discussion on Slack, we keep the `ResponseAndKeys`.

nrc (approved):
lgtm

PR #253, opened 2021-04-09
Fix two bugs with insert and delete

### LINKED ISSUE #234: Committing a delete must unlock
When we delete in a pessimistic transaction, we first lock the key. However, if our transaction first inserts a value, then deletes it, we do this all in the buffer and our prewrite only sends a 'check not exists' message to TiKV. That leaves the key locked, we should unlock it.

nrc:
Fixes #234

PTAL @sticnarf 

Previously there were two bugs:

* If the user deletes a key, then inserts a value to the same key, then there will be false positive failure of the insert invariant. This is fixed by converting the insert into a put.
* If the user inserts then deletes a key, the delete is converted into a check that the value doesn't exist. However, if the transaction is pessimistic, TiKV cannot handle that check and the key is left locked (#234). This is fixed by checking for previous values when we lock the key for insertion, then removing the insert rather than converting it if it is deleted.


ekexium:
If we consider the buffer as a state machine, there might be several more issues that are worth consideration:
1. Locked + Lock: we don't need to lock again, unless auto_heartbeat disabled
2. CheckNotExist + Delete: should be CheckNotExist
3. Cached + Insert: we should report an error immediately?
4. Locked(with cache) + Insert: should report error?
5. Locked(without cache) + Insert: insert?
6. Put + Insert: should be error

And I think `CheckNotExist` is a little special one. I suggest we treat it as a normal operation in the buffer, though we don't provide a direct corresponding operation in `Transaction`. So that for example in https://github.com/tikv/client-rust/blob/4870985dcdce8faaab9f701b055e1deb8ffe3d52/src/transaction/buffer.rs#L20-L22, we should also exclude `CheckNotExist`.

ekexium:
And the different implementations of the semantics of `CheckNotExist` in optimistic/pessimistic txns can be fixed in another PR 🤦🏻‍♂️. What a weird protocol

sticnarf:
> If we consider the buffer as a state machine, there might be several more issues that are worth consideration:
> 
>     1. Locked + Lock: we don't need to lock again, unless auto_heartbeat disabled
> 
>     2. CheckNotExist + Delete: should be CheckNotExist
> 

+1 for "Locked + Lock".

CheckNotExist was introduced for TiDB to simulate MySQL like behavior in optimistic transactions. I am truly not sure what users expect under "insert and delete". Is there a case people expect commit not to fail under this case?

There may also be an alternative that we always check existence inplace at insert instead postponing it until commit.

tests/integration_tests.rs
```diff
@@ -596,6 +596,49 @@ async fn pessimistic_rollback() -> Result<()> {
...
 
+#[tokio::test]
+#[serial]
+async fn pessimistic_delete() -> Result<()> {
+    clear_tikv().await;
+    let client =
+        TransactionClient::new_with_config(vec!["127.0.0.1:2379"], Default::default()).await?;
+
+    // The transaction will lock the keys and must release the locks on commit, even though we don't
+    // values.
```
ekexium, line 607: Does the comment need update?

sticnarf (approved):
This PR itself looks good. The remaining problems may not be the subject of this PR :)

ekexium (commented):
LGTM

ti-srebot (approved):

nrc:
https://github.com/rust-lang/rust/pull/82608 caused a compiler regression which causes the build to fail (due to spurious errors in the clang-sys crate). I've added a rust-toolchain file to address this temporarily.

PR #254, opened 2021-04-09
Don't set primary key in a pessimistic lock request until it succeeds

ekexium:
Pessimistic lock requests can fail. If we set the primary key before it is executed, and it fails, the primary key may be invalid.


src/transaction/transaction.rs
```diff
@@ -664,6 +670,9 @@ impl<PdC: PdClient> Transaction<PdC> {
             .plan();
         let pairs = plan.execute().await;
 
+        // primary key will be set here if needed
+        self.buffer.get_primary_key_or(&first_key).await;
```
nrc, line 674: This seems like an odd API - I think it should be something like `primary_key_or` and not return the existing primary key. Or we could keep the call earlier and use `get_primary_key_or_set` and return an empty key if it was not previously set.

ekexium: How about just renaming it to `get_primary_key_or_insert`? Similar to `Option::get_or_insert`.

nrc: My point is not so much the name, but that this is the only place it is used and the caller throws away the returned value.

ekexium: Oh that makes sense.

nrc (approved):
lgtm

PR #256, opened 2021-04-09
Remove the Mutex in Buffer

ekexium:
I don't see the reason of having a Mutex here... Do I miss anything?


src/transaction/buffer.rs
```diff
@@ -34,24 +33,24 @@ impl InnerBuffer {
...
 pub struct Buffer {
-    mutations: Mutex<InnerBuffer>,
+    inner: InnerBuffer,
 }
 
 impl Buffer {
     pub fn new(is_pessimistic: bool) -> Buffer {
         Buffer {
-            mutations: Mutex::new(InnerBuffer::new(is_pessimistic)),
+            inner: InnerBuffer::new(is_pessimistic),
```
nrc, line 42: If we don't have the Mutex, then I don't think we need the inner/outer split.

nrc (commented):
Nice, that this works, good idea to try it!

src/transaction/buffer.rs
ekexium: Yes. I hoped to make the PR simple :). Will update it.

nrc (approved):
lgtm!

PR #258, opened 2021-04-12
Make get_for_update work for optimistic txns

ekexium:
I think we can make `get_for_update` and `batch_get_for_update` support optimistic transactions, just like `SELECT ... FOR UPDATE` in TiDB.

Then the API looks more consistent: in both optimistic and pessimistic transactions, `lock_keys` just lock the keys, and `(batch_)get_for_update` locks keys and retrieves values.


src/transaction/transaction.rs
```diff
@@ -116,6 +116,17 @@ impl<PdC: PdClient> Transaction<PdC> {
...
+    ///
+    /// # Optimistic transaction
+    /// Once resolved this request will retrieve the value just like a normal `get` request,
+    /// and "locks" the key. This lock will not affect other (concurrent) transactions, but will
+    /// prevent the current transaction from successfully committing if there is another write
+    /// containing the "locked" key which is committed between the start and commit of the current transaction.
+    ///
+    /// The value is read from the `start timestamp`, thus it is cached in the local buffer.
+    ///
+    /// # Pessimistic transaction
```
nrc, line 129: This documentation is not very clear to me. Is it the case that the operation behaves the same in both cases, except that a different kind of lock is used? In the optimistic case, is there actually a lock or not? How is this implemented in TiKV?

nrc: From this PR, it looks like there is nothing special about the lock in the optimistic case.

ekexium: I think the differences are:
1. Optimistic reads at `start_ts`, pessimistic reads at `current_ts`.
2. Because of (1), in the client, optimistic caches the value, pessimistic does not cache the value.
3. In the optimistic case, the lock is written in prewrite and commit, so it cannot avoid concurrent transactions from writing the same key, but can only prevent itself from committing. While in the pessimistic case it is locked immediately, so other concurrent transactions won't be able to mutate the locked key.

nrc (approved):
lgtm, thanks!

PR #259, opened 2021-04-12
Allow txn to rollback after commit failed

ekexium:
If a transaction failed to commit (e.g. due to write conflict), it should be allowed to rollback.


nrc (approved):
lgtm

PR #262, opened 2021-04-14
Support atomic operations (CAS, put, and delete)

ekexium:
This is useful for the Jepsen test for TiKV.


nrc (commented):
Rather than add atomic versions of the commands, I think we should have api similar to `with_cf` (i.e., `with_atomic`) which gives a client which sets the atomic flag for all possible operations.

ekexium:
> Rather than add atomic versions of the commands, I think we should have api similar to `with_cf` (i.e., `with_atomic`) which gives a client which sets the atomic flag for all possible operations.

Sounds reasonable. But it might be confusing for users to know which operations are atomic and which are not. For example, we don't support atomic `batch_delete` but have atomic `delete`.

nrc:
> But it might be confusing for users to know which operations are atomic and which are not. For example, we don't support atomic batch_delete but have atomic delete.

Either we could panic, or we could use phantom types so that only the correct methods are available for an atomic/non-atomic client.

ekexium:
I'd rather not use phantom types here because that may require the user to write some type parameter and bring more cognitive burden.

nrc:
> I'd rather not use phantom types here because that may require the user to write some type parameter and bring more cognitive burden.

I don't think there should be any extra burden - there is a clear default, so we can use a default type parameter, and for the non-default case, type inference should spare the user from writing the types.

(Not saying we should definitely use phantom types here, just for future reference).

ekexium:
> > I'd rather not use phantom types here because that may require the user to write some type parameter and bring more cognitive burden.
> 
> I don't think there should be any extra burden - there is a clear default, so we can use a default type parameter, and for the non-default case, type inference should spare the user from writing the types.
> 
> (Not saying we should definitely use phantom types here, just for future reference).

I was considering when users pass the client objects around among functions, they have to write the type like `RawClient<Ph>`?

tikv-client-common/src/errors.rs
```diff
@@ -28,6 +28,12 @@ pub enum Error {
     /// An operation requires a primary key, but the transaction was empty.
     #[error("transaction has no primary key")]
     NoPrimaryKey,
+    #[error(
+        "The operation does is not supported in raw-atomic mode, please consider using an atomic client"
+    )]
+    UnsupportedInAtomicMode,
+    #[error("The operation is only supported in raw-atomic mode, please consider using a non-atomic raw client")]
+    UnsupportedInNonAtomicMode,
```
nrc, line 36: I think just one UnsupportedMode error would be better.

src/raw/client.rs
```diff
@@ -89,6 +95,22 @@ impl Client {
...
+    }
+
+    /// Set to use the atomic mode.
+    ///
+    /// The only reason of using atomic mode is the
+    /// [`compare_and_swap`](Client::compare_and_swap) operation. To guarantee
+    /// the atomicity of CAS, write operations like [`put`](Client::put) or
+    /// [`delete`](Client::delete) in atomic mode are more expensive. Some
+    /// operations are not supported in the mode.
+    pub fn with_atomic(&self) -> Client {
```
nrc, line 109: maybe we should rename to `with_atomic_for_cas` so as not to give the impression that the user should use this to get atomic put/delete (which are atomic in the usual sense with or without this option).

ekexium: Yeah it makes a lot of sense

tikv-client-common/src/errors.rs
ekexium: OK. I'm also thinking that the error handling does not provide a good experience. It's often hard to determine the concrete error from the error message. Maybe we should attach the trace with the error or find some other ways to improve it.

nrc (approved):
lgtm

nrc:
> I was considering when users pass the client objects around among functions, they have to write the type like RawClient<Ph>?

I think this would't happen very often, and when it did, then 90% of the time would use the default parameter so there is no need for the `<Ph>`. And in the times when you are passing a client to another function and the caller sets some setting, then I think it is actually nice to have that documented and checked in the type

ekexium:
> > I was considering when users pass the client objects around among functions, they have to write the type like RawClient?
> 
> I think this would't happen very often, and when it did, then 90% of the time would use the default parameter so there is no need for the `<Ph>`. And in the times when you are passing a client to another function and the caller sets some setting, then I think it is actually nice to have that documented and checked in the type

Right, sounds reasonable.

PR #266, opened 2021-04-15
Don't resolve lock in raw requests

ekexium:
There shouldn't be locks in a raw-only TiKV cluster. It's users duty to guarantee it.


nrc (approved):
lgtm

Issue #251, opened 2021-03-15
Panic with multiple regions and 1pc


ekexium:
When there are multiple regions, 1PC transactions can panic.

PR #268, opened 2021-04-21
Support multi-reigon test

ekexium:
Previously bugs with multiple regions are hard to find in tests (e.g. #251), we should improve our test to cover such situation.

Changes:
- Support integration tests with multiple regions
- Improve the `write_million` tests since they don't have to write much data to make region split. As a result tests should be faster

How it works:
- Use specific config for TiKV and PD to enforce region splitting
- Enforce region splitting in integration tests when `MULTI_REGION=1` is set

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,13 +48,15 @@ script:
   # For now we only run full integration tests on Linux. Here's why:
   # * Docker on OS X is not supported by Travis.
   # * Docker on Windows seems to not have the correct binary at `"/c/Program Files/Docker/Docker/DockerCli.exe" to switch it to Linux containers.
-  - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then docker run -d --net=host --name pd --rm pingcap/pd:nightly --name "pd" --data-dir "pd" --client-urls "http://127.0.0.1:2379" --advertise-client-urls "http://127.0.0.1:2379"; fi
-  - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then docker run -d --net=host --name kv --rm --ulimit nofile=90000:90000 pingcap/tikv:nightly --pd-endpoints "127.0.0.1:2379" --addr "127.0.0.1:2378" --data-dir "kv"; fi
+  - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then docker run -d -v $(pwd)/config:/config --net=host --name pd --rm pingcap/pd:nightly --name "pd" --data-dir "pd" --client-urls "http://127.0.0.1:2379" --advertise-client-urls "http://127.0.0.1:2379" --config /config/pd.toml; fi
+  - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then docker run -d -v $(pwd)/config:/config --net=host --name kv --rm --ulimit nofile=90000:90000 pingcap/tikv:nightly --pd-endpoints "127.0.0.1:2379" --addr "127.0.0.1:2378" --data-dir "kv" --config /config/tikv.toml; fi
   - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then docker ps; fi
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -48,13 +48,15 @@ script:
...
   - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then docker ps; fi
   - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then docker logs pd; fi
   - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then docker logs kv; fi
   - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then sleep 60; fi
-  - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then PD_ADDRS="127.0.0.1:2379" cargo test --all --features integration-tests -- --nocapture; fi
+  - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then PD_ADDRS="127.0.0.1:2379" MULTI_REGION=1 cargo test txn_ --all --features integration-tests -- --nocapture; fi
+  - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then PD_ADDRS="127.0.0.1:2379" MULTI_REGION=1 cargo test raw_ --all --features integration-tests -- --nocapture; fi
+  - if [[ $TRAVIS_OS_NAME == "linux" && $TRAVIS_RUST_VERSION == "nightly" ]]; then PD_ADDRS="127.0.0.1:2379" MULTI_REGION=1 cargo test misc_ --all --features integration-tests -- --nocapture; fi
```
nrc, line 59: can we use make on Travis?

tests/common/mod.rs
```diff
@@ -14,7 +21,59 @@ pub async fn clear_tikv() {
...
+            .map(|x| x.to_be_bytes().to_vec());
+
+        ensure_region_splitted(keys_1.chain(keys_2), 100).await?;
+    }
+
+    clear_tikv().await;
+    Ok(())
+}
+
+async fn ensure_region_splitted(
```
nrc, line 46:
```suggestion
async fn ensure_region_split(
```

tests/common/mod.rs
```diff
@@ -14,7 +21,59 @@ pub async fn clear_tikv() {
...
+    keys: impl IntoIterator<Item = impl Into<Key>>,
+    region_count: u32,
+) -> Result<()> {
+    if ctl::get_region_count().await? as u32 >= region_count {
+        return Ok(());
+    }
+
+    // 1. write plenty transactional keys
+    // 2. wait until regions splitted
```
nrc, line 55:
```suggestion
    // 2. wait until regions split
```

tests/common/mod.rs
```diff
@@ -14,7 +21,59 @@ pub async fn clear_tikv() {
...
+    for key in keys.into_iter() {
+        txn.put(key.into(), vec![0, 0, 0, 0]).await?;
+    }
+    txn.commit().await?;
+    let mut txn = client.begin_optimistic().await?;
+    let _ = txn.scan(vec![].., 2048).await?;
+    txn.commit().await?;
+
+    info!("splitting regions...");
+    loop {
```
nrc, line 68: Should we have a safety check exit in case PD never splits the regions?

tests/common/mod.rs
```diff
@@ -23,3 +82,10 @@ pub fn pd_addrs() -> Vec<String> {
         .map(From::from)
         .collect()
 }
+
+fn enable_multi_region() -> bool {
+    match env::var(ENV_ENABLE_MULIT_REGION) {
+        Ok(s) => s == "1" || s.to_lowercase() == "true",
```
nrc, line 88: I think it is usual practice to enable a feature if the env var is set to anything

ekexium: That's my first thought too. Then I thought `MULTI_REGION=0` would be weird. I don't know a lot about the convention here.

tests/common/mod.rs
ekexium: Good idea.

tests/common/mod.rs
nrc: I think you must unset an env var not set it to 0 if you want it to be falsey

ekexium: OK

nrc (approved):
lgtm, thanks for the changes

PR #276, opened 2021-05-05
README: quote version in dependencies

dveeden:
Without this it doesn't seem to get parsed as valid TOML


dveeden:
Whoops, looks like  #275 is doing the same thing

PR #277, opened 2021-05-07
Add getting started docs

nrc:
Adds a bit more detail to the README docs to cover using Tokio.

PTAL @ekexium

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -1,0 +1,64 @@
+# Getting started
+
+The TiKV client is a Rust library (crate). To use this crate in your project, add the following dependencies to your `Cargo.toml`:
+
+```toml
+[dependencies]
+tikv-client = "0.1"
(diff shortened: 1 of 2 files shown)


getting-started.md
```diff
@@ -0,0 +1,64 @@
+# Getting started
+
+The TiKV client is a Rust library (crate). To use this crate in your project, add the following dependencies to your `Cargo.toml`:
+
+```toml
+[dependencies]
+tikv-client = "0.1"
```
ekexium, line 7: Shall we keep it 0.0.99 and change it when we publish 0.1?

ekexium (commented):
LGTM

getting-started.md
nrc: I hope it won't take long, and we use 0.0.99 in the README

PR #279, opened 2021-05-07
CI: Try to increase region-split-keys

### LINKED ISSUE #278: CI: EpochNotMatch
Sometimes CI fails for `RegionError(Error { message: "EpochNotMatch ...`. It's probably caused by frequent region split.

ekexium:
Fixes #278


nrc (approved):
lgtm

PR #290, opened 2021-05-11
doc: 1.0 roadmap

ekexium:
The roadmap towards 1.0.

### DIFF
--- a/doc/1.0-roadmap.md
+++ b/doc/1.0-roadmap.md
@@ -53,10 +53,10 @@ The timeline is a rough estimate. It is subject to change if we find new issues,
 ##### May 2021
 Release 0.1.0
 
-##### July 2021
+##### Aug 2021
 Release 0.2.0
 
-##### Sep 2021
+##### Oct 2021
 Release 0.3.0
 
 ##### Jan 2022
--- a/doc/1.0-roadmap.md
+++ b/doc/1.0-roadmap.md
@@ -1,6 +1,6 @@
 # client-rust 1.0 roadmap
 
-The client-rust project complements the TiKV ecosystem by providing a reliable Rust client. It can be a foundation for implementing clients in other languages.
+The client-rust project complements the TiKV ecosystem by providing a reliable Rust client. It can be a foundation for implementing clients in other languages. There have been preliminary works in [client-py](https://github.com/tikv/client-py) and [client-cpp](https://github.com/tikv/client-cpp). There are also plans for general-purpose client-java and [client-node](https://github.com/tikv/tikv/issues/10054).
 
 The document describes our plan to push client-rust towards its general availability (1.0).
 
@@ -69,7 +69,7 @@ We will take a relative longer time to let users try the tikv-client, and improv
 ## Risks
 
 - Inaccurate estimated workload. Considering the complexity of the client in TiDB, we may underestimate the workload.
-- Lacking staff. Regular contributors include Ziqian Qin, Nick Cameron and Andy Lok.
+- Lacking staff. Regular contributors include Ziqian Qin and Andy Lok.
 - Too much or too little external contribution - too much means core contributors don’t have time to focus on code and design, or that we add too many new features which do not get us to 1.0. Too little means we can’t iterate effectively based on user experience.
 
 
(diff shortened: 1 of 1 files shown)


doc/1.0-roadmap.md
```diff
@@ -0,0 +1,110 @@
...
+- Usability improvement based on user feedback
+
+## Estimated Timeline
+
+The timeline is a rough estimate. It is subject to change if we find new issues, underestimate the workload, or have insufficient contributors’ time.
+
+##### May 2021
+Release 0.1.0
+
+##### July 2021
```
andylokandy, line 56: The time for it is tight...

ekexium: Yes. I adjusted the plan a bit. 

Since we cannot precisely evaluate the manpower, we can hardly view the timeline as hard deadlines. It's very likely to be optimistic.

doc/1.0-roadmap.md
```diff
@@ -0,0 +1,110 @@
...
+The document describes our plan to push client-rust towards its general availability (1.0).
+
+## Deliverable
+
+client-rust 1.0 that
+- enables new users to start using it in 10 mins
+- has been heavily tested under simulated and real workload
+- is easy to use, debug and seek help from documentation
+- is performant
+- supports the most-wanted features: GC, async commit, etc.
```
zz-jason, line 14: "GC": does it mean the API to manually trigger GC?

doc/1.0-roadmap.md
```diff
@@ -0,0 +1,110 @@
...
+Release 0.1.0
+
+##### Aug 2021
+Release 0.2.0
+
+##### Oct 2021
+Release 0.3.0
+
+##### Jan 2022
+Release 1.0
```
zz-jason, line 63: we may need to shorten the timeline

doc/1.0-roadmap.md
```diff
@@ -0,0 +1,110 @@
+# client-rust 1.0 roadmap
+
+The client-rust project complements the TiKV ecosystem by providing a reliable Rust client. It can be a foundation for implementing clients in other languages.
```
zz-jason, line 3: could you list all the planned to be supported languages here?

doc/1.0-roadmap.md
```diff
@@ -0,0 +1,110 @@
...
+
+client-rust 1.0 that
+- enables new users to start using it in 10 mins
+- has been heavily tested under simulated and real workload
+- is easy to use, debug and seek help from documentation
+- is performant
+- supports the most-wanted features: GC, async commit, etc.
+
+
+## Milestones
```
zz-jason, line 17: Would it be better to illustrate the milestones for RawKV and TxnKV respectively?

doc/1.0-roadmap.md
ekexium: Yes. In the current implementation, we only provide a `gc` function to update GC safepoint. Because client-rust serves as a library, users must manually manage the GC and trigger it appropriately. 

Things might change if the GC mechanism changes (https://github.com/tikv/rfcs/pull/65).

doc/1.0-roadmap.md
ekexium: I think the timeline is a just a reference since we cannot accurately evaluate how much effort will be put into the project. We can shorten the timeline (as in the initial version), but it doesn't mean we can manage to stick to it.

Nick suggested the long gap between 0.3 and 1.0. @andylokandy suggested we put off the release of 0.2 a bit - from July to August.

doc/1.0-roadmap.md
ekexium: IMO it's unnecessary. Currently, I don't see any features in the plan that are only for RawKV. So the milestones are for TxnKV (or both). 
If there may be divergent progress in TxnKV and in RawKV (like raw_coprocessor?), then we can set separate milestones for them.

leiysky:
Any update on this?

ekexium:
> Any update on this?

It's up-to-date. We haven't merged it in case there are more reviews & comments.

andylokandy:
Since there is no more comment, I'm going to merge it.

andylokandy (approved):

PR #116, opened 2019-09-05
Add region cache for PD

sticnarf:

Issue #246, opened 2021-03-05
Don't bind to tokio


ekexium:
We use several tokio components in the code, which requires the tokio runtime. We should replace them with runtime independent ones, e.g. `futures-timer`.

PR #291, opened 2021-05-19
The minimal region cache

ekexium:
The PR adds a minimal region cache implementation. It's far away from maturity. It's intended to discuss how the structure should look.

Related issue: #299 

## Main changes:
- Place the region cache as a component in `PdRpcClient`. `PdRpcClient` -> region cache -> retry client (the low level pd client)
- Make single-region plan a special case of multi-region plan

Credit to #116, #121 and #139. Since there have been some refactoring after the previous work, I didn't continue development based on those commits.

## Naive manual benchmark

Environment: single machine, 1 PD, 3 TiKV, default scheduler + shuffle-region + shuffle-leader
Workload: 6 threads, 512 workers,  50% single-key-read, 50% single-key-write

- zipfian distribution, pessimistic txns

master: TPS: 1943.9. Latency: mean: 41ms, P99: 1018ms, maximum: 1815ms
This PR: TPS 2108.5. Latency: mean: 12ms, P99: 108ms, maximum: 1061ms

- uniform distribution, optimistic txns

master: TPS: 4279.4, mean: 70ms, P99: 1048ms, maximum: 2601ms
This PR: TPS: 21129.8, mean: 22ms, P99: 44ms, maximum: 286ms

## Todo
- [x] Remove `read_through_cache` parameters, if we can always know the invalid cache entry
- [x] Treat single-region as a special case of multi-region
- [ ] Test
- [ ] Save unnecessary PD requests if they are querying the same region
- [ ] Finer grained error handling
- [ ] Naming

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ regex = "1"
 serde = "1.0"
 serde_derive = "1.0"
 thiserror = "1"
-tokio = { version = "1", features = [ "sync", "time" ] }
+tokio = { version = "1", features = [ "sync", "rt-multi-thread", "macros" ] }
 async-recursion = "0.3"
 
 tikv-client-common = { version = "0.1.0", path = "tikv-client-common" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,8 @@ regex = "1"
 serde = "1.0"
 serde_derive = "1.0"
 thiserror = "1"
-tokio = { version = "1.0", features = [ "sync", "time" ] }
+tokio = { version = "1", features = [ "sync", "time" ] }
+async-recursion = "0.3"
 
 tikv-client-common = { version = "0.0.99", path = "tikv-client-common" }
 tikv-client-pd = { version = "0.0.99", path = "tikv-client-pd" }
@@ -48,7 +49,7 @@ proptest = "1"
 proptest-derive = "0.3"
 serial_test = "0.5.0"
 simple_logger = "1"
-tokio = { version = "1.0", features = [ "sync", "rt-multi-thread", "macros" ] }
+tokio = { version = "1", features = [ "sync", "rt-multi-thread", "macros" ] }
 reqwest = {version = "0.11", default-features = false, features = ["native-tls-vendored"]}
 serde_json = "1"
 
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -15,7 +15,7 @@ use async_trait::async_trait;
 use futures::{future::try_join_all, stream::StreamExt};
 use std::{collections::VecDeque, marker::PhantomData, sync::Arc};
 use tikv_client_proto::{errorpb::EpochNotMatch, kvrpcpb};
-use tikv_client_store::{HasError, HasRegionError, KvClient};
+use tikv_client_store::{HasKeyErrors, HasRegionError, HasRegionErrors, KvClient};
 use tokio::sync::Semaphore;
 
 /// A plan for how to execute a request. A user builds up a plan with various
@@ -76,7 +76,7 @@ pub struct RetryableMultiRegion<P: Plan, PdC: PdClient> {
 
 impl<P: Plan + Shardable, PdC: PdClient> RetryableMultiRegion<P, PdC>
 where
-    P::Result: HasError,
+    P::Result: HasKeyErrors + HasRegionError,
 {
     // A plan may involve multiple shards
     #[async_recursion]
@@ -123,8 +123,8 @@ where
         let mut resp = plan.execute().await?;
         drop(permit);
 
-        if let Some(e) = resp.error() {
-            Ok(vec![Err(e)])
+        if let Some(e) = resp.key_errors() {
+            Ok(vec![Err(Error::MultipleKeyErrors(e))])
         } else if let Some(e) = resp.region_error() {
             match backoff.next_delay_duration() {
                 Some(duration) => {
@@ -269,7 +269,7 @@ impl<P: Plan, PdC: PdClient> Clone for RetryableMultiRegion<P, PdC> {
 #[async_trait]
 impl<P: Plan + Shardable, PdC: PdClient> Plan for RetryableMultiRegion<P, PdC>
 where
-    P::Result: HasError,
+    P::Result: HasKeyErrors + HasRegionError,
 {
     type Result = Vec<Result<P::Result>>;
 
@@ -325,12 +325,12 @@ pub struct Collect;
 /// A merge strategy that only takes the first element. It's used for requests
 /// that should have exactly one response, e.g. a get request.
 #[derive(Clone, Copy)]
-pub struct CollectFirst;
+pub struct CollectSingle;
 
 #[macro_export]
 macro_rules! collect_first {
     ($type_: ty) => {
-        impl Merge<$type_> for CollectFirst {
+        impl Merge<$type_> for CollectSingle {
             type Out = $type_;
 
             fn merge(&self, mut input: Vec<Result<$type_>>) -> Result<Self::Out> {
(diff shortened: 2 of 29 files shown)


src/request/plan.rs
```diff
@@ -131,6 +322,25 @@ impl<In: Clone + Send + Sync + 'static, P: Plan<Result = Vec<Result<In>>>, M: Me
 #[derive(Clone, Copy)]
 pub struct Collect;
 
+/// A merge strategy that only takes the first element. It's used for requests
+/// that should have exactly one response, e.g. a get request.
+#[derive(Clone, Copy)]
+pub struct CollectFirst;
```
andylokandy, line 328:
```suggestion
pub struct CollectSingle;
```

tikv-client-store/src/errors.rs
```diff
@@ -182,13 +182,12 @@ impl<T: HasRegionError, E> HasRegionError for Result<T, E> {
 
 impl<T: HasRegionError> HasRegionError for Vec<T> {
     fn region_error(&mut self) -> Option<Error> {
```
andylokandy, line 184: Return `Vec<Errror>` instead of `Error::MultipleErrors`.

tests/integration_tests.rs
```diff
@@ -360,7 +359,7 @@ async fn txn_bank_transfer() -> Result<()> {
             .await?;
         txn.put(bob.clone(), bob_balance.to_be_bytes().to_vec())
             .await?;
-        txn.commit().await?;
+        let _ = txn.commit().await;
```
andylokandy, line 362: why ignore?

ekexium: Failure of a single transfer shouldn't affect the final invariant check. I think it's fine to revert the change now.

tests/common/mod.rs
```diff
@@ -88,8 +88,8 @@ async fn ensure_region_split(
 
 pub fn pd_addrs() -> Vec<String> {
     env::var(ENV_PD_ADDRS)
-        .expect(&format!("Expected {}:", ENV_PD_ADDRS))
-        .split(",")
+        .unwrap_or_else(|_| "127.0.0.1:2379".to_owned())
```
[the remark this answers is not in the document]
ekexium: Is it better an INFO or a WARN? I guess it would be the most common case to use the default address when testing.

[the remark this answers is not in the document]
andylokandy: info is good for me

src/request/plan.rs
```diff
@@ -63,42 +66,230 @@ impl<Req: KvRequest + HasKeys> HasKeys for Dispatch<Req> {
...
+                        region_store,
+                        e.take_epoch_not_match(),
+                    )
+                    .await
+                } else if e.has_server_is_busy()
+                    || e.has_stale_command()
+                    || e.has_region_not_found()
+                    || e.has_max_timestamp_not_synced()
+                {
+                    pd_client.invalidate_region_cache(ver_id).await;
```
sticnarf, line 198: I think `ServerIsBusy` and `MaxTimestampNotSynced` don't need to invalidate cache.

src/request/plan.rs
```diff
@@ -63,42 +66,230 @@ impl<Req: KvRequest + HasKeys> HasKeys for Dispatch<Req> {
...
+                    Ok(false)
+                } else if e.has_raft_entry_too_large() {
+                    Err(Error::RegionError(e))
+                } else {
```
sticnarf, line 202: I recommend logging unhandled unknown error.

src/request/plan.rs
```diff
@@ -63,42 +66,230 @@ impl<Req: KvRequest + HasKeys> HasKeys for Dispatch<Req> {
...
+                    pd_client.invalidate_region_cache(ver_id).await;
+                    Ok(false)
+                }
+            }
+            // errors from PD requests, backoff and retry
```
sticnarf, line 207: Emm, I don't see why there can be PD errors.

Cargo.toml
```diff
@@ -32,7 +32,8 @@ regex = "1"
 serde = "1.0"
 serde_derive = "1.0"
 thiserror = "1"
-tokio = { version = "1.0", features = [ "sync", "time" ] }
+tokio = { version = "1", features = [ "sync", "time" ] }
```
sticnarf, line 35: I don't know why it works before, but it looks that the `rt` feature is needed. Because of `tokio::spawn`.

src/request/plan.rs
```diff
@@ -63,42 +66,230 @@ impl<Req: KvRequest + HasKeys> HasKeys for Dispatch<Req> {
...
+        if let Some(e) = resp.key_errors() {
+            Ok(vec![Err(Error::MultipleKeyErrors(e))])
+        } else if let Some(e) = resp.region_error() {
+            match backoff.next_delay_duration() {
+                Some(duration) => {
+                    let region_error_resolved =
+                        Self::handle_region_error(pd_client.clone(), e, region_store).await?;
+                    // don't sleep if we have resolved the region error
+                    if !region_error_resolved {
+                        futures_timer::Delay::new(duration).await;
```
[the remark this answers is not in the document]
ekexium: There is a wish that we can remove the dependency of tokio one day...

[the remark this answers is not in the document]
sticnarf: At first I refused to introduce the runtime. But I must admit, with a runtime, it is both easy for us to design the API and code, and for the users to use...

src/request/plan.rs
```diff
@@ -63,42 +66,230 @@ impl<Req: KvRequest + HasKeys> HasKeys for Dispatch<Req> {
...
-                    Some(e) => Err(e),
-                    None => Ok(response),
-                }
-            })
-            .collect()
-            .await)
+        // The plans to be executed. There can be more children plans because one
+        // shard can become more if it returns region error.
+        let mut children_plans = VecDeque::new();
+        children_plans.push_back((self.inner.clone(), self.backoff.clone()));
```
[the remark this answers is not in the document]
ekexium: My bad. It's stale and I forgot to remove it.

src/request/plan.rs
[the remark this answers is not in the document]
sticnarf: The client inevitably needs some background work to do. Rust does not provide an async runtime interface, and it is not so suitable to start a runtime by the client. Relying on a specific runtime looks like the only solution in the near future :(

src/request/plan.rs
ekexium: I didn't considerably think whether it is possible. Just in case :)

src/request/plan.rs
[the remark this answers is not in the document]
ekexium: All right. We can do it in another PR. #246

src/request/plan.rs
sticnarf: The function name is `handle_region_error`, so it is unreasonable that non-region errors occur here. Maybe the definition of `HasRegionError::region_error` can be changed to just return an `errorpb::Error)`.

But it indeed reminds me that the error handling for PD errors is coarse grained. Not a big problem for now as there is typically not much trouble PD can cause.

ekexium: >The function name is handle_region_error, so it is unreasonable that non-region errors occur here.

Right. I just made the change.

ekexium: > Maybe the definition of HasRegionError::region_error can be changed to just return an errorpb::Error)

Good idea

sticnarf (commented):
I'm fine with the current code.

By the way, the backoff design is a bit from the client go. It is simpler because it always use the same backoff type (for different errors) and does not record the backoff history. These features are nice to have, but not an urgent task.

ekexium:
> By the way, the backoff design is a bit from the client go. It is simpler because it always use the same backoff type (for different errors) and does not record the backoff history. These features are nice to have, but not an urgent task.

Yep, there is an issue there #310

src/request/plan.rs
```diff
@@ -63,42 +66,213 @@ impl<Req: KvRequest + HasKeys> HasKeys for Dispatch<Req> {
...
+        error: EpochNotMatch,
+    ) -> Result<bool> {
+        let ver_id = region_store.region_with_leader.ver_id();
+        if error.get_current_regions().is_empty() {
+            pd_client.invalidate_region_cache(ver_id).await;
+            return Ok(true);
+        }
+
+        for r in error.get_current_regions() {
+            if r.get_id() == region_store.region_with_leader.id() {
```
youjiali1995, line 221: It doesn't insert new regions to the cache?

youjiali1995 (commented):
The minimal region cache LGTM.

src/request/plan.rs
ekexium: The current error handling is coarse-grained. I'd like to leave these improvements to following PRs to keep this one relatively simple.

sticnarf:
Please fix the clippy warnings.

sticnarf (approved):
LGTM

PR #293, opened 2021-05-21
Add raw_coprocessor

andylokandy:
Issue: https://github.com/tikv/tikv/issues/9747

Blocking on https://github.com/pingcap/kvproto/pull/765

Add support for sending raw coprocessor requests.

## Details

- Rename `PreserveKey` to `PreserveShard` and remove `HasKeys`.
- Rename `CollectAndMatchKeys` to `CollectWithShard`
- Rename `ResponseAndKeys` to `ResponseWithShard`
- Add RawCoproceesorRequest wrapper for dynamically building custom request data after sharding


andylokandy:
cc @Jibbow

src/raw/client.rs
```diff
@@ -491,6 +492,28 @@ impl Client {
         plan.execute().await
     }
 
+    pub async fn coprocessor(
+        &self,
+        copr_name: String,
+        copr_version_req: String,
```
Jibbow, line 498: Should we maybe use `semver::VersionReq` here so that we can be sure that the user passes a valid string?

Jibbow (approved):

src/raw/client.rs
andylokandy: Good idea!

src/transaction/requests.rs
```diff
@@ -346,61 +342,62 @@ impl Shardable for kvrpcpb::PessimisticLockRequest {
...
+    ) -> Result<Self::Out> {
         input
             .into_iter()
-            .flat_map_ok(|mut resp| {
+            .flat_map_ok(|ResponseWithShard(mut resp, mutations)| {
                 let values = resp.take_values();
+                let values_len = values.len();
                 let not_founds = resp.take_not_founds();
-                let v: Vec<_> = if not_founds.is_empty() {
+                let kvparis = mutations
```
ekexium, line 380:
```suggestion
                let kvpairs = mutations
```

src/request/shard.rs
```diff
@@ -30,21 +36,27 @@ pub trait Shardable {
     type Shard: Send;
 
     fn shards(
```
ekexium, line 38: I think there can be a better name like `group_shards_by_store`

src/request/shard.rs
```diff
@@ -30,21 +36,27 @@ pub trait Shardable {
...
-        &self,
+        shard: Self::Shard,
         pd_client: &Arc<impl PdClient>,
     ) -> BoxStream<'static, Result<(Self::Shard, Store)>>;
 
+    fn get_shard(&self) -> Self::Shard;
```
ekexium, line 43: I think `get_` had better return a reference. Let the user clone it if necessary.

src/request/plan.rs
```diff
@@ -315,102 +301,64 @@ where
...
+        Ok(ResponseWithShard(res, shard))
     }
 }
 
-pub trait HasKeys {
-    fn get_keys(&self) -> Vec<Key>;
-}
-
 // contains a response and the corresponding keys
 // currently only used for matching keys and values in pessimistic lock requests
```
ekexium, line 336: The comment needs update. And I think we can add a hint saying that the order of the elements in the response and in the shard must match.

ekexium (commented):
I think an alternative of the method is to add a new plan that do some pre-pocessing. The plan can be placed between `MultiRegion` and `Dispatch`, so that we don't have to create a special request type.

We should merge #292 first.

src/raw/requests.rs
```diff
@@ -316,6 +330,107 @@ impl Process<kvrpcpb::RawCasResponse> for DefaultProcessor {
...
+    ) -> Result<Self::Out> {
+        input?
+            .into_iter()
+            .map(|shard_resp| {
+                shard_resp.map(|ResponseWithShard(mut resp, ranges)| {
+                    (
+                        resp.take_data(),
+                        ranges
+                            .into_iter()
+                            .map(|range| range.start_key.into()..range.end_key.into())
```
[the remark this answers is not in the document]
andylokandy: Empty end key is handled in TiKV. Besides, this is how `KeyRange` is defined in kvproto.

src/request/shard.rs
andylokandy: If the shard is computed, `self` doesn't have the ownership, thus not able to return a reference. Anyway, the only usage of get_shard are `MultiRegion` and `PreserveShard` which will always clone it.

src/request/shard.rs
andylokandy: I'd like to make a following PR to refactor these parts.

src/request/shard.rs
ekexium: Then we can name it after something like `get_cloned_shard`

andylokandy: I reverted get_shard.

andylokandy:
@ekexium Conflict resovled, PTAL

andylokandy:
oh, stupid me...

andylokandy:
> oh, stupid me...

Fixed in https://github.com/tikv/client-rust/commit/eb1d2da05cc857818e078bc3d91ce50343031642

Issue #267, opened 2021-04-20
Logging


nrc:
We do very minimal logging - a total of six locations log any data. Either we should do proper logging, or we should remove those six logging locations and the dependency.

ekexium:
Checklist for the slog migration:
- [ ] Pass the logger around
- [ ] Update examples and docs

PR #294, opened 2021-05-25
added logs

shashwatj07:
~resolves~ Part of #267


ekexium:
Thanks! I think in most cases we don't want to see plenty of such logs. So they'd better be in debug level.

andylokandy:
Maybe we need to use slog so that we can connect the logs between user invocation and internal requests.

ekexium (approved):
LGTM. Thanks!

In terms of other logging, I think we can have more `info!` for non-trivial events. Maybe some `warn!` for unexpected but recoverable errors.

PR #296, opened 2021-05-26
Use TiUP to start cluster in CI

ekexium:
CI often failed because of docker pull rate limit. The PR tries to use TiUP instead.


config/tikv.toml
```diff
@@ -7,4 +7,10 @@ batch-split-limit = 100
...
 pd-store-heartbeat-tick-interval = "5s"
-split-region-check-tick-interval = "1s"
\ No newline at end of file
+split-region-check-tick-interval = "1s"
+
+[rocksdb]
+max-open-files = 10000
+
+[raftdb]
+max-open-files = 10000
```
andylokandy, line 16:
```suggestion
max-open-files = 10000

```

andylokandy (approved):

PR #298, opened 2021-05-30
migrate to slog

shashwatj07:
See https://github.com/tikv/client-rust/pull/294#issuecomment-849344011

### DIFF
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -44,11 +44,14 @@ impl Client {
     /// # use tikv_client::RawClient;
     /// # use futures::prelude::*;
     /// # futures::executor::block_on(async {
-    /// let client = RawClient::new(vec!["192.168.0.100"]).await.unwrap();
+    /// let client = RawClient::new(vec!["192.168.0.100"], None).await.unwrap();
     /// # });
     /// ```
-    pub async fn new<S: Into<String>>(pd_endpoints: Vec<S>) -> Result<Client> {
-        Self::new_with_config(pd_endpoints, Config::default()).await
+    pub async fn new<S: Into<String>>(
+        pd_endpoints: Vec<S>,
+        logger: Option<Logger>,
+    ) -> Result<Client> {
+        Self::new_with_config(pd_endpoints, Config::default(), logger).await
     }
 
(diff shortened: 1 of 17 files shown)


ekexium:
I think we should let the user pass the logger when creating a Client.
It may accept an `Option<Logger>` and create a default one if it's `None`.

shashwatj07:
> I think we should let the user pass the logger when creating a Client.
> It may accept an `Option<Logger>`. We can create a default one if it's `None`.

@ekexium done, please take a look

src/raw/client.rs
```diff
@@ -67,15 +70,22 @@ impl Client {
...
     /// # });
     /// ```
     pub async fn new_with_config<S: Into<String>>(
         pd_endpoints: Vec<S>,
         config: Config,
+        optional_logger: Option<Logger>,
     ) -> Result<Client> {
-        let plain = slog_term::PlainSyncDecorator::new(std::io::stdout());
-        let logger = Logger::root(slog_term::FullFormat::new(plain).build().fuse(), o!());
+        let logger = match optional_logger {
```
ekexium, line 82: Can be simplified as `unwrap_or_else`

src/transaction/client.rs
```diff
@@ -75,15 +78,22 @@ impl Client {
...
     /// # });
     /// ```
     pub async fn new_with_config<S: Into<String>>(
         pd_endpoints: Vec<S>,
         config: Config,
+        optional_logger: Option<Logger>,
     ) -> Result<Client> {
-        let plain = slog_term::PlainSyncDecorator::new(std::io::stdout());
-        let logger = Logger::root(slog_term::FullFormat::new(plain).build().fuse(), o!());
+        let logger = match optional_logger {
```
ekexium, line 90: Same here

ekexium (commented):
Just a nitpick, mostly LGTM. Thanks!
Please fix the CI and the `examples`.

src/raw/client.rs
shashwatj07: done

src/transaction/client.rs
shashwatj07: done

shashwatj07:
@ekexium ping

ekexium (approved):

Issue #302, opened 2021-06-14
How to run a GC job?


Little-Wallace:
I think the readme should better give an example about how to run a GC job.  According to my understanding, when there are some transactions running and have not committed, the timestamp which we call `gc` with shall be less than the start time of all the running transactions, is it right? 
So the transaction shall allow user to get start_time from a transaction.

PR #303, opened 2021-06-14
Add method of gettimestamp for transaction

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

to close: https://github.com/tikv/client-rust/issues/302 

We need a method to get started time of transaction, so that we can know that what is the time point of the earliest started transaction which has still not commit or rollback.

### DIFF
--- a/src/transaction/client.rs
+++ b/src/transaction/client.rs
@@ -183,7 +183,9 @@ impl Client {
 
     /// Request garbage collection (GC) of the TiKV cluster.
     ///
-    /// GC deletes MVCC records whose timestamp is lower than the given `safepoint`.
+    /// GC deletes MVCC records whose timestamp is lower than the given `safepoint`. We must guarantee
+    ///  that all transactions started before this timestamp had committed. We can keep an active
+    /// transaction list in application to decide the minimal start timestamp of them.
     ///
     /// For each key, the last mutation record (unless it's a deletion) before `safepoint` is retained.
     ///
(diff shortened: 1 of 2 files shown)


src/transaction/transaction.rs
```diff
@@ -635,6 +635,11 @@ impl<PdC: PdClient> Transaction<PdC> {
         res
     }
 
+    /// Get the start timestamp of this transaction.
+    pub fn get_start_timestamp(&self) -> Timestamp {
```
andylokandy, line 639: Does this more close to the Rust idiom?
```suggestion
    pub fn start_timestamp(&self) -> Timestamp {
```

src/transaction/client.rs
```diff
@@ -183,7 +183,9 @@ impl Client {
 
     /// Request garbage collection (GC) of the TiKV cluster.
     ///
-    /// GC deletes MVCC records whose timestamp is lower than the given `safepoint`.
+    /// GC deletes MVCC records whose timestamp is lower than the given `safepoint`. We must guarantee
```
andylokandy, line 186: In fact, gc() is deprecated. The current GC mechanism used by TiDB 4.0 is described in the background section in [this RFC](https://github.com/andylokandy/rfcs/blob/gc/text/2021-04-23-builtin-gc-worker.md).

andylokandy: Oh, it is.

ekexium (approved):

PR #324, opened 2021-12-07
Add ttl for raw client

andylokandy:
Signed-off-by: andylokandy <andylokandy@hotmail.com>

- [x] Implementation
- [x] Manual test
- [ ] Add docs

closes #https://github.com/tikv/client-rust/pull/370


jiashiwen:
This is the feature  I need

src/raw/lowering.rs
```diff
@@ -21,21 +21,30 @@ pub fn new_raw_batch_get_request(
...
 }
 
 pub fn new_raw_batch_put_request(
-    pairs: impl Iterator<Item = KvPair>,
+    pairs_with_ttl: impl Iterator<Item = (KvPair, u64)>,
     cf: Option<ColumnFamily>,
     atomic: bool,
 ) -> kvrpcpb::RawBatchPutRequest {
-    requests::new_raw_batch_put_request(pairs.map(Into::into).collect(), cf, atomic)
+    let (pairs, ttls) = pairs_with_ttl.map(|(pair, ttl)| (pair.into(), ttl)).unzip();
```
ekexium, line 46: Why bother zipping and unzipping?

ekexium (commented):
Need some doc comments. Others LGTM.

src/raw/lowering.rs
andylokandy: Do you mean that the API may accept both `pairs` and `ttls`? If so `ttls` and `pairs` can be with different length and we've to add a new error type for it.

ekexium: The unzipping introduces performance overhead. IMO the method is supposed to be used internally so either returning a string error or just panicking seems fine to me.

supercmmetry:
I really really need this feature!!

Is there a reason why this PR was not merged / worked upon anymore?

andylokandy:
Oh, it was just forgotten.

iosmanthus (approved):

PR #329, opened 2022-01-18
pessimistic locks use MAX_TTL

### LINKED ISSUE #328: Transaction::lock_keys should wait instead of return a `Failed to resolve lock` Error.
Transaction::lock_keys should wait instead of return a `Failed to resolve lock` Error. As other language's standard library `lock` call will block current thread or coroutine. So Transaction::lock_keys should block until got the lock. And we can provide a `try_lock_keys` function that return immediately.

ekexium:
close #328 

Pessimistic locks now uses the default ttl(3s). The heartbeat mechanism by default works with the interval of 10s. It cannot keep pessimistic locks alive. Change it to MAX_TTL(20s).

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -48,6 +48,7 @@ impl Config {
     /// # use tikv_client::Config;
     /// let config = Config::default().with_security("root.ca", "internal.cert", "internal.key");
     /// ```
+    #[must_use]
     pub fn with_security(
         mut self,
         ca_path: impl Into<PathBuf>,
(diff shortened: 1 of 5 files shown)


blacktear23:
When will this PR can be merged?

ekexium:
@blacktear23 Awaiting at least 1 approval from reviewers with write access.
ping @andylokandy

andylokandy (approved):

Issue #313, opened 2021-08-06
Error handling for pessimistic locks


ekexium:
- If a pessimistic lock request fails, we should rollback locks
- The primary key should only be set if the request succeeds (on all keys).

pingyu:
Hi @ekexium , I would like to try to fix this issue, but I have two questions. Could you take a look and give me some advices ? Thanks~

1. The issue is leaded by TiKV server that breaks the loop when encounters an already locked key [here](https://github.com/tikv/tikv/blob/19f521896be77b6bf9ecc4eec3cecd3cd8004263/src/storage/txn/commands/acquire_pessimistic_lock.rs#L121-L123), with some other keys successful locked. Isn't it ?
2. The issue should be fixed by checking return of `transaction::Transaction::pessimistic_lock` (e.g. [here](https://github.com/tikv/client-rust/blob/196b06eb9da24de6cd5803dd2cb779685d72d969/src/transaction/transaction.rs#L418-L419)) and send a `new_pessimistic_rollback_request` request to TiKV server on failure. Shouldn't it ?

ekexium:
1. No. I think for most cases where the pessimistic lock requests fails, we should rollback the keys that has been locked. You may refer to how client-go does it. https://github.com/tikv/client-go/blob/50bafbb1e9b17948e5f2e71d4a3545f1a5211713/txnkv/transaction/txn.go#L675
2. Yeah it's basically what you said. I'd prefer to do it inside the `pessimistic_lock` function.

pingyu:
Got it, thanks~
I will have a try.

pingyu:
@ekexium I tried to reproduce the issue but failed. Could you please take a look ? Thanks~

(Full codes [here](https://github.com/pingyu/client-rust/blob/c0adaae666859d1cbdeb362a09e7b16292d61612/examples/pessimistic.rs#L78-L89), and [diff](https://github.com/tikv/client-rust/compare/master...pingyu:pessimistic-lock-rollback#diff-4b2df9a11b611a5fa38c4448059732805b093d837d73fa6b24acc2353a2a4f4fR64)).
I used a modified version of example `pessimistic.rs`. `key1` is locked by `txn1`.
Result of `txn5` is expected to be `Err(ResolveLockError)` but it's `Ok(())` actually.

```rust
   // txn4: partial successful locking
    let mut txn4 = client
        .begin_with_options(TransactionOptions::new_pessimistic().no_resolve_locks())
        .await
[...]

PR #332, opened 2022-02-05
txn: Error handling for pessimistic locks

pingyu:
Close: #313

# What's changed
Automatic rollback successful locks when `pessimistic_lock` is failed.

# How it is done
Add a new error `PessimisticLockError` holding successful locks return by TiKV server.
On failure, `pessimistic_lock` acquires the successful keys from `PessimisticLockError` and do the rollback.

# Tests
- [x] Manual
- [x] Unit test
- [x] Integrated test

### DIFF
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -16,6 +16,7 @@ use crate::{
     stats::tikv_stats,
     store::RegionStore,
     transaction::{resolve_locks, HasLocks},
+    util::iter::FlatMapOkIterExt,
     Error, Result,
 };
 
@@ -63,6 +64,7 @@ pub struct RetryableMultiRegion<P: Plan, PdC: PdClient> {
     pub(super) inner: P,
     pub pd_client: Arc<PdC>,
     pub backoff: Backoff,
+    pub preserve_results: bool,
 }
 
 impl<P: Plan + Shardable, PdC: PdClient> RetryableMultiRegion<P, PdC>
@@ -76,6 +78,7 @@ where
         current_plan: P,
         backoff: Backoff,
         permits: Arc<Semaphore>,
+        preserve_results: bool,
     ) -> Result<<Self as Plan>::Result> {
         let shards = current_plan.shards(&pd_client).collect::<Vec<_>>().await;
         let mut handles = Vec::new();
@@ -89,16 +92,27 @@ where
                 region_store,
                 backoff.clone(),
                 permits.clone(),
+                preserve_results,
             ));
             handles.push(handle);
         }
-        Ok(try_join_all(handles)
-            .await?
-            .into_iter()
-            .collect::<Result<Vec<_>>>()?
-            .into_iter()
-            .flatten()
-            .collect())
+
+        let iter = try_join_all(handles).await?.into_iter();
+        if preserve_results {
+            Ok(iter
+                .flat_map_ok(|x| x.into_iter())
+                .map(|x| match x {
+                    Ok(r) => r,
+                    Err(e) => Err(e),
+                })
+                .collect())
+        } else {
+            Ok(iter
+                .collect::<Result<Vec<_>>>()?
+                .into_iter()
+                .flatten()
+                .collect())
+        }
     }
 
     #[async_recursion]
@@ -108,6 +122,7 @@ where
         region_store: RegionStore,
         mut backoff: Backoff,
         permits: Arc<Semaphore>,
+        preserve_results: bool,
     ) -> Result<<Self as Plan>::Result> {
         // limit concurrent requests
         let permit = permits.acquire().await.unwrap();
@@ -125,7 +140,8 @@ where
                     if !region_error_resolved {
                         futures_timer::Delay::new(duration).await;
                     }
-                    Self::single_plan_handler(pd_client, plan, backoff, permits).await
+                    Self::single_plan_handler(pd_client, plan, backoff, permits, preserve_results)
+                        .await
                 }
                 None => Err(Error::RegionError(e)),
             }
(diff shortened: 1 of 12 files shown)


src/transaction/requests.rs
```diff
@@ -379,35 +380,51 @@ impl Merge<ResponseWithShard<kvrpcpb::PessimisticLockResponse, Vec<kvrpcpb::Muta
...
-                    assert_eq!(kvpairs.len(), not_founds.len());
-                    Either::Right(kvpairs.zip(not_founds).filter_map(|(kvpair, not_found)| {
-                        if not_found {
-                            None
-                        } else {
-                            Some(kvpair)
-                        }
-                    }))
-                }
+        let (success, mut errors): (Vec<_>, Vec<_>) = input.into_iter().partition(Result::is_ok);
```
ekexium, line 383: It's unncessary to always partition which introduces overhead of memory copy. We can check if there is error first.

src/request/plan.rs
```diff
@@ -63,6 +64,7 @@ pub struct RetryableMultiRegion<P: Plan, PdC: PdClient> {
     pub(super) inner: P,
     pub pd_client: Arc<PdC>,
     pub backoff: Backoff,
+    pub preserve_results: bool,
```
ekexium, line 67: I'm thinking whether there is a better name, since "result" is too generic and can be ambiguous. Or at least we can add some comment to explain.

src/transaction/transaction.rs
```diff
@@ -744,27 +744,60 @@ impl<PdC: PdClient> Transaction<PdC> {
...
 
-        for key in keys {
-            self.buffer.lock(key.key());
+            pairs
         }
+    }
 
-        pairs
+    /// Rollback pessimistic lock
+    async fn pessimistic_lock_rollback(
```
ekexium, line 786: Had better print some logs here.

tests/integration_tests.rs
```diff
@@ -674,6 +674,38 @@ async fn txn_lock_keys() -> Result<()> {
...
 }
 
+#[tokio::test]
+#[serial]
+async fn txn_lock_keys_error_handle() -> Result<()> {
+    init().await?;
+    let client = TransactionClient::new_with_config(pd_addrs(), Default::default(), None).await?;
+    let mut rng = thread_rng();
+
+    let k = gen_u32_keys(4, &mut rng)
```
ekexium, line 684: In this test I think it would be better not to use rand numbers. We can manually choose 4 keys that should locate in different regions.

src/transaction/transaction.rs
ekexium: I think we need to appropriately change the buffer state. In the current PR it doesn't affect at all, but I'm afriad the function might be misused. Or we can add a comment to explicitly warn such behavior.

ekexium (commented):
Great job! I just have a few suggestions.

And I noticed we didn't set the `MULTI_REGION` correctly in Github actions. Please do me a favor to fix it :). Simply change the last command in ci.yml to `MULTI_REGION=1 make integration-test`

pingyu:
@ekexium 
Comments addressed. I'm sorry for the late. PTAL, thanks~

src/transaction/transaction.rs
pingyu: > Had better print some logs here.

En... I print a debug msg "rollback pessimistic lock" at the beginning.
What more logs do we need ? *(since there is also a simple debug log at the beginning of other methods)*

andylokandy: @ekexium Do we allow the user to retry pessimistic lock? If we do, we should not eagerly rollback the successfully locked keys.

ekexium: > @ekexium Do we allow the user to retry pessimistic lock? If we do, we should not eagerly rollback the successfully locked keys.

Users can retry, but we cannot make assumptions on that. The lock request may already have retried, subject to the backoff option.

ekexium: > 



> > Had better print some logs here.
> 
> En... I print a debug msg "rollback pessimistic lock" at the beginning. What more logs do we need ? _(since there is also a simple debug log at the beginning of other methods)_

I didn't see it 🤦 . The logging is a mess for now. Let's just keep it as is in this PR.

andylokandy: > Users can retry, but we cannot make assumptions on that. The lock request may already have retried, subject to the backoff option.

Let me explain my point again. If we allow the user to retry the pessimistic lock when the client returns errors, the client should remember which keys are locked, but not eagerly roll back them. Besides, the client should roll back the locked key when `client.rollback()` is called, or panic if get dropped without rolling back or committing.

ekexium: I think the major drawback of your idea is that if the user doesn't retry the same set of keys, then the locks will remain and may block other txns for some time.

If users want to retry the same set of keys, they can use a different backoff strategy (and we may consider enhance the retry of pessimistic locks to use a newer for_update_ts).

Reference: client-go immediately rollback the successful pessimistic locks, but I think part of the reason is to cooperate with TiDB, client-rust doesn't have to bahave in the same way.

pingyu: I agree with @ekexium . I also think that it would be better to provide an "atomic" method to developers, which will do nothing when return error.

pingyu:
The integration test case `txn_optimistic_heartbeat` is stably failed. Let me see see.

pingyu:
> The integration test case `txn_optimistic_heartbeat` is stably failed. Let me see see.

The [error](https://github.com/tikv/client-rust/runs/5232223077?check_suite_focus=true) is raised from `clear_tikv` in first `tests::common::init()`, and should be caused by the 500 ms timeout of `DEFAULT_REGION_BACKOFF` is not enough for CI environment.

I fixed it by prolonging the max delay to 10 seconds.

andylokandy (approved):

andylokandy:
@ekexium PTAL

ekexium:
LGTM. We should consider support optional backoffs for all raw requests.

PR #338, opened 2022-03-16
Add gRPC local subchannel support for tikv client.

blacktear23:
As default configuration gRPC will reuse TCP connections to each TiKV server what ever how many TiKV client instance created. This PR will enable `grpc. use_local_subchannel_pool` settings to ensure every TiKV client will create it self owned TCP connections to TiKV (or PD). 

This will help to use more CPU when heavy load on gRPC calls. And as default settings it will make user confused when create many TiKV client but OS shows only one TCP connection to each TiKV server.

### DIFF
--- a/tikv-client-common/src/security.rs
+++ b/tikv-client-common/src/security.rs
@@ -4,6 +4,7 @@ use crate::Result;
 use grpcio::{Channel, ChannelBuilder, ChannelCredentialsBuilder, Environment};
 use regex::Regex;
 use std::{
+    ffi::CString,
     fs::File,
     io::Read,
     path::{Path, PathBuf},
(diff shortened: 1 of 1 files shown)


tikv-client-common/src/security.rs
```diff
@@ -76,9 +77,12 @@ impl SecurityManager {
 
         let addr = SCHEME_REG.replace(addr, "");
 
+        // Set grpc.use_local_subchannel_pool will ensure every TiKV client instance can has it's own
+        // connection to TiKV server.
         let cb = ChannelBuilder::new(env)
             .keepalive_time(Duration::from_secs(10))
-            .keepalive_timeout(Duration::from_secs(3));
+            .keepalive_timeout(Duration::from_secs(3))
+            .raw_cfg_int(CString::new("grpc.use_local_subchannel_pool").unwrap(), 1);
```
ekexium, line 85: `raw_cfg_int` is not intended to be used directly according to its doc.I think it would be better to add a method grpc-rs to set it.

ekexium (commented):
Can we add a test for this change?

tikv-client-common/src/security.rs
blacktear23: OK, I will create a PR for grpc-rs first. But this is the easy way to set the configuration at rust-client side.

blacktear23:
@ekexium  https://github.com/tikv/grpc-rs/pull/565 is created.

yongman:
@blacktear23 any update to this pr?
Or I have made an alternative pr https://github.com/tikv/client-rust/pull/361 for this.

blacktear23:
@yongman nothing changed, we can go ahead with your pr.

PR #345, opened 2022-04-07
Fix potential stuck in region cache

### LINKED ISSUE #344: In `RegionCache`, concurrent calls to `get_region_by_id()` may get stuck?
TL;DR: The tokio task waiting for others to finish retrieving a cache entry may block forever, because it registers its interest __after__ the other task notifies the waiters.

---
Refering to the following code, a task (call it task A) waiting for a concurrent task (call it task B) to finish retrieving a cache entry will:
1. clone a `Notify` from `on_my_way_id` (line 102)
2. release the region cache lock (line 103)
3. call the `notified()` method on the `Notify`, __which registers interest on this `Notify`__ (line 105)
4. Wait for the concurrent cache retrieving to finish by awaiting on the future created by `notified()` (line 106)

https://github.com/tikv/client-rust/blob/5714b2f2637f0e6edd4be1ad7ce5c0bb363e9504/src/region_cache.rs#L101-L106

When task B finishes retrieving the cache entry, it will do the following to notify other tasks __that has registered interest__:

https://github.com/tikv/client-rust/blob/5714b2f2637f0e6edd4be1ad7ce5c0bb363e9504/src/region_cache.rs#L145-L150

If this happens between task A's step 2 and step 3 (line 103 and line 105), task A will never get notified.

stamp711:
Fixes #344 by registering interest prior to releasing the region cache lock.

### DIFF
--- a/src/region_cache.rs
+++ b/src/region_cache.rs
@@ -100,10 +100,11 @@ impl<C: RetryClientTrait> RegionCache<C> {
 
             // check concurrent requests
             let notify = region_cache_guard.on_my_way_id.get(&id).cloned();
+            let notified = notify.as_ref().map(|notify| notify.notified());
             drop(region_cache_guard);
 
-            if let Some(n) = notify {
-                n.notified().await;
+            if let Some(n) = notified {
+                n.await;
                 continue;
             } else {
                 return self.read_through_region_by_id(id).await;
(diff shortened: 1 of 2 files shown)


ekexium (approved):
Thank you! Could you fix the clippy error please?

stamp711:
It seems that the nightly version of clippy now enables the [crate_in_macro_def](https://rust-lang.github.io/rust-clippy/master/index.html#crate_in_macro_def) lint by default.
I've included a commit that fixes those warnings.

PR #347, opened 2022-05-23
Fix region cache update error when handle not leader

yongman:
The `region cache map` may be updated by concurrency when do `handle_region_error`, this may result in a lot of `EntrynotFoundInRegionCache` error. 
We should check if region entry exists, ignore update leader if entry is none. The entry cache will be filled up in the next get region reqeust.

### DIFF
--- a/src/region_cache.rs
+++ b/src/region_cache.rs
@@ -207,9 +207,7 @@ impl<C: RetryClientTrait> RegionCache<C> {
         leader: metapb::Peer,
     ) -> Result<()> {
         let mut cache = self.region_cache.write().await;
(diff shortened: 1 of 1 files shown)


yongman:
@ekexium The ci check failure has no relation with this pr, can we fix it in another pr?

yongman:
CI clippy failure is the new feature and have an issue not resolved https://github.com/rust-lang/rust-clippy/issues/8867 in nightly version.

iosmanthus:
Rest LGTM, please fix the CI and clippy.

iosmanthus (approved):
LGTM

PR #348, opened 2022-05-26
Invalidate region cache in key out of region error

yongman:
I check the protobuf defination and code, the error happened here https://github.com/tikv/tikv/blob/v6.0.0/components/raftstore/src/store/region_snapshot.rs#L228

and set response to key error here
https://github.com/tikv/tikv/blob/v6.0.0/src/storage/errors.rs#L409

In client side, may be this region in cache is stale, we just need to invalidate this region in cache and retry.

Or this type error should be treated as some `RegionError` in TiKV server?

The error message details like
```
KeyError { locked: None, retryable: "", abort: "Error(Txn(Error(Mvcc(Error(Kv(Error(Other(\"[components/tikv_kv/src/[raftstore_impls.rs:75](http://raftstore_impls.rs:75/)]: \\\"[components/raftstore/src/store/[region_snapshot.rs:228](http://region_snapshot.rs:228/)]: Key 785F000000000000FF00015F4D5F613334FF3530303030333538FF0000000000000000F7 is out of [region 514] [785F000000000000FF00015F4D5F613334FF3530303032383139FF0000000000000000F7, 785F000000000000FF00015F4D5F613337FF3530313332320000FD)\\\"\"))))))))", conflict: None, already_exist: None, deadlock: None, commit_ts_expired: None, txn_not_found: None, commit_ts_too_large: None, assertion_failed: None }
```

### DIFF
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -2,6 +2,9 @@
 
 use std::{marker::PhantomData, sync::Arc};
 
+use lazy_static::lazy_static;
+use regex::Regex;
+
 use async_recursion::async_recursion;
 use async_trait::async_trait;
 use futures::{future::try_join_all, prelude::*};
@@ -19,6 +22,11 @@ use crate::{
     util::iter::FlatMapOkIterExt,
     Error, Result,
 };
+
+lazy_static! {
+    static ref KEY_NOT_IN_REGION_ERR_REG: Regex =
+        Regex::new(r"Key \w+ is out of \[region \d+\]").unwrap();
+}
 
 /// A plan for how to execute a request. A user builds up a plan with various
 /// options, then exectutes it.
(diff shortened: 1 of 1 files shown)


yongman:
@ekexium PTAL

ekexium:
Matching error strings is a bad idea. I think we should try to fix TiKV.

yongman:
> Matching error strings is a bad idea. I think we should try to fix TiKV.

I agree, we should fix it in TiKV new version.
But is there an elegant way to fix rust client to handle this error for region cache update if client running with old TiKV versions ?

ekexium:
client-rust hasn't reached its 1.0, so legacy defects are usually ignored.

yongman:
> client-rust hasn't reached its 1.0, so legacy defects are usually ignored.

I have opened an issue for TiKV https://github.com/tikv/tikv/issues/12665 
I will keep following this up.

PR #353, opened 2022-06-20
Introduce KeySpace interface for API v2.

iosmanthus:
Signed-off-by: iosmanthus <myosmanthustree@gmail.com>

This pull request is still working in progress, and the interfaces for request encoding are stable. However, lots of the request structs need to implement with these interfaces.

The core interfaces to encode the request are:

```rust
/// Abstracts any request sent to a TiKV server.
#[async_trait]
pub trait KvRequest<C>: Request + Sized + Clone + Sync + Send + 'static {
    /// The expected response to the request.
    type Response: HasKeyErrors + HasLocks + Clone + Send + 'static;
    fn encode_request(&self, _codec: &C) -> Cow<Self>;
    fn decode_response(&self, _codec: &C, _resp: Self::Response) -> crate::Result<Self::Response>;
}
```

When the `C` is bound to `RequestCodec`, the requests are able to be encoded:

```rust
pub trait RequestCodec: Sized + Clone + Sync + Send + 'static {
    fn encode_key(&self, key: Vec<u8>) -> Vec<u8> { key }
    fn decode_key(&self, key: Vec<u8>) -> Result<Vec<u8>> { Ok(key) }
    fn encode_range(&self, start: Vec<u8>, end: Vec<u8>) -> (Vec<u8>, Vec<u8>) { (start, end) }
    fn encode_pd_query(&self, key: Vec<u8>) -> Vec<u8> { key }
    fn decode_region(&self, region: Region) -> Result<Region> { Ok(region) }
}

```

TODO:
- [x] Compile pass for API V1
- [x] Fix test for API v1
- [x] Impl V2 codec
- [ ] Write tests for V2 codec.

### DIFF
--- a/src/transaction/client.rs
+++ b/src/transaction/client.rs
@@ -1,6 +1,6 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
-use std::{future::Future, marker::PhantomData, mem, sync::Arc};
+use std::{future::Future, mem, sync::Arc};
 
 use slog::Logger;
 
@@ -41,7 +41,6 @@ const SCAN_LOCK_BATCH_SIZE: u32 = 1024;
 pub struct Client<C: TxnCodec> {
     pd: Arc<PdRpcClient<C>>,
     logger: Logger,
-    _phantom: PhantomData<C>,
 }
 
 impl<C: TxnCodec> Clone for Client<C> {
@@ -49,7 +48,6 @@ impl<C: TxnCodec> Clone for Client<C> {
         Self {
             pd: self.pd.clone(),
             logger: self.logger.clone(),
-            _phantom: PhantomData,
         }
     }
 }
@@ -184,11 +182,7 @@ where
         let ClientContext { pd, logger } =
             ClientContext::new(pd_endpoints, config, codec_factory, optional_logger).await?;
 
-        Ok(Client {
-            pd,
-            logger,
-            _phantom: PhantomData,
-        })
+        Ok(Client { pd, logger })
     }
 
     /// Creates a new optimistic [`Transaction`].


src/pd/client.rs
```diff
@@ -204,22 +209,27 @@ pub trait PdClient: Send + Sync + 'static {
     async fn update_leader(&self, ver_id: RegionVerId, leader: metapb::Peer) -> Result<()>;
 
     async fn invalidate_region_cache(&self, ver_id: RegionVerId);
+
+    fn get_request_codec(&self) -> Self::RequestCodec;
```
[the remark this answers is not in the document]
iosmanthus: Since the only use of `RequestCodec` is in the construction of `Dispatch`, and it requires a move, I think `Self::RequestCodec` is more direct.

examples/pessimistic.rs
```diff
@@ -20,7 +23,7 @@ async fn main() {
     };
 
     // init
-    let client = Client::new_with_config(args.pd, config, None)
+    let client = Client::new_with_config(args.pd, config, TxnApiV1, None)
```
sunxiaoguang, line 26: `Client` itself can tell we are using TxnKV already, can we hide things like TxnApiV1, TxnApiV2, RawApiV1, RawApiV2. Just let user chose the appropriate client and tell he wants to use ApiV1 or ApiV2.

src/lib.rs
```diff
@@ -66,8 +66,9 @@
 //! # use tikv_client::{RawClient, Result};
 //! # use futures::prelude::*;
 //! # fn main() -> Result<()> {
-//! # futures::executor::block_on(async {
-//! let client = RawClient::new(vec!["127.0.0.1:2379"], None).await?;
+//! # use tikv_client::request::request_codec::RawApiV1;
+//! futures::executor::block_on(async {
+//! let client = RawClient::new(vec!["127.0.0.1:2379"], RawApiV1, None).await?;
```
iosmanthus, line 71: This might break the original API, we should add a default implementation for API V1.

examples/pessimistic.rs
iosmanthus: We may reexport them in their own module, like: `raw::ApiV1` , `txn::ApiV1`

src/request/mod.rs
```diff
@@ -22,12 +24,103 @@ pub mod plan;
...
+                *resp.mut_pairs() = codec.decode_pairs(resp.take_pairs())?;
+
+                Ok(resp)
+            }
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! impl_kv_request_for_scan_op {
```
iosmanthus, line 91: These macro rules have the same patterns which is to take some field out and then encode or decode it. We should extract them into one rule.

src/request/codec.rs
```diff
@@ -0,0 +1,343 @@
...
+        let mut encoded = Vec::with_capacity(key.len() + KEYSPACE_PREFIX_LEN);
+        let prefix: Prefix = self.keyspace.into();
+
+        encoded.extend_from_slice(&prefix);
+        encoded.append(&mut key);
+        encoded
+    }
+
+    fn decode_key(&self, key: &mut Vec<u8>) -> Result<()> {
+        let prefix: Prefix = self.keyspace.into();
```
sunxiaoguang, line 285: There is an allocation every time encode and decode is called. We can make Keyspace store a final prefix internally and get reference of Prefix instead.

iosmanthus: There is no heap allocation happening here since the `Prefix` is a type alias of `[u8; 4]`. The reference size in the stack should be the same as `Prefix`.

src/request/codec.rs
```diff
@@ -0,0 +1,452 @@
...
+    }
+}
+
+impl DerefMut for KeySpaceId {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl KeySpaceId {
```
iosmanthus, line 227: @sunxiaoguang  PTAL. This is the logic for constructing keyspace from the name.

adamf:
I am very excited for this!

PR #354, opened 2022-07-04
Add scan_reverse and scan_keys_reverse support for txnkv

yongman:


src/transaction/requests.rs
```diff
@@ -125,13 +125,20 @@ pub fn new_scan_request(
     timestamp: u64,
     limit: u32,
     key_only: bool,
+    reverse: bool,
 ) -> kvrpcpb::ScanRequest {
     let mut req = kvrpcpb::ScanRequest::default();
-    req.set_start_key(start_key);
-    req.set_end_key(end_key);
+    if !reverse {
+        req.set_start_key(start_key);
```
iosmanthus, line 132: This violates the request spec in TiKV, we assume the start key is always no larger than the end key.

iosmanthus: You can check the client-go implementation: https://github.com/tikv/client-go/blob/bb026bdad8f34604cc01fdafbaf4063729583a44/txnkv/txnsnapshot/scan.go#L194

yongman: @iosmanthus I check the proto definition in kvrpcpb.proto https://github.com/tikv/client-rust/blob/master/tikv-client-proto/proto/kvrpcpb.proto#L57 

```
    // For compatibility, when scanning forward, the range to scan is [start_key, end_key), where start_key < end_key;
    // and when scanning backward, it scans [end_key, start_key) in descending order, where end_key < start_key.
```

The start_key and end_key is define as `bound range` , (start_key < end_key is guaranteed).

If reverse scan request be created, we use the range [req.end_key, req.start_key) to create scan request which should be initialized use 

```
        req.set_start_key(end_key);
        req.set_end_key(start_key);
```
which  guarantee the req.end_key<req.start_key.

yongman: And the integration test works as expected.

src/transaction/buffer.rs
```diff
@@ -158,7 +159,12 @@ impl Buffer {
             .into_iter()
             .map(|(k, v)| KvPair::new(k, v))
             .collect::<Vec<_>>();
-        res.sort_by_cached_key(|x| x.key().clone());
+
+        if reverse {
```
iosmanthus, line 163: I think the integration tests don't break because of this sorting. I add some debug logs and find the origin result from the TiKV is in asceeding order. We are not doing a reverse scan in server.

iosmanthus, line 163: Would you mind adding a TODO here? Since we can use BTreeMap to eliminate this sort.

tests/integration_tests.rs
```diff
@@ -883,6 +883,42 @@ async fn txn_scan() -> Result<()> {
...
+async fn txn_scan_reverse() -> Result<()> {
+    init().await?;
+    let client = TransactionClient::new_with_config(pd_addrs(), Default::default(), None).await?;
+
+    let k1 = b"a1".to_vec();
+    let k2 = b"a2".to_vec();
+    let v1 = b"b1".to_vec();
+    let v2 = b"b2".to_vec();
+
+    let _reverse_resp = vec![
```
iosmanthus, line 897: This variable is used, so we can ignore the underline prefix.

src/transaction/buffer.rs
yongman: OK, I will add a TODO note here

tests/integration_tests.rs
```diff
@@ -883,6 +883,42 @@ async fn txn_scan() -> Result<()> {
...
+    // pessimistic
+    let option = TransactionOptions::new_pessimistic().drop_check(tikv_client::CheckLevel::Warn);
+    let mut t = client.begin_with_options(option.clone()).await?;
+    t.put(k1.clone(), v1).await?;
+    t.put(k2.clone(), v2).await?;
+    t.commit().await?;
+
+    let mut t2 = client.begin_with_options(option).await?;
+    let bound_range: BoundRange = (k1..=k2).into();
+    let _resp = t2
```
iosmanthus, line 911: Ditto.

iosmanthus:
Rest LGTM.

iosmanthus (approved):

Issue #356, opened 2022-07-13
Need Prerequisites for compiling, avoiding failure or error in building


yongpan0709:
It needs more details for gcc, cmake, Rust version before contributing to client-rust. There are some errors during compiling. For example, 
1. error: failed to parse manifest at `/root/.cargo/registry/src/github.com-1ecc6299db9ec823/indexmap-1.9.1/Cargo.toml`
Caused by:
  failed to parse the `edition` key
Caused by:
  supported edition values are `2015` or `2018`, but `2021` is unknown

2. error: package `hashbrown v0.12.1` cannot be built because it requires rustc 1.56.1 or newer, while the currently active rustc version is 1.56.0

3. running: "cmake" "--build" "." "--target" "grpc" "--config" "Debug" "--parallel" "8"
  --- stderr
  CMake Warning at cmake/protobuf.cmake:51 (message):
    gRPC_PROTOBUF_PROVIDER is "module" but PROTOBUF_ROOT_DIR is wrong
  Call Stack (most recent call first):
    CMakeLists.txt:254 (include)
  CMake Warning:
    Manually-specified variables were not used by the project:
      CMAKE_ASM_COMPILER
      CMAKE_ASM_FLAGS
  Unknown argument --parallel
  Unknown argument 8

PR #357, opened 2022-07-13
Enhance: provide more details about compiling, 

yongpan0709:
Compilation of Client-Rust is supported by some dependencies. Before compiling, we need more details about the minimum version of different dependencies:
image

More details, please refer to https://pingcap.feishu.cn/docx/doxcnlZkOyF0YoubchAIeyFksJe

Close: #356

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,8 @@ jobs:
           profile: minimal
           toolchain: nightly
           override: true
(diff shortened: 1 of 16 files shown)


pingyu:
The feishu doc is great. But it's not available except PingCAPers.

How about convert to markdown and also add to the repo ? To describe why we have some version requirement.

Rest LGTM~

PR #360, opened 2022-07-21
Optimize key_exists performance

yongman:
In TiKV `scan` is a more costly operation than `get`, and `scan` can not take advantage of bloom filter, so use `get` will be more efficient.


ekexium (commented):
Could you add a test for it?

yongman:
@ekexium test case added, PTAL

ekexium:
And you need to fix cilppy

ekexium (approved):

PR #368, opened 2022-09-07
Add new stream scan APIs for transaction

yongman:
Add new `scan_stream`, `scan_keys_stream`, `scan_reverse_stream`, `scan_keys_reverse_stream` APIs for transaction scan.
Use these APIs, scan request will be invoke request to one region at a time in a configurable batch size, which is more reasonable when user scan range is wide-range, such as all key space.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ futures-timer = "3.0"
 grpcio = { version = "0.10", features = [ "prost-codec", "openssl-vendored" ], default-features = false }
 lazy_static = "1"
 log = "0.4"
+ordered-stream = "0.1.1"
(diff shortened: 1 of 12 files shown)


yongman:
@ekexium @iosmanthus PTAL

andylokandy:
IIRC, you can limit the scan size in batch using the current API like:

```rust
let mut start_key = Key::from(vec![]);
let end_key = Key::from(vec![]);
let limit = 100;

let mut buffer = vec![];
let mut last_len = 0;

loop {
    buffer.extend(client.scan((start_key, end_key), limit));
    if buffer.len() - last_len < limit {
        break;
    }
    last_len = buffer.len();
    start_key = buffer.last().unwrap().0.clone();
    start_key.push_zero();
}

PR #378, opened 2022-12-08
txn: Resolve locks for async commit (region level)

pingyu:
### Changes

- Add resolve locks (of region level) for async commit in GC scenario.

Note:
1. Resolve locks for async commit of request level is NOT done in this PR.
2. This PR mainly refer to [`LockResolver`](https://github.com/tikv/client-go/blob/v2.0.3/txnkv/txnlock/lock_resolver.go#L225) of `client-go`.
3. The following will be done in next PRs:
  a. More tests for key/region error scenarios.
  b. Polish backoffs.
  c. Benchmark and performance improvnment on large data.

### Tests
- Integration tests

### DIFF
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -72,6 +72,12 @@ impl AsRef<Key> for kvrpcpb::Mutation {
     }
 }
 
+impl AsRef<Key> for kvrpcpb::LockInfo {
+    fn as_ref(&self) -> &Key {
+        self.primary_lock.as_ref()
+    }
+}
+
 impl Key {
     /// The empty key.
     pub const EMPTY: Self = Key(Vec::new());
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -72,12 +72,6 @@ impl AsRef<Key> for kvrpcpb::Mutation {
     }
 }
 
-impl AsRef<Key> for kvrpcpb::LockInfo {
-    fn as_ref(&self) -> &Key {
-        self.primary_lock.as_ref()
-    }
-}
-
 impl Key {
     /// The empty key.
     pub const EMPTY: Self = Key(Vec::new());
--- a/src/region.rs
+++ b/src/region.rs
@@ -2,7 +2,6 @@
 
 use crate::{Error, Key, Result};
 use derive_new::new;
-use std::ops::Bound;
 use tikv_client_proto::{kvrpcpb, metapb};
 
 /// The ID of a region
@@ -38,20 +37,6 @@ impl RegionWithLeader {
         let start_key = self.region.get_start_key();
         let end_key = self.region.get_end_key();
         key >= start_key && (key < end_key || end_key.is_empty())
-    }
-
-    pub fn contains_start_bound(&self, bound: Bound<&Key>) -> bool {
-        let start_key = self.region.get_start_key();
-        let end_key = self.region.get_end_key();
-
-        match bound {
-            Bound::Included(key) => self.contains(key),
-            Bound::Excluded(key) => {
-                let key: &[u8] = key.into();
-                key >= start_key && (key <= end_key || end_key.is_empty())
-            }
-            Bound::Unbounded => start_key.is_empty(),
-        }
     }
 
     pub fn context(&self) -> Result<kvrpcpb::Context> {
--- a/src/region.rs
+++ b/src/region.rs
@@ -2,6 +2,7 @@
 
 use crate::{Error, Key, Result};
 use derive_new::new;
+use std::ops::Bound;
 use tikv_client_proto::{kvrpcpb, metapb};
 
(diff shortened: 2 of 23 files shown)


sticnarf:
> * (CI have some issues and would be fixed in another PR)

Would you help fix the CI first? It's not a good idea to merge a PR with CI failing.

pingyu:
> > * (CI have some issues and would be fixed in another PR)
> 
> Would you help fix the CI first? It's not a good idea to merge a PR with CI failing.

OK~

src/kv/key.rs
```diff
@@ -72,6 +72,12 @@ impl AsRef<Key> for kvrpcpb::Mutation {
     }
 }
 
+impl AsRef<Key> for kvrpcpb::LockInfo {
```
sticnarf, line 75: Can we avoid this impl? It makes me feel a bit uncomfortable because the corresponding key is not the primary key.

src/region.rs
```diff
@@ -39,6 +40,20 @@ impl RegionWithLeader {
...
 
+    pub fn contains_start_bound(&self, bound: Bound<&Key>) -> bool {
+        let start_key = self.region.get_start_key();
+        let end_key = self.region.get_end_key();
+
+        match bound {
+            Bound::Included(key) => self.contains(key),
+            Bound::Excluded(key) => {
+                let key: &[u8] = key.into();
+                key >= start_key && (key <= end_key || end_key.is_empty())
```
sticnarf, line 51: I don't get the difference between `contains` and `contains_start_bound`.

```
key space: ---------------------
        region: [_____)
               range: (_______
```

The end key of the region is excluded. If the given start bound is also excluded, when they equal, the region doesn't seem to _contain_ any part of the given range (so it seems to me that we should not say it contains the start bound).

src/transaction/transaction.rs
```diff
@@ -1266,8 +1270,40 @@ impl<PdC: PdClient> Committer<PdC> {
         debug!(self.logger, "committing secondary");
         let mutations_len = self.mutations.len();
         let primary_only = mutations_len == 1;
+        #[cfg(not(feature = "integration-tests"))]
         let mutations = self.mutations.into_iter();
 
+        // #[cfg(feature = "integration-tests")]
+        // let mut mutations = self.mutations.into_iter().take(mutations_len).cloned();
```
sticnarf, line 1277: Should remove

src/transaction/client.rs
```diff
@@ -235,18 +237,62 @@ impl Client {
...
+            .clone()
+            .update_safepoint(safepoint.version())
+            .await?;
+        if !res {
+            info!(self.logger, "new safepoint != user-specified safepoint");
+        }
+        Ok(res)
+    }
+
+    pub async fn cleanup_locks(
```
sticnarf, line 259: What is the purpose of providing a public function `cleanup_locks`?

src/transaction/lock.rs
```diff
@@ -132,6 +142,277 @@ async fn resolve_lock_with_retry(
...
+#[derive(Default, Clone)]
+pub struct ResolveLocksContext {
+    // Record the status of each transaction.
+    pub(crate) resolved: Arc<RwLock<HashMap<u64, Arc<TransactionStatus>>>>,
+    pub(crate) clean_regions: Arc<RwLock<HashMap<u64, HashSet<RegionVerId>>>>,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct ResolveLocksOptions {
+    pub async_commit_only: bool,
```
sticnarf, line 154: Why providing this option?

pingyu:
> > * (CI have some issues and would be fixed in another PR)
> 
> Would you help fix the CI first? It's not a good idea to merge a PR with CI failing.

CI is fixed by https://github.com/tikv/client-rust/pull/381, PTAL~

src/transaction/lock.rs
coocood: It is used for cloud-engine native restore.
We only need to resolve async commit locks, other locks can be resolved later by tidb.

src/kv/key.rs
pingyu: I'm sorry that I forgot to remove this block.
They are useless and just for debug.

src/region.rs
pingyu: Emm... I change to `(Vec<u8>, Vec<u8>)` as it seems that `BoundRange` is not necessary.
`contains_start_bound` is useless now and I remove it.

pingyu:
@sticnarf 
All comments are addressed or replied. PTAL, thanks~

sticnarf (approved):
LGTM but I suggest changing the title a bit.

The lock resolving for common requests still doesn't support async commit. Only the region-level cleanup supports resolving async commit locks.

PR #386, opened 2023-01-03
Resolve lock in range

haojinming:
Add range in `ResolveLocksOptions`

### DIFF
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -450,6 +450,12 @@ pub struct CleanupLocksResult {
     // TODO: pub resolved_locks: usize,
 }
 
+impl CleanupLocksResult {
+    fn has_error(&self) -> bool {
+        self.key_error.is_some() || self.region_error.is_some()
+    }
+}
+
 impl Clone for CleanupLocksResult {
     fn clone(&self) -> Self {
         Self {
@@ -478,8 +484,14 @@ impl Merge<CleanupLocksResult> for Collect {
         input
             .into_iter()
             .fold(Ok(CleanupLocksResult::default()), |acc, x| {
+                let new_ret = x?;
+                let new_lock_cnt = if new_ret.has_error() {
+                    0
+                } else {
+                    new_ret.meet_locks
+                };
                 Ok(CleanupLocksResult {
-                    meet_locks: acc?.meet_locks + x?.meet_locks,
+                    meet_locks: acc?.meet_locks + new_lock_cnt,
                     ..Default::default()
                 })
             })
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -479,7 +479,7 @@ impl Merge<CleanupLocksResult> for Collect {
             .into_iter()
             .fold(Ok(CleanupLocksResult::default()), |acc, x| {
                 Ok(CleanupLocksResult {
-                    meet_locks: acc.unwrap().meet_locks + x?.meet_locks,
+                    meet_locks: acc?.meet_locks + x?.meet_locks,
                     ..Default::default()
                 })
             })
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -446,20 +446,13 @@ where
 pub struct CleanupLocksResult {
     pub region_error: Option<errorpb::Error>,
     pub key_error: Option<Vec<Error>>,
-    pub meet_locks: usize,
-    // TODO: pub resolved_locks: usize,
-}
-
-impl CleanupLocksResult {
-    fn has_error(&self) -> bool {
-        self.key_error.is_some() || self.region_error.is_some()
-    }
+    pub resolved_locks: usize,
 }
 
 impl Clone for CleanupLocksResult {
     fn clone(&self) -> Self {
         Self {
-            meet_locks: self.meet_locks,
+            resolved_locks: self.resolved_locks,
             ..Default::default() // Ignore errors, which should be extracted by `extract_error()`.
         }
     }
@@ -484,14 +477,8 @@ impl Merge<CleanupLocksResult> for Collect {
         input
             .into_iter()
             .fold(Ok(CleanupLocksResult::default()), |acc, x| {
-                let new_ret = x?;
-                let new_lock_cnt = if new_ret.has_error() {
-                    0
-                } else {
-                    new_ret.meet_locks
-                };
                 Ok(CleanupLocksResult {
-                    meet_locks: acc?.meet_locks + new_lock_cnt,
+                    resolved_locks: acc?.resolved_locks + x?.resolved_locks,
                     ..Default::default()
                 })
             })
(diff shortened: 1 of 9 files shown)


src/transaction/requests.rs
```diff
@@ -445,11 +445,13 @@ impl Merge<ResponseWithShard<kvrpcpb::PessimisticLockResponse, Vec<kvrpcpb::Muta
 
 pub fn new_scan_lock_request(
     start_key: Vec<u8>,
+    end_key: Vec<u8>,
     safepoint: u64,
     limit: u32,
 ) -> kvrpcpb::ScanLockRequest {
     let mut req = kvrpcpb::ScanLockRequest::default();
     req.set_start_key(start_key);
+    req.set_end_key(end_key);
```
pingyu, line 454: The `impl Shardable`, `impl HasNextBatch`, and `impl NextBatch` below also need some changes.

haojinming: Range is added in `impl Shardable`. `TODO` is added in `impl HasNextBatch`, as it's not easy to get the request range in `ScanLockResponse`. It does not scan useless regions as `has_more_batch` will be judged before another scan. PTAL~

haojinming:
@pingyu Review comments are addressed and some unstable cases are fixed. PTAL~ thanks.

tests/integration_tests.rs
```diff
@@ -894,7 +894,7 @@ async fn txn_scan_reverse() -> Result<()> {
     ];
 
     // pessimistic
-    let option = TransactionOptions::new_pessimistic().drop_check(tikv_client::CheckLevel::Warn);
```
sticnarf, line 897: Why changes this?

haojinming: > Why changes this?

This case is unstable. Via debug, I found that `new_pessimistic` use `no_backoff` for `lock_backoff`, which is the reason to cause it.  `new_optimistic` will do retry when failure. It seems the transaction option is not the test point in this case, so I prefer to use the more stable one.

sticnarf: Got it. Can you change the comment above and briefly describe the reason in it?

haojinming: Fixed. Thanks.

sticnarf (approved):

sticnarf:
@pingyu Can you please take a look again?

src/transaction/client.rs
```diff
@@ -263,11 +263,11 @@ impl Client {
     ) -> Result<CleanupLocksResult> {
         debug!(self.logger, "invoking cleanup async commit locks");
         // scan all locks with ts <= safepoint
-        let mut start_key = vec![];
         let ctx = ResolveLocksContext::default();
         let backoff = Backoff::equal_jitter_backoff(100, 10000, 50);
         let req = new_scan_lock_request(
-            mem::take(&mut start_key),
+            options.start_key.clone(),
```
pingyu, line 269: Suggest to add `range: impl Into<BoundRange>` argument to `cleanup_locks` instead, to make it consistent in style with other methods.

src/transaction/client.rs
```diff
@@ -287,9 +287,15 @@ impl Client {
         &self,
         safepoint: &Timestamp,
         mut start_key: Vec<u8>,
+        mut end_key: Vec<u8>,
```
pingyu, line 290: Ditto, suggest to use `range: impl Into<BoundRange>`.

src/transaction/client.rs
```diff
@@ -287,9 +287,15 @@ impl Client {
...
         batch_size: u32,
     ) -> Result<Vec<tikv_client_proto::kvrpcpb::LockInfo>> {
-        let req = new_scan_lock_request(mem::take(&mut start_key), safepoint.version(), batch_size);
+        let req = new_scan_lock_request(
+            std::mem::take(&mut start_key),
```
pingyu, line 294: The `mem::take` seems to be not necessary.

src/request/plan.rs
```diff
@@ -574,13 +586,15 @@ where
...
-            result.meet_locks += locks.len();
 
+            let lock_size = locks.len();
             match lock_resolver
                 .cleanup_locks(self.store.clone().unwrap(), locks, self.pd_client.clone())
                 .await
             {
-                Ok(()) => {}
+                Ok(()) => {
+                    result.meet_locks += lock_size;
```
pingyu, line 596: I just recall that at that time I wrote these code, I expected `meet_locks` is the number of locks `cleanup_locks` "meet", and `resolved_locks` the number of "resolved" (there is a "TODO" in `CleanupLocksResult`).

I think it would be better to rename `meet_locks` to `resolved_locks` for recording the number of locks here.

Moreover, if CI is still flaky, we can just remove the check of `meet_locks`. The `count_locks` should be enough for correctness. It's not the issue of this PR. And we can polish them later.

src/transaction/client.rs
```diff
@@ -242,7 +242,7 @@ impl Client {
             batch_size: SCAN_LOCK_BATCH_SIZE,
             ..Default::default()
         };
-        self.cleanup_locks(&safepoint, options).await?;
+        self.cleanup_locks(&safepoint, vec![].., options).await?;
```
pingyu, line 245: `..` should be enough.

```suggestion
        self.cleanup_locks(&safepoint, .., options).await?;
```

src/transaction/client.rs
```diff
@@ -259,18 +259,14 @@ impl Client {
     pub async fn cleanup_locks(
         &self,
         safepoint: &Timestamp,
+        range: impl Into<BoundRange>,
```
pingyu, line 262: Nit: Suggest to keep the same order of argument with `new_scan_lock_request`.

tests/failpoint_tests.rs
```diff
@@ -27,14 +30,17 @@ async fn txn_optimistic_heartbeat() -> Result<()> {
     let mut heartbeat_txn = client
         .begin_with_options(
             TransactionOptions::new_optimistic()
-                .heartbeat_option(HeartbeatOption::FixedTime(Duration::from_secs(1))),
+                .heartbeat_option(HeartbeatOption::FixedTime(Duration::from_secs(1)))
+                .drop_check(CheckLevel::Warn),
```
pingyu, line 36: Because it's not stable in CI ? Suggest to send an issue for this, and leave comments here about that the warning level for drop check should be removed later.

tests/failpoint_tests.rs
```diff
@@ -83,6 +92,7 @@ async fn txn_cleanup_locks_batch_size() -> Result<()> {
 
     init().await?;
     let scenario = FailScenario::setup();
+    let full_range = vec![]..;
```
pingyu, line 97: `..` should work. The same to follows.

src/request/plan.rs
haojinming: Fixed,  renamed `meet_locks` to `resolved_locks`.

src/request/plan.rs
```diff
@@ -478,8 +484,14 @@ impl Merge<CleanupLocksResult> for Collect {
         input
             .into_iter()
             .fold(Ok(CleanupLocksResult::default()), |acc, x| {
+                let new_ret = x?;
```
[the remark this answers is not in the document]
haojinming: Fixed, PTAL~

pingyu (approved):
LGTM~

PR #388, opened 2023-01-12
feat: add DeleteRange support in TransactionClient

guoxiangCN:
As the title said~


andylokandy:
DeleteRange command in the proto is not **safe** to use in a normal transcation process. Specifically, it should be an internal-use-only command used by the GC process.

Issue #369, opened 2022-09-07
raft entry is too large, region 2, entry size 12014018


bxq2011hust:
Write ut in rust client, put 10240 kv pairs in one txn, the value is random bytes and length is 10240, the key is `format!("key_________________________{}", j)` which j is index in `0..10240`, commit the txn failed. the error message is 

```
Error: RegionError(Error { message: "raft entry is too large, region 2, entry size 12014018", not_leader: None, region_not_found: None, key_not_in_region: None, epoch_not_match: None, server_is_busy: None, stale_command: None, store_not_match: None, raft_entry_too_large: Some(RaftEntryTooLarge { region_id: 2, entry_size: 12014018 }), max_timestamp_not_synced: None, read_index_not_ready: None, proposal_in_merging_mode: None, data_is_not_ready: None, region_not_initialized: None, disk_full: None })
```

I run the same ut in c-client, it works fine, I guess the reason is rust client does not split kv pairs into batches in the same region?

ekexium:
> I run the same ut in c-client, it works fine, I guess the reason is rust client does not split kv pairs into batches in the same region?

Yes. It should be improved.

PR #390, opened 2023-01-29
Batch split for prewrite and commit requests 

akoshchiy:
ref #369
Still in progress, here is draft api for splitting/merging requests into batches (RawBatchPutRequest as example). 
After approve i will finish for the rest.

### DIFF
--- a/config/tikv.toml
+++ b/config/tikv.toml
@@ -8,6 +8,7 @@ region-split-check-diff = "1B"
 pd-heartbeat-tick-interval = "2s"
 pd-store-heartbeat-tick-interval = "5s"
 split-region-check-tick-interval = "1s"
+raft-entry-max-size = "1MB"
 
 [rocksdb]
 max-open-files = 10000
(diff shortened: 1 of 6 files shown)


roguepnz:
PTAL @ekexium

ekexium:
Hey thank you for the work! 
I think we'd better not split a raw request which could make it non-atomic.
I guess supporting prewrite and commit requests in transactional API would be enough for us.
And we need a test to verify that it works

roguepnz:
@ekexium Since we don't need reusable api for batches, is it ok to implement batching for prewrite/commit requests directly in Shardable trait implementation? As an example, if we have two regions and two batches inside each region, shards() method will return four "shards".
So, it won't be necessary to merge batch responses, and other logic in request pipeline will be untouched.

ekexium:
> @ekexium Since we don't need reusable api for batches, is it ok to implement batching for prewrite/commit requests directly in Shardable trait implementation? As an example, if we have two regions and two batches inside each region, shards() method will return four "shards". So, it won't be necessary to merge batch responses, and other logic in request pipeline will be untouched.

Yes we can

roguepnz:
@ekexium PTAL.
Added raft-entry-max-size to the config to be able to test with small amount of data.

ekexium (approved):

Issue #382, opened 2022-12-14
gRPC error: RpcFailure: 8-RESOURCE_EXHAUSTED Received message larger than max (4563651 vs. 4194304)


bxq2011hust:
Is there a way to solve this problem?

PR #392, opened 2023-02-23
Unlimited Received Size in common tikv client

BoynChan:
# Purpose
Due to limit received message size in Tikv default client (4MB), it's not enough for many applications that use it to store binary data. I unrestricted the receive message length by adding a option when creating ChannelBuilder. 

# Does it backward compatibility?
Yes, user will not be affected after changes.

# Related Issue
#382 
tikv/grpc-rs#602


andylokandy:
Thank you for your contribution! But I believe it's resolved by https://github.com/tikv/client-rust/pull/399.

PR #394, opened 2023-04-23
Fix get the error ranges in group_ranges_by_region

joker53-1:
When I have multiple ranges and want to get ranges through group_ranges_by_region, if my first range is in the first region, and the start_key of the next range is greater than the end_key of the first region, I will get the following error. At this time The start_key of the obtained range is less than end_key.

running 1 test
thread 'pd::client::test::test_group_ranges_by_region' panicked at 'assertion failed: `(left == right)`
  left: `[KeyRange { start_key: [1], end_key: [5, 2] }, KeyRange { start_key: [11, 4], end_key: [10] }]`,
 right: `[KeyRange { start_key: [1], end_key: [5, 2] }]`', src/pd/client.rs:467:9

Signed-off-by:  joker53-1 johnikezheng@gmail.com


joker53-1:
@ekexium @iosmanthus PTAL

andylokandy:
@joker53-1 Thank you!

andylokandy (approved):

PR #401, opened 2023-07-08
Improve rustfmt config

andylokandy:


rustfmt.toml
```diff
@@ -1,3 +1,10 @@
-imports_granularity="Crate"
+edition = "2021"
+version = "Two"
+reorder_imports = true
+imports_granularity = "Item"
```
pingyu, line 4: Why we change to `Item` ?

Some places will look verbose, e.g. [`src/lib`](https://github.com/tikv/client-rust/pull/401/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759).

andylokandy: item style is more convenient to find and replace, while compact style doesn't have too  much help to efficiency.

iosmanthus (approved):

pingyu (approved):

PR #402, opened 2023-07-08
Combine workspace into a unicrate

andylokandy:


pingyu:
Why we change `slog` back to `log` (see https://github.com/tikv/client-rust/pull/298) ?

It will break the compatibility as we change the interface.

Xuanwo:
> It will break the compatibility as we change the interface.

Since we have already switched to the `tonic` grpc framework, our upcoming release must be `v0.3`, which is considered a breaking change. Therefore, compatibility is not a concern and all we need to consider is whether the answer is `YES` or `NO`.

In my opinion, a library should always use `log` (maybe `tracing`) instead of a specified logger like `slog`. This makes it easier to integrate the library. And please note that `slog` itself is inactive and recommand users to migrate to `tracing` instead.

image

src/common/security.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2018 TiKV Project Authors. Licensed under Apache-2.0.
 
-use crate::Result;
+use crate::{internal_err, Result};
+use log::info;
 // use grpcio::{Channel, ChannelBuilder, ChannelCredentialsBuilder, Environment};
```
Xuanwo, line 5: How about removing this line.

Xuanwo (commented):
Mostly LGTM, thanks!

Xuanwo (approved):

pingyu (approved):
LGTM~

PR #403, opened 2023-07-09
docs: Cmake and gcc is not needed anymore

Xuanwo:
This PR removed not needed cmake and gcc deps and update the links of CI.


Xuanwo:
cc @andylokandy and @ekexium, Would you like to take a look? Thanks.

andylokandy (approved):

Issue #398, opened 2023-07-06
The rawkv scan performance of is severely degraded.


Smityz:
I put some data in my TiKV cluster(version v6.5.3), and test the performance between Rust SDK and Go SDK, and found a serious performance problem. 
I tried to scan the entire table with a limit of 1000 data per RPC and found that the delay per RPC for the Go SDK was 3ms while the delay for the Rust SDK was 200+ms. However, the server showed the same delay in both cases.
Another strange thing is that when testing with Rust, the server displays a scan QPS of 1000 while GoLang is around 500. It is obvious that the Rust SDK is performing abnormally in the test.
Here is my test code:
```rust
// run in release mode
#![type_length_limit = "8165158"]

mod common;

use std::time::Instant;

use crate::common::parse_args;
use tikv_client::{Config, IntoOwnedRange,Key,  RawClient as Client, Result};

#[tokio::main]
async fn main() -> Result<()> {
    // You can try running this example by passing your pd endpoints
    // (and SSL options if necessary) through command line arguments.
    let args = parse_args("raw");

    // Create a configuration to use for the example.
    // Optionally encrypt the traffic.
    let config = if let (Some(ca), Some(cert), Some(key)) = (args.ca, args.cert, args.key) {
        Config::default().with_security(ca, cert, key)
    } else {
        Config::default()
    };

    // When we first create a client we receive a `Connect` structure which must be resolved before
    // the client is actually connected and usable.
    let client = Client::new_with_config(args.pd, config, None).await?;
    let client = client.clone();

    let mut scan_range =("".."z").into_owned();  
    loop {
        let start = Instant::now();
        let pairs = client
            .scan(scan_range.clone(), 1000)
            .await
            .expect("Could not scan");
        println!("duration: {}", start.elapsed().as_millis());
        let keys: Vec<_> = pairs.into_iter().map(|p| p.key().clone()).collect();
        scan_range = (keys.last().unwrap().clone()..=String::from("z").clone().into()).into();
    }
}
```

```Golang
package main

import (
	"context"
	"fmt"
	"time"

	"github.com/tikv/client-go/v2/config"
	"github.com/tikv/client-go/v2/rawkv"
)

func main() {
	cli, err := rawkv.NewClient(context.TODO(), []string{"127.0.0.1:2379"}, config.DefaultConfig().Security)
	if err != nil {
		panic(err)
	}
	defer cli.Close()
	startKey := ""
	endKey := "zzz"
	for {
		startTime := time.Now()
[...]

PR #409, opened 2023-07-13
fix raw scan

Smityz:
related issue: https://github.com/tikv/client-rust/issues/398

In this PR, I changed the behavior of Rawkv scan. Previously, it would send RPCs to multiple regions in parallel, which was not reasonable and could result in incorrect results. Now, I have changed it so that each RPC is only sent to one region at a time. In addition, I have made special handling for scan RPCs that cross regions.


Smityz:
@ekexium @andylokandy Hi masters, PTAL~

Smityz:
@andylokandy ready for review now, PTAL

src/raw/client.rs
```diff
@@ -591,17 +592,50 @@ impl<PdC: PdClient> Client<PdC> {
...
+                        );
+                        rs
+                    }
+                    Some(Err(e)) => return Err(e),
+                    None => {
+                        return Ok(result);
+                    }
+                };
+            }
+            limit -= res_len as u32;
```
andylokandy, line 636: this could panic if the server returns more than requested, since `limit` if merely a soft limit.

Smityz: fixed

andylokandy (approved):

andylokandy:
Thanks!

Issue #410, opened 2023-07-13
Add vendor support so users don't need to install PROTOC


Xuanwo:
Maybe it's worth to add vendor support so users can use tikv-client without install PROTOC.

PR #411, opened 2023-07-13
feat: Add vendored support

Xuanwo:
Close https://github.com/tikv/client-rust/issues/410


Xuanwo:
Should wait for https://github.com/tikv/client-rust/pull/412

Xuanwo:
cc @andylokandy, @ekexium & @pingyu, would you like to take a review? Thanks!

.gitattributes
```diff
@@ -0,0 +1 @@
+src/generated/ diff=false
```
tisonkun, line 1: Why do we need this line?

tisonkun (changes requested):
> without install PROTOC

I'd prefer to bundle PROTOC or use protobuf-src instead. Like what is done in https://github.com/tikv/protobuf-build/pull/67

.gitattributes
Xuanwo: > Why do we need this line?

protoc generated files should be treated as binary files that no need (and no way, since we can't change them by hand) for review.

tisonkun:
Vendor generated looks like a step backward. If you do want to decouple with a PROTOC from the user side, it should be the default manner instead of a feature flag.

That is, we always generated, ship the generated, and check the generated updated with CI.

Xuanwo:
> If you do want to decouple with a PROTOC from the user side, it should be the default manner instead of a feature flag.

We need a feature flag if we want to do this in `build.rs`. Otherwise, we will need an extra build script to do this which adds extra burden for developers.

Current design is:

- Without `vendored` feature, generate code and use those code.
- With `vendored` feature, use code directly.

Xuanwo:
cc @tisonkun, would you like to take a review again?

tisonkun (approved):

andylokandy (approved):

andylokandy:
Thanks!

proto-build/src/main.rs
```diff
@@ -1,12 +1,14 @@
...
-
-use tonic_disable_doctest::BuilderEx;
+// Copyright 2023 TiKV Project Authors. Licensed under Apache-2.0.
 
 fn main() {
+    use tonic_disable_doctest::BuilderEx;
+
     tonic_build::configure()
         .disable_doctests_for_types([".google.api.HttpRule"])
+        .emit_rerun_if_changed(false)
```
andylokandy, line 8: By the way, it's interesting to know why the auto rerun is disabled.

Xuanwo: > By the way, it's interesting to know why the auto rerun is disabled.

After this change, we are running codegen in a new bin instead of `build.rs`, the `rerun` hint for cargo is not working anymore. Instead, they just print to the console.

PR #412, opened 2023-07-14
build: Make sure tikv-client works on stable rust

Xuanwo:
This PR add `rust-toolchain.toml` for this project to make sure all our development happen under the stable rust version.

> It would be ideal to have an MSRV policy, but it is not urgent. We can establish one once we have a sufficient user base.


Xuanwo:
cc @andylokandy, @ekexium & @pingyu, would you like to take a review? Thanks!

sticnarf (approved):

rustfmt.toml
```diff
@@ -1,10 +1,11 @@
...
 reorder_imports = true
-imports_granularity = "Item"
-group_imports = "StdExternalCrate"
-where_single_line = true
-trailing_comma = "Vertical"
-overflow_delimited_expr = true
-format_code_in_doc_comments = true
-normalize_comments = true
+
+# version = "Two"
```
sticnarf, line 4: What about just removing these style settings?

Xuanwo: I expect to enable them back once got stabilized. It looks good to me to remove them.

sticnarf: I see.

I think you can keep them but leave a comment about adding them back once stabilized. Otherwise, we will forget why they are commented out.

andylokandy (approved):

pingyu (approved):
LGTM~

Xuanwo:
Fixed, please review again @sticnarf

sticnarf:
Thanks @Xuanwo

PR #414, opened 2023-08-09
add config to solve by custom dns server

Smityz:
solve: https://github.com/tikv/client-go/issues/829

I have provided a tool for client-rust that allows for the customization of DNS server addresses. This way, by only exposing the coreDNS address of the k8s cluster, it is possible to achieve client access to tikv across k8s clusters.


andylokandy:
The DNS/cloud/k8s should be transparent to tikv's client. It could be done by the outer code who is the caller of the client.

Smityz:
Users can modify system settings to use the correct DNS address, but it requires root permission and not inconvenient.

andylokandy:
The client user must be another rust code that are using the client. And that is there is room for such customization without needing root permission.

Smityz:
When TiKV runs on Kubernetes, PD stores the domain of the TiKV node instead of its IP address. Therefore, this function must be configured in the client.

Issue #359, opened 2022-07-15
Support KeySpace


iosmanthus:
[KeySpace](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md) is fundamental for transaction safety and multi-tenant features for TiKV. This is the tracking issue for the feature in client-rust. For tracking issues of other components, check:

* TiKV: https://github.com/tikv/tikv/issues/12999
* PD: https://github.com/tikv/pd/issues/5293
* client-go: https://github.com/tikv/client-go/issues/545

Currently, https://github.com/tikv/client-rust/pull/353 has introduced a new type parameter for both the transaction and the raw client to support an extensible request codec. We have two implementations for now: plain API V1 and API V2 with the keyspace feature. However, we only support constructing a new API V2 instance by `ApiV2::default()`. More `new` functions should be provided to enable users to create keyspaces by `keyspace_name` or `keyspace_id`. This requires:

- [ ] implement acquiring keyspace info from PD. @iosmanthus 
- [ ] introduces a keyspace cache for each instance of the client. @iosmanthus
- [ ] unit/integration tests for API v2 @iosmanthus

eduralph:
Shipped via #518/#522 — API-v2 keyspace encoding is wired through both the raw and transactional paths (`Config::with_keyspace`). Can this be closed?

PR #415, opened 2023-08-15
*: Support API v2 (part 1)

pingyu:
Ref: #359 

This PR is part of #353 to make the code review easier.

### Changes
- Implement the basic skeleton of codec.
- Support set api version of API v2 for request context.

### Coming PRs
- Encode request by adding keyspace prefix.
- Decode response by removing keyspace prefix.
- PD client interface for keyspace.
- Unit & integration tests for API v2.

### DIFF
--- a/src/request/codec.rs
+++ b/src/request/codec.rs
@@ -9,16 +9,21 @@ pub trait Codec: Clone + Sync + Send + 'static {
 }
 
 #[derive(Clone, Default)]
-pub struct ApiV1Codec {}
+pub struct ApiV1TxnCodec {}
 
-impl Codec for ApiV1Codec {}
+impl Codec for ApiV1TxnCodec {}
+
+#[derive(Clone, Default)]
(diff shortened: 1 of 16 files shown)


Smityz:
This is really great, I thought this feature wouldn't be developed before

src/request/codec.rs
```diff
@@ -0,0 +1,54 @@
...
+        // TODO: req.encode_request(self);
+    }
+}
+
+// TODO: pub struct ApiV2RawCodec
+
+// EncodeRequest is just a type wrapper to avoid passing not encoded request to `PlanBuilder` by mistake.
+#[derive(Clone)]
+pub struct EncodedRequest<Req: KvRequest> {
+    pub inner: Req,
```
iosmanthus, line 46: How about making it an Option, and taking it out after first encoding to avoid repeated encoding?

iosmanthus (approved):

PR #429, opened 2023-10-31
transaction: use AtomicU8 for status

### LINKED ISSUE #428: transaction: `Transaction` would be hanged in dropping.
In a long running program which using `client-rust` to request TiKV servers, there are low chances that the program will hang.

The hang position should be in dropping the `Transaction` object. After adding logs, we can see the log of ["dropping transaction"](https://github.com/tikv/client-rust/blob/788d6e211bdb424b854d5058b63acd48df1d576a/src/transaction/transaction.rs#L984) but the outer code block did not exit.

To further confirm the issue, I write a snippet to make the problem reproducible (see [here](https://github.com/tikv/client-rust/compare/master...pingyu:client-rust:drop-hang)).

```rust
    async fn txn_task() {
        let pd_client = Arc::new(MockPdClient::default());
        let options = TransactionOptions::new_pessimistic().read_only();
        let txn = Transaction::new(Timestamp::default(), pd_client, options);
        txn.check_allow_operation().await.unwrap();
        drop(txn);
    }

    #[tokio::test]
    async fn test_transaction_drop_st() -> Result<(), io::Error> {
        txn_task().await;
        Ok(())
    }

    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
    async fn test_transaction_drop_mt() -> Result<(), io::Error> {
        const CONCURRENCY: usize = 10;
        let mut handles = Vec::with_capacity(CONCURRENCY);
        for _ in 0..CONCURRENCY {
            handles.push(tokio::spawn(txn_task()));
        }

        for handle in handles {
            handle.await.unwrap();
        }
        Ok(())
    }
```

There is also a small change on `Transaction::drop` (the `tokio::time::sleep(Duration::from_millis(100)).await`), otherwise it's not possible to reproduce the problem. The `sleep` is to simulate the scene that the lock can no be acquired immediately, which would happen when, i.e, the [`heartbeat_task`](https://github.com/tikv/client-rust/blob/788d6e211bdb424b854d5058b63acd48df1d576a/src/transaction/transaction.rs#L949) has acquired the lock.

```rust
    fn drop(&mut self) {
        debug!("dropping transaction");
        if std::thread::panicking() {
            return;
        }
        let mut status = futures::executor::block_on(async {
            tokio::time::sleep(Duration::from_millis(100)).await;
            self.status.write().await
        });
        if *status == TransactionStatus::Active {
            match self.options.check_level {
                CheckLevel::Panic => {
                    panic!("Dropping an active transaction. Consider commit or rollback it.")
                }
                CheckLevel::Warn => {
                    warn!("Dropping an active transaction. Consider commit or rollback it.")
                }

                CheckLevel::None => {}
            }
        }
        *status = TransactionStatus::Dropped;
    }
```

Running these two tests will definitely hang:

```sh
❯ cargo test test_transaction_drop_mt -- --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.14s
     Running unittests src/lib.rs (target/debug/deps/tikv_client-802bb59de2ea8dfe)

running 1 test
test transaction::transaction::tests::test_transaction_drop_mt has been running for over 60 seconds
^C
❯
❯ cargo test test_transaction_drop_st -- --nocapture
   Compiling tikv-client v0.2.0 (/data/nvme0n1/home/pingyu/workspace/client-rust)
    Finished test [unoptimized + debuginfo] target(s) in 10.50s
     Running unittests src/lib.rs (target/debug/deps/tikv_client-802bb59de2ea8dfe)

running 1 test
test transaction::transaction::tests::test_transaction_drop_st has been running for over 60 seconds
^C
```

This should be a known issue when using `futures::executor::block_on` in an async context. See https://github.com/tokio-rs/tokio/issues/2376.

pingyu:
Close #428 

### Changes

- Use `AtomicU8` for `Transaction::status` to avoid `futures::executor::block_on` in `Transaction::drop`. Besides, as status of transaction object is unlikely to have contention, atomic operations should have better performance.

### DIFF
--- a/src/transaction/lock.rs
+++ b/src/transaction/lock.rs
@@ -98,7 +98,7 @@ pub async fn resolve_locks(
         .await?;
         clean_regions
             .entry(lock.lock_version)
-            .or_insert_with(HashSet::new)
+            .or_default()
             .insert(cleaned_region);
     }
     Ok(live_locks)
--- a/src/transaction/snapshot.rs
+++ b/src/transaction/snapshot.rs
@@ -22,7 +22,7 @@ use crate::Value;
 ///
 /// See the [Transaction](struct@crate::Transaction) docs for more information on the methods.
 #[derive(new)]
-pub struct Snapshot<Cod: Codec = ApiV1TxnCodec, PdC: PdClient = PdRpcClient<Cod>> {
+pub struct Snapshot<Cod: Codec = ApiV1TxnCodec, PdC: PdClient<Codec = Cod> = PdRpcClient<Cod>> {
     transaction: Transaction<Cod, PdC>,
     phantom: PhantomData<Cod>,
 }
--- a/src/transaction/transaction.rs
+++ b/src/transaction/transaction.rs
@@ -2,6 +2,8 @@
 
 use std::iter;
 use std::marker::PhantomData;
+use std::sync::atomic;
+use std::sync::atomic::AtomicU8;
 use std::sync::Arc;
 use std::time::Instant;
 
@@ -10,7 +12,6 @@ use fail::fail_point;
 use futures::prelude::*;
 use log::debug;
 use log::warn;
-use tokio::sync::RwLock;
 use tokio::time::Duration;
 
 use crate::backoff::Backoff;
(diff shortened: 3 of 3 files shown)


iosmanthus (approved):
LGTM

Issue #407, opened 2023-07-13
prometheus should be optional and hide under feature gate


Xuanwo:
`prometheus` should be enabled by default when users don't need it.

Xuanwo:
Related to https://github.com/apache/incubator-opendal/issues/3612, `prometheus` will also include `reqwest` with `native-tls` enable.

Xuanwo:
@pingyu, do you think it's a good idea to make prometheus optional?

pingyu:
> @pingyu, do you think it's a good idea to make prometheus optional?

Yes~

Xuanwo:
Can you make a patch release for this change? Thanks!

iosmanthus:
> Can you make a patch release for this change? Thanks!

@andylokandy PTAL

PR #433, opened 2023-11-20
refactor: Make prometheus features `push` and `process` optional to get rid of openssl

Xuanwo:
This PR fix https://github.com/tikv/client-rust/issues/407 in a compatible way:

~~- Users enable `default` feature still enables `prometheus/process`.~~
~~- Users disable `default` feature can enable feature `prometheus` based on their needs.~~

`prometheus` is so deep in the tikv-client's code, we can't remove it entirely. This PR changes it goal to make `push` and `process` optional.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ edition = "2021"
 
 [features]
 default = ["prometheus"]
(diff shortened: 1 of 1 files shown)


pingyu (approved):
LGTM~

PR #438, opened 2023-11-25
Fix reverse scan for scene of multiple regions

### LINKED ISSUE #436: InvalidReqRange with `scan_reverse`
Hello,

It seems there is some issue which results in `InvalidReqRange` error where the start key of the region is greater than the start key of the request, which is triggered by the following check in TiKV:
https://github.com/tikv/tikv/blob/master/src/storage/txn/store.rs#L465-L477

It appears that when using `scan_reverse` the [start and end keys are swapped](https://github.com/tikv/client-rust/blob/master/src/transaction/requests.rs#L155-L161), such that the greater key is the start key. But, in `store_stream_for_range` the `range_intersection` function appears to always treat the start key of the request range (`bnd_range`) as the lower key and the end key as the upper key when comparing with the region range, which breaks the lower bound of the resulting range for the store. Perhaps this is the issue?

https://github.com/tikv/client-rust/blob/master/src/store/mod.rs#L79-L112

Thanks

pingyu:
Close #436 

Fix the issue by swap `start_key` & `end_key` before requesting regions from PD, and swap back before apply shards to request. As the `start_key` & `end_key` in reverse range request are in the meaning of `[end_key, start_key)`.

The issue can be reproduced before the `shardable_range` was fixed. See 435cf0a36dbb1026c32f757f94bf86f7e713fbb1:
```bash
MULTI_REGION=1 cargo test txn_scan_reverse --all --features "integration-tests" -- --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.11s
     Running unittests src/lib.rs (target/debug/deps/tikv_client-ebeddab360b427a7)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 39 filtered out; finished in 0.00s

     Running tests/failpoint_tests.rs (target/debug/deps/failpoint_tests-f2d522e63dab2b69)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 5 filtered out; finished in 0.00s

     Running tests/integration_tests.rs (target/debug/deps/integration_tests-aae509a3e2e0836d)

running 1 test
init finish with 290 regions
Error: MultipleKeyErrors([KeyError(KeyError { locked: None, retryable: "", abort: "Error(Txn(Error(InvalidReqRange { start: Some([0, 0, 0, 0, 0, 0, 0, 0, 251]), end: Some([192, 0, 0, 0, 0, 0, 0, 0, 252]), lower_bound: Some([191, 191, 253, 1, 0, 0, 0, 0, 251]), upper_bound: Some([193, 63, 252, 251, 0, 0, 0, 0, 251]) })))", conflict: None, already_exist: None, deadlock: None, commit_ts_expired: None, txn_not_found: None, commit_ts_too_large: None, assertion_failed: None })])
test txn_scan_reverse ... FAILED

failures:

failures:
    txn_scan_reverse
```

See also https://github.com/tikv/client-rust/actions/runs/6989951811/job/19018973704?pr=438.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,8 @@ export RUSTFLAGS=-Dwarnings
 
 .PHONY: default check unit-test integration-tests test doc docker-pd docker-kv docker all
 
-PD_ADDRS ?= "127.0.0.1:2379"
-MULTI_REGION ?= 1
+export PD_ADDRS     ?= 127.0.0.1:2379
+export MULTI_REGION ?= 1
 
 ALL_FEATURES := integration-tests
 
--- a/src/raw/requests.rs
+++ b/src/raw/requests.rs
@@ -16,6 +16,7 @@ use crate::proto::kvrpcpb;
 use crate::proto::kvrpcpb::ApiVersion;
 use crate::proto::metapb;
 use crate::proto::tikvpb::tikv_client::TikvClient;
+use crate::range_request;
 use crate::request::plan::ResponseWithShard;
 use crate::request::Collect;
 use crate::request::CollectSingle;
@@ -23,6 +24,7 @@ use crate::request::DefaultProcessor;
 use crate::request::KvRequest;
 use crate::request::Merge;
 use crate::request::Process;
+use crate::request::RangeRequest;
 use crate::request::Shardable;
 use crate::request::SingleKey;
 use crate::shardable_key;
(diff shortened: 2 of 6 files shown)


pingyu:
CC @jmhrpr. Thank you for providing the clue! I will fix it this weekend.

iosmanthus (approved):

PR #439, opened 2023-12-14
implement the complete keyspace feature

andylokandy:
- Revert "*: Support API v2 (part 1) (#415)"
- --wip-- [skip ci]


pingyu:
What's wrong with #415 ?

andylokandy:
> What's wrong with https://github.com/tikv/client-rust/pull/415 ?

This is the complete implementation and #415 seems not handy while I continued its work.

src/pd/cluster.rs
```diff
@@ -91,6 +93,18 @@ impl Cluster {
         req.safe_point = safepoint;
         req.send(&mut self.client, timeout).await
     }
+
+    pub async fn get_keyspace_id(&self, keyspace: &str) -> Result<u32> {
```
iosmanthus, line 97: There is a grpc method for this interface.

andylokandy: I'll modify this part in the next PR.

tests/integration_tests.rs
```diff
@@ -1262,8 +1296,10 @@ async fn do_mutate(is_pessimistic: bool) -> Result<()> {
     }
 }
 
-async fn verify_mutate(is_pessimistic: bool) {
-    let client = TransactionClient::new(pd_addrs()).await.unwrap();
+async fn verify_mutate(is_pessimistic: bool) -> Result<()> {
+    let client =
+        TransactionClient::new_with_config(pd_addrs(), Config::default().with_default_keyspace())
+            .await?;
```
pingyu, line 1302: It's LGTM here. But actually I prefer using `unwrap()` in test cases, as sometimes it's not easy to find out the error position when using `?`.

config/tikv.toml
```diff
@@ -15,3 +15,7 @@ max-open-files = 10000
 
 [raftdb]
 max-open-files = 10000
+
+[storage]
```
pingyu, line 19: I think it's necessary to also test for API v1.

It seems that it will make codes verbose when construct `Config` in integration tests. Maybe we can write a method to detect the API version and then create the `Config` accordingly. It's OK to improve here in another PR.

pingyu (commented):
BTW, please fix the DCO error.

tests/integration_tests.rs
andylokandy: I was just trying to keep the style with other tests.

andylokandy:
> fix the DCO error.

I can not fix it unless force pushing. But since we are going to squash and merge, it'll be ok because github will generate a good commit with DCO.

src/request/keyspace.rs
```diff
@@ -149,6 +149,15 @@ impl TruncateKeyspace for Vec<KvPair> {
     }
 }
 
+impl TruncateKeyspace for Vec<crate::proto::kvrpcpb::LockInfo> {
+    fn truncate_keyspace(mut self, keyspace: Keyspace) -> Self {
+        for lock in self.locks.iter_mut() {
```
pingyu, line 154: Fields of `primary_lock` & `secondaries` may also need to be truncated keyspace.

andylokandy:
@pingyu Done. PTAL

tests/common/mod.rs
```diff
@@ -24,17 +24,11 @@ const REGION_SPLIT_TIME_LIMIT: Duration = Duration::from_secs(15);
 
 // Delete all entries in TiKV to leave a clean space for following tests.
 pub async fn clear_tikv() {
-    // DEFAULT_REGION_BACKOFF is not long enough for CI environment. So set a longer backoff.
```
pingyu, line 32: Why remove the `backoff` ? It was added as we had faced the timeout of `delete_range` in CI before.

andylokandy: i removed because it seems ok on ci now

pingyu: It's ok now doesn't mean that it will always be. In my experience, the running environment of Github actions are not stable.

pingyu (approved):
LGTM~

@iosmanthus PTAL~

PR #445, opened 2024-03-11
Fix stale region cache with no leader

yongman:
When we create client during tikv-server startup and the region has no leader been elected yet, the region cache in client may be stale with no leader.

It will cause the region access return no leader error until the region id_ver changed.

### DIFF
--- a/src/region_cache.rs
+++ b/src/region_cache.rs
@@ -85,7 +85,8 @@ impl<C: RetryClientTrait> RegionCache<C> {
                 .get(&candidate_region_ver_id)
                 .unwrap();
 
-            if region.contains(key) {
+            // Region in cache maybe stale if the region has not been elected leader yet during start.
+            if region.contains(key) && region.leader.is_some() {
                 return Ok(region.clone());
             }
         }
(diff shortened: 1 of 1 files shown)


pingyu:
It seems that if there is still no leader when read through PD server, we would all the same get the no leader error.

How about try to handle this situation uniformly by `handle_region_error` ? Then this error can be retried, as well as backoff to avoid cause too much press to PD servers.

*(It's likely that some related codes need to be changed too as this error is raised at [`apply_shard`](https://github.com/tikv/client-rust/blob/c6110dd8087a4d9e7bf48da77bbbfafa79614b10/src/request/plan.rs#L121). Maybe we can try to pass the `region_store` to `single_shard_handler` and handle the condition of no leader there.)*

yongman:
@pingyu Thanks your advise. It's not enough just handling the `NotLeader` error in `single_shard_handler`.  In `Shardable::shards`, `store_stream_for_keys`, `store_stream_for_range`, `store_stream_for_ranges` and `resolve_locks` will also raise this error.

This seems to require lots of modifications, which could take a lot of time and introduce more risks. Moreover, the logic of the application should have the ability to retry and backoff during handling this error, so just refresh the region cache seems reasonable.

PR #450, opened 2024-05-22
prevent PANIC when leader is None

cre4ture:
Hello,

I'm currently experimenting with TiFS (https://github.com/Hexilee/tifs).
When copying large files, the TiKV cluster that I have locally gets under a constant load for a longer period of time (~1-2 hours).
During these tests I came accross a panic caused by the rust-client. I located the code responsible for this, and implemented a fix. 
Please have a look, tell me what you think and what to adapt to let it get into the main branch. 

Thanks and best regards,
cre4ture

log entry of the panic:
```
thread 'tokio-runtime-worker' panicked at /home/uli/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tikv-client-0.3.0/src/pd/cluster.rs:270:47: called `Option::unwrap()` on a `None` value
```

### DIFF
--- a/src/common/errors.rs
+++ b/src/common/errors.rs
@@ -91,7 +91,9 @@ pub enum Error {
     RegionNotFoundInResponse { region_id: u64 },
     /// No leader is found for the given id.
     #[error("Leader of region {} is not found", region_id)]
-    LeaderNotFound { region_id: u64 },
+    LeaderOfRegionNotFound { region_id: u64 },
+    #[error("Leader of cluster {} is not found", cluster_id)]
+    LeaderOfClusterNotFound { cluster_id: u64 },
     /// Scan limit exceeds the maximum
     #[error("Limit {} exceeds max scan limit {}", limit, max_limit)]
     MaxScanLimitExceeded { limit: u32, max_limit: u32 },
--- a/src/pd/cluster.rs
+++ b/src/pd/cluster.rs
@@ -279,7 +279,7 @@ impl Connection {
         keyspacepb::keyspace_client::KeyspaceClient<Channel>,
         pdpb::GetMembersResponse,
     )> {
-        let previous_leader = previous.leader.as_ref().unwrap();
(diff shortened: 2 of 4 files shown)


pingyu (commented):
The panic seems to happen only when PD return `GetMembersResponse` without `leader` field. But I think it should not happen as only the PD leader can return a successful response of get members request.

So I think the root cause would be, we actually get a `GetMemebersResponse` with error (see [`GrpcServer.GetMembers`](https://github.com/tikv/pd/blob/7cc3b4e08298d51d278c797f1989022bb2ae02e3/server/grpc_service.go#L449)). We need the check the `header.error` at the [first place](https://github.com/tikv/client-rust/blob/c6110dd8087a4d9e7bf48da77bbbfafa79614b10/src/pd/cluster.rs#L233).

pingyu:
The issue of handling `get_members` of PD was fixed by #452. We should not encounter the situation that `leader` is `None` anymore.

@cre4ture Could you help with some testing to confirm that this issue has been completely resolved ? Thank you ~

cre4ture:
> The issue of handling `get_members` of PD was fixed by #452. We should not encounter the situation that `leader` is `None` anymore.
> 
> @cre4ture Could you help with some testing to confirm that this issue has been completely resolved ? Thank you ~

I will test and report in case I still see it. Thanks for notification.

I assume this PR would be obsolete then? Or should I prepare still some parts of it for merging?

pingyu:
> > The issue of handling `get_members` of PD was fixed by #452. We should not encounter the situation that `leader` is `None` anymore.
> > @cre4ture Could you help with some testing to confirm that this issue has been completely resolved ? Thank you ~
> 
> I will test and report in case I still see it. Thanks for notification.
> 
> I assume this PR would be obsolete then? Or should I prepare still some parts of it for merging?

It is obsolete now. Unless it's not the reason of `get_members` in your case.

Issue #455, opened 2024-06-17
Scan returning empty results


limbooverlambda:
We were running into an issue with scans where periodically we noticed scans returning empty results for datasets that were present in a cluster. The hypothesis was that the scans were returning empties when the regions are undergoing splits. While trying to reproduce the issue (by issuing splits from pd-ctl), we found out that when there's a region error (epoch_version_mismatch et al), the scan_inner is not triggering any cache invalidations and subsequent retries. The scan simply returns an empty. Here's the PR https://github.com/tikv/client-rust/pull/456 to fix the issue by wrapping the scans in a retryer upon any region errors.

cc: @pingyu @ekexium @andylokandy.

limbooverlambda:
Fix in https://github.com/tikv/client-rust/pull/456. Thanks @pingyu @iosmanthus @hbisheng @zhangjinpeng87 for shaping the PR.

PR #456, opened 2024-06-17
adding retryable to scan

limbooverlambda:
solve: https://github.com/tikv/client-rust/issues/455

We were running into an issue with scans where periodically we noticed scans returning empty results for datasets that were present in a cluster. The hypothesis was that the scans were returning empties when the regions are undergoing splits. While trying to reproduce the issue (by issuing splits from pd-ctl), we found out that when there's a region error (epoch_version_mismatch et al), the scan_inner is not triggering any cache invalidations and subsequent retries. The scan simply returns an empty. This PR is fixing the issue by triggering the invalidations and retry for such issues.

@pingyu @ekexium @andylokandy.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ prometheus = ["prometheus/push", "prometheus/process"]
 # Enable integration tests with a running TiKV and PD instance.
 # Use $PD_ADDRS, comma separated, to set the addresses the tests use.
 integration-tests = []
-protobuf-codec = []
 
 [lib]
 name = "tikv_client"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ prometheus = ["prometheus/push", "prometheus/process"]
 # Enable integration tests with a running TiKV and PD instance.
 # Use $PD_ADDRS, comma separated, to set the addresses the tests use.
 integration-tests = []
+protobuf-codec = []
 
 [lib]
 name = "tikv_client"
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ generate:
 check: generate
 	cargo check --all --all-targets --features "${ALL_FEATURES}"
 	cargo fmt -- --check
-	cargo clippy --fix --allow-dirty --all-targets --features "${ALL_FEATURES}" -- -D clippy::all
+	cargo clippy --all-targets --features "${ALL_FEATURES}" -- -D clippy::all
 
 unit-test: generate
 	cargo nextest run --all --no-default-features
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ generate:
 check: generate
 	cargo check --all --all-targets --features "${ALL_FEATURES}"
 	cargo fmt -- --check
-	cargo clippy --all-targets --features "${ALL_FEATURES}" -- -D clippy::all
+	cargo clippy --fix --allow-dirty --all-targets --features "${ALL_FEATURES}" -- -D clippy::all
 
 unit-test: generate
 	cargo nextest run --all --no-default-features
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -806,16 +806,16 @@ impl<PdC: PdClient> Client<PdC> {
     ) -> Result<(Option<RawScanResponse>, Key)> {
         let start_key = scan_args.start_key;
 
-        let region = self.rpc.clone().region_for_key(&start_key).await?;
-        let store = self.rpc.clone().store_for_id(region.id()).await?;
-        let request = new_raw_scan_request(
-            scan_args.range.clone(),
-            scan_args.limit,
-            scan_args.key_only,
-            scan_args.reverse,
-            self.cf.clone(),
-        );
         loop {
+            let region = self.rpc.clone().region_for_key(&start_key).await?;
+            let store = self.rpc.clone().store_for_id(region.id()).await?;
+            let request = new_raw_scan_request(
+                scan_args.range.clone(),
+                scan_args.limit,
+                scan_args.key_only,
+                scan_args.reverse,
+                self.cf.clone(),
+            );
             let resp = self.do_store_scan(store.clone(), request.clone()).await;
             return match resp {
                 Ok(mut r) => {
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -1,29 +1,34 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
+use async_recursion::async_recursion;
 use core::ops::Range;
 use std::str::FromStr;
 use std::sync::Arc;
 
-use futures::StreamExt;
 use log::debug;
+use tokio::sync::Semaphore;
+use tokio::time::sleep;
 
-use crate::backoff::DEFAULT_REGION_BACKOFF;
+use crate::backoff::{DEFAULT_REGION_BACKOFF, DEFAULT_STORE_BACKOFF};
 use crate::common::Error;
 use crate::config::Config;
 use crate::pd::PdClient;
 use crate::pd::PdRpcClient;
+use crate::proto::kvrpcpb::{RawScanRequest, RawScanResponse};
 use crate::proto::metapb;
 use crate::raw::lowering::*;
-use crate::request::Collect;
 use crate::request::CollectSingle;
 use crate::request::EncodeKeyspace;
 use crate::request::KeyMode;
 use crate::request::Keyspace;
 use crate::request::Plan;
 use crate::request::TruncateKeyspace;
+use crate::request::{plan, Collect};
+use crate::store::{HasRegionError, RegionStore};
 use crate::Backoff;
 use crate::BoundRange;
 use crate::ColumnFamily;
+use crate::Error::RegionError;
 use crate::Key;
 use crate::KvPair;
 use crate::Result;
@@ -755,57 +760,42 @@ impl<PdC: PdClient> Client<PdC> {
                 max_limit: MAX_RAW_KV_SCAN_LIMIT,
             });
         }
-
-        let mut cur_range = range.into().encode_keyspace(self.keyspace, KeyMode::Raw);
+        let backoff = DEFAULT_STORE_BACKOFF;
+        let permits = Arc::new(Semaphore::new(16));
+        let range = range.into().encode_keyspace(self.keyspace, KeyMode::Raw);
         let mut result = Vec::new();
-        let mut scan_regions = self.rpc.clone().stores_for_range(cur_range.clone()).boxed();
-        let mut region_store =
-            scan_regions
-                .next()
-                .await
-                .ok_or(Error::RegionForRangeNotFound {
-                    range: (cur_range.clone()),
-                })??;
-        let mut cur_limit = limit;
-
-        while cur_limit > 0 {
-            let request = new_raw_scan_request(
-                cur_range.clone(),
-                cur_limit,
+        let mut current_limit = limit;
+        let (start_key, end_key) = range.clone().into_keys();
+        let mut current_key: Option<Key> = Some(start_key);
+        while current_limit > 0 {
+            let scan_args = ScanInnerArgs {
+                start_key: current_key.clone(),
+                range: range.clone(),
+                limit,
                 key_only,
                 reverse,
-                self.cf.clone(),
-            );
-            let resp = crate::request::PlanBuilder::new(self.rpc.clone(), self.keyspace, request)
-                .single_region_with_store(region_store.clone())
-                .await?
-                .plan()
-                .execute()
-                .await?;
-            let mut region_scan_res = resp
-                .kvs
-                .into_iter()
-                .map(Into::into)
-                .collect::<Vec<KvPair>>();
-            let res_len = region_scan_res.len();
-            result.append(&mut region_scan_res);
+                permits: permits.clone(),
+                backoff: backoff.clone(),
+            };
+            let (res, next_key) = self.retryable_scan(scan_args).await?;
 
-            // if the number of results is less than cur_limit, it means this scan range contains more than one region, so we need to scan next region
-            if res_len < cur_limit as usize {
-                region_store = match scan_regions.next().await {
-                    Some(Ok(rs)) => {
-                        cur_range = BoundRange::new(
-                            std::ops::Bound::Included(region_store.region_with_leader.range().1),
-                            cur_range.to,
-                        );
-                        rs
-                    }
-                    Some(Err(e)) => return Err(e),
-                    None => break,
-                };
-                cur_limit -= res_len as u32;
-            } else {
+            let mut kvs = res
+                .map(|r| r.kvs.into_iter().map(Into::into).collect::<Vec<KvPair>>())
+                .unwrap_or(Vec::new());
+
+            if !kvs.is_empty() {
+                current_limit -= kvs.len() as u32;
+                result.append(&mut kvs);
+            }
+            if end_key
+                .as_ref()
+                .map(|ek| ek <= next_key.as_ref() && !ek.is_empty())
+                .unwrap_or(false)
+                || next_key.is_empty()
+            {
                 break;
+            } else {
+                current_key = Some(next_key);
             }
         }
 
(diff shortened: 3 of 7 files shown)


src/raw/client.rs
```diff
@@ -756,57 +761,42 @@ impl<PdC: PdClient> Client<PdC> {
...
-                })??;
-        let mut cur_limit = limit;
-
-        while cur_limit > 0 {
-            let request = new_raw_scan_request(
-                cur_range.clone(),
-                cur_limit,
+        let mut current_limit = limit;
+        let (start_key, end_key) = range.clone().into_keys();
+        let mut current_key: Option<Key> = Some(start_key);
```
pingyu, line 770: It seems that `current_key` will never be `None`.

src/raw/client.rs
```diff
@@ -819,6 +809,61 @@ impl<PdC: PdClient> Client<PdC> {
...
+    #[async_recursion]
+    async fn retryable_scan(
+        &self,
+        mut scan_args: ScanInnerArgs,
+    ) -> Result<(Option<RawScanResponse>, Key)> {
+        let start_key = match scan_args.start_key {
+            None => return Ok((None, Key::EMPTY)),
+            Some(ref sk) => sk,
+        };
+        let permit = scan_args.permits.acquire().await.unwrap();
```
pingyu, line 821: `permit` seems to be not necessary as there is no multi-threading.

src/raw/client.rs
```diff
@@ -756,57 +761,42 @@ impl<PdC: PdClient> Client<PdC> {
...
+                .map(|r| r.kvs.into_iter().map(Into::into).collect::<Vec<KvPair>>())
+                .unwrap_or(Vec::new());
+
+            if !kvs.is_empty() {
+                current_limit -= kvs.len() as u32;
+                result.append(&mut kvs);
+            }
+            if end_key
+                .as_ref()
+                .map(|ek| ek <= next_key.as_ref() && !ek.is_empty())
```
pingyu, line 793: `ek.is_empty()` seems to be not necessary. `ek.is_empty()` is valid and it means that the end key is empty (but not unbounded), although it's meaningless.

src/raw/client.rs
```diff
@@ -819,6 +809,61 @@ impl<PdC: PdClient> Client<PdC> {
...
+        let resp = self.do_store_scan(store.clone(), request).await;
+        drop(permit);
+        match resp {
+            Ok(mut r) => {
+                if let Some(err) = r.region_error() {
+                    let status =
+                        plan::handle_region_error(self.rpc.clone(), err.clone(), store.clone())
+                            .await?;
+                    return if status {
+                        self.retryable_scan(scan_args.clone()).await
```
pingyu, line 840: Suggest to eliminate the recursion and let caller do the retry, to reduce overhead.

pingyu (commented):
Thanks for your contribution !

The PR overall looks good. And I left some minor suggestions.

Besides, If you would like to verify the correctness when region error happens, consider to use failpoint. Please refer to [`failpoint_test.rs`](https://github.com/tikv/client-rust/blob/master/tests/failpoint_tests.rs). It's not a must before the PR is accepted.

src/raw/client.rs
limbooverlambda: Good point. Removed the Option wrapper for the start key.

src/raw/client.rs
limbooverlambda: Removed

src/raw/client.rs
limbooverlambda: Removed.

limbooverlambda:
Hi @pingyu, are there any more changes you need me to make before this change can be merged? Thanks for taking the time to look at this.

src/raw/client.rs
```diff
@@ -953,4 +1016,63 @@ mod tests {
         );
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_raw_scan_retryer() -> Result<()> {
```
pingyu, line 1021: This test case looks a little not necessary to me, as it just check that we call error handler on region error, but introduce more complexity.

The core changes in this PR is the retry on region error, so if we want to verify the correctness, I think we need to simulate such scene.

For example, we first put some little entries, and perform the scan to fill the region cache. Then we put much more (or large) entries to trigger the region split (see `tests::common::init()`), and scan again. The later scan should meet region errors.

The test case would be a little complex to implement, so it's OK to me if we don't have it in this PR.

src/raw/client.rs
```diff
@@ -755,57 +759,45 @@ impl<PdC: PdClient> Client<PdC> {
...
+
+        let region_error_handler =
+            |pd_rpc_client: Arc<PdC>, err: errorpb::Error, store: RegionStore| {
+                Box::pin(plan::handle_region_error(pd_rpc_client, err, store))
+            } as _;
+        while current_limit > 0 {
+            let scan_args = ScanInnerArgs {
+                start_key: current_key.clone(),
+                range: range.clone(),
+                limit,
```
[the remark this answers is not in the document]
limbooverlambda: Good catch. Changed.

src/raw/client.rs
```diff
@@ -755,57 +759,45 @@ impl<PdC: PdClient> Client<PdC> {
...
+
+            let mut kvs = res
+                .map(|r| r.kvs.into_iter().map(Into::into).collect::<Vec<KvPair>>())
+                .unwrap_or(Vec::new());
+
+            if !kvs.is_empty() {
+                current_limit -= kvs.len() as u32;
+                result.append(&mut kvs);
+            }
+            if end_key
```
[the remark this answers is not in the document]
limbooverlambda: Thanks. Changed. `end_key` had to be cloned however.

src/raw/client.rs
```diff
@@ -755,57 +759,45 @@ impl<PdC: PdClient> Client<PdC> {
...
+        let mut current_key: Key = start_key;
+
+        let region_error_handler =
+            |pd_rpc_client: Arc<PdC>, err: errorpb::Error, store: RegionStore| {
+                Box::pin(plan::handle_region_error(pd_rpc_client, err, store))
+            } as _;
+        while current_limit > 0 {
+            let scan_args = ScanInnerArgs {
+                start_key: current_key.clone(),
+                range: range.clone(),
```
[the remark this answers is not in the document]
limbooverlambda: I was trying to trace through the logic and from what I understand, we will only be looping if the limit of the scan has not been exhausted. So regardless of a split or merge, won't we be resuming the next scan from the end_key returned by the previous scan call? So if the first scan result returns an end_key "foo", doesn't the system guarantee that if we start the next scan starting from "foo", we are guaranteed to return all results that are lexicographically larger than "foo" and smaller than whatever end_key has been provided. This is regardless of whether the underlying regions are undergoing any churn(due to any splits or merges). There may be gaps in my understanding so will be more than happy to get some more feedback here.

src/raw/client.rs
limbooverlambda: Good point. The current setup, to your point, is a bit cumbersome. I have removed this test case. I will add the test as you outlined in a subsequent PR.

limbooverlambda:
@pingyu Thanks for your feedback. It will be great if you could take another look.

limbooverlambda:
Hi @pingyu , the integration tests pass locally. Not sure why they failed on CI. Please let me know if you need me to make any other change to this PR.

src/raw/client.rs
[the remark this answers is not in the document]
pingyu: `end_key.as_ref().is_some_and(|ek| ek <= &next_key)` should work.

src/raw/client.rs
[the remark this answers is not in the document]
pingyu: > I was trying to trace through the logic and from what I understand, we will only be looping if the limit of the scan has not been exhausted. So regardless of a split or merge, won't we be resuming the next scan from the end_key returned by the previous scan call? So if the first scan result returns an end_key "foo", doesn't the system guarantee that if we start the next scan starting from "foo", we are guaranteed to return all results that are lexicographically larger than "foo" and smaller than whatever end_key has been provided. This is regardless of whether the underlying regions are undergoing any churn(due to any splits or merges). There may be gaps in my understanding so will be more than happy to get some more feedback here.

You are right. But I think the issue is here in `retryable_scan`:

```rust
        let request = new_raw_scan_request(
            scan_args.range.clone(),
            scan_args.limit,
            scan_args.key_only,
            scan_args.reverse,
            self.cf.clone(),
        );
```

As the start key in `scan_args.range` is not the end key of previous loop.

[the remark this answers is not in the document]
pingyu: @limbooverlambda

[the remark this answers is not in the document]
limbooverlambda: Hi @pingyu. Apologies for not getting to this comment earlier, got caught up with some other deliverables. From the way I understand it, the retryable scan may not need to change the ranges since it is simply retrying the same ranges after the cache has been invalidated. The caller of retryable_scan should be changing the ranges as new regions are encountered. Let me know if that makes sense. I have however tried to simplify the logic using the loop and gotten rid of the pinned future.

zhangjinpeng87:
@limbooverlambda Please click the "Details" item on the right side of failed CI to check why CI failed. For example:
```
 Compiling tikv-client v0.3.0 (/home/runner/work/client-rust/client-rust)
error: unexpected `cfg` condition value: `protobuf-codec`
   --> src/transaction/requests.rs:907:12
    |
907 | #[cfg_attr(feature = "protobuf-codec", allow(clippy::useless_conversion))]
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `default`, `integration-tests`, and `prometheus`
    = help: consider adding `protobuf-codec` as a feature in `Cargo.toml`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: `-D unexpected-cfgs` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unexpected_cfgs)]`

error: could not compile `tikv-client` (lib test) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
make: *** [Makefile:26: integration-test] Error 101
Error: Process completed with exit code 2.
```

Makefile
```diff
@@ -17,7 +17,7 @@ generate:
 check: generate
 	cargo check --all --all-targets --features "${ALL_FEATURES}"
 	cargo fmt -- --check
-	cargo clippy --all-targets --features "${ALL_FEATURES}" -- -D clippy::all
+	cargo clippy --fix --allow-dirty --all-targets --features "${ALL_FEATURES}" -- -D clippy::all
```
pingyu, line 20: It's not a good idea to add `--fix --allow-dirty`, as the fix will be done in CI only and not applied to code base. Besides, some automatic change may not be correct, we need to have a check.

I will open a PR in these few days to fix the error introduced by the latest toolchain.

limbooverlambda: Thanks @pingyu . It seems like the toolchain upgrade started flagging the autogenerated docs. `--fix --allow-dirty` is a workaround to make the build go through. I did't want to pin the toolchain to a certain version, so this was a quick way to bypass the clippy errors. I agree with you on not using `--fix --allow-dirty`. Do you have any suggestion on how I can get this PR merged? Looks like all the tests are passing.

pingyu: > It's not a good idea to add `--fix --allow-dirty`, as the fix will be done in CI only and not applied to code base. Besides, some automatic change may not be correct, we need to have a check.
> 
> I will open another PR in these few days to fix the errors introduced by the latest toolchain.

The check errors for latest toolchain will be fixed by #460.

src/raw/client.rs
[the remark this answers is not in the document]
pingyu: `retryable_scan` should scan from `current_key`.

Assume that we scan from `key000` to `key100`. And there are two regions: region1: `key000` ~ `key050`, region2: `key050` ~ `key100`.

First iteration, we scan kvs from region1.
Second iteration, `current_key = key050`, but this time we get an error of `EpochNotMatch` because region1 is just merged to region2, and now region2 is `key000` ~ `key100`.
Then we retry, and we should scan from `current_key`, which is `key050`, instead of `key000`. Otherwise, we will get duplicated kvs of frist iteration.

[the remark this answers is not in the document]
limbooverlambda: If you look at the logic, I am. scanning from current_key in the retry instead of the start_key. Let me break down the logic:

The very first call is from the start key:

`let mut current_key: Key = start_key;` //line 166

Then we try scanning from this start key to scan:

```
let scan_args = ScanInnerArgs {
                start_key: **current_key.clone(),** // This is mutated for every iteration
                range: range.clone(),
                limit: current_limit,
                key_only,
                reverse,
                backoff: backoff.clone(),
            };
let (res, next_key) = self.retryable_scan(scan_args).await?; //This is retried with the **current state of scan_args**

```
In the retryable_scan method we are encapsulating the calls to the scan for the start_key (which is the **current_key**), now in this call if we encounter the EpochNotMatch, we will be retrying from **current_key**, not from the start_key. The retry is taking place within the function. So, no duplicate will be returned in this scenario. 

If you scroll further down in `scan_inner`, we change the current_key with the end_key returned by the `retryable_scan` function.

```
 if end_key.clone().is_some_and(|ek| ek <= next_key) {
                break;
            } else {
              current_key = next_key;
   }
```
I understand your concern but given the logic, I am having a hard time seeing how duplicates will be returned. Maybe a good way to address your concern will be to write a test that injects a fault and then change the region mappings in the mock pd and see if the test passes.

src/raw/client.rs
```diff
@@ -818,6 +800,58 @@ impl<PdC: PdClient> Client<PdC> {
...
+
+        let region = self.rpc.clone().region_for_key(&start_key).await?;
+        let store = self.rpc.clone().store_for_id(region.id()).await?;
+        let request = new_raw_scan_request(
+            scan_args.range.clone(),
+            scan_args.limit,
+            scan_args.key_only,
+            scan_args.reverse,
+            self.cf.clone(),
+        );
```
limbooverlambda, line 817: @pingyu  The problem might be over here, we are not retrieving the new regions or stores once it is updated. This may need to be moved into the loop.

src/raw/client.rs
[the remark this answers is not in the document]
pingyu: But the `start_key` (or `from`) we pass to `new_raw_scan_request` through `range` argument is not the `scan_args.start_key`, but the `start_key` from the very beginning, the `range` argument of `scan_inner`.

```rust
            let request = new_raw_scan_request(
                scan_args.range.clone(),
                scan_args.limit,
                scan_args.key_only,
                scan_args.reverse,
                self.cf.clone(),
            );
```

Cargo.toml
```diff
@@ -14,6 +14,7 @@ prometheus = ["prometheus/push", "prometheus/process"]
 # Enable integration tests with a running TiKV and PD instance.
 # Use $PD_ADDRS, comma separated, to set the addresses the tests use.
 integration-tests = []
+protobuf-codec = []
```
pingyu, line 17: This feature can be removed. We don't have it.

src/raw/client.rs
```diff
@@ -818,6 +800,58 @@ impl<PdC: PdClient> Client<PdC> {
...
+        &self,
+        mut scan_args: ScanInnerArgs,
+    ) -> Result<(Option<RawScanResponse>, Key)> {
+        let start_key = scan_args.start_key;
+
+        loop {
+            let region = self.rpc.clone().region_for_key(&start_key).await?;
+            let store = self.rpc.clone().store_for_id(region.id()).await?;
+            let request = new_raw_scan_request(
+                scan_args.range.clone(),
```
hbisheng, line 813: I agree with @pingyu and IIUC the problem is that `scan_args.range` never changes across iterations. 

`scan_args.start_key` is totally fine; it moves forward as the scan makes progress. But since `scan_args.range` doesn't change, it's possible for `retryable_scan` to have `scan_args.range=(key000, key100) scan_args.start_key=key050`. In this case, if we encounter a region merge, the raw scan request constructed here will scan the new region from key000 again, getting duplicate KVs. 

I think one way to fix it is to gradually shrink the current range in the loop of `scan_inner`, similar to what the old code did:
```
cur_range = BoundRange::new(
    std::ops::Bound::Included(region_store.region_with_leader.range().1),
    cur_range.to,
);
```

limbooverlambda: Yes, now I see it. The scan request is being constructed on the range but the range itself is not being changed for every iteration. So essentially, mutating the range with the updated start_key will fix the issue. Thanks @hbisheng and @pingyu for being so patient with the explanation. Adding the fix.

Cargo.toml
limbooverlambda: Done

src/raw/client.rs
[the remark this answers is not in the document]
limbooverlambda: Thanks @pingyu. Now I see where the issue is. Made the change.

src/raw/client.rs
```diff
@@ -864,6 +898,16 @@ impl<PdC: PdClient> Client<PdC> {
     }
 }
 
+#[derive(Clone)]
+struct ScanInnerArgs {
+    start_key: Key,
+    range: BoundRange,
```
pingyu, line 904:
```suggestion
    end_key: Option<Key>,
```

As `start_key` and `range.from` are always the same, I think we can keep only one from them. And as it's easier to convert `start_key` and `end_key` to `range` than opposite, I suggest to have `start_key` and `end_key` in `ScanInnerArgs`.

Rest LGTM.

limbooverlambda: Changed! Thanks for the review.

src/raw/client.rs
```diff
@@ -755,57 +758,37 @@ impl<PdC: PdClient> Client<PdC> {
...
+
+            if !kvs.is_empty() {
+                current_limit -= kvs.len() as u32;
+                result.append(&mut kvs);
+            }
+            if end_key.clone().is_some_and(|ek| ek <= next_key) {
                 break;
+            } else {
+                current_key = next_key;
+                range = BoundRange::new(std::ops::Bound::Included(current_key.clone()), range.to);
```
pingyu, line 791:
```suggestion
```

pingyu (approved):
LGTM~

pingyu:
> /cc @iosmanthus

PTAL~

iosmanthus (approved):

pingyu:
@limbooverlambda Thanks for your contribution !

PR #459, opened 2024-06-17
Resolving the TLS url issue

limbooverlambda:
While running the client we are running into issues where the client fails to connect to a TLS enabled PD endpoint with the error:

[WARN] [config_logging.go:287] ["rejected connection"] [remote-addr="[addr]:port"] [server-name=] [error="tls: first record does not look like a TLS handshake"] 

Digging into the code, the problem seems to stem from the following logic in the client (

[client-rust/src/common/security.rs](https://github.com/tikv/client-rust/blob/c6110dd8087a4d9e7bf48da77bbbfafa79614b10/src/common/security.rs#L80)

Line 80 in [c6110dd](https://github.com/tikv/client-rust/commit/c6110dd8087a4d9e7bf48da77bbbfafa79614b10)

```
 let addr = "http://".to_string() + &SCHEME_REG.replace(addr, ""); 
)
let addr = "http://".to_string() + &SCHEME_REG.replace(addr, "");

let mut builder = Channel::from_shared(addr)?
    .tcp_keepalive(Some(Duration::from_secs(10)))
    .keep_alive_timeout(Duration::from_secs(3));
```

Here SCHEME_REG is the regex to find "https" in the prefix of the address (

[client-rust/src/common/security.rs](https://github.com/tikv/client-rust/blob/c6110dd8087a4d9e7bf48da77bbbfafa79614b10/src/common/security.rs#L19)

Line 19 in [c6110dd](https://github.com/tikv/client-rust/commit/c6110dd8087a4d9e7bf48da77bbbfafa79614b10)

```
 lazy_static::lazy_static! { 
).
lazy_static::lazy_static! {
    static ref SCHEME_REG: Regex = Regex::new(r"^\s*(https?://)").unwrap();
}

```
This is a problem during the ssl handshake since the server is not acknowledging the requests sent to an addressed prefixed with "http" instead of "https". Adding the logic to handle these variations.

### DIFF
--- a/src/common/security.rs
+++ b/src/common/security.rs
@@ -8,10 +8,10 @@ use std::time::Duration;
 
 use log::info;
 use regex::Regex;
-use tonic::transport::Certificate;
 use tonic::transport::Channel;
 use tonic::transport::ClientTlsConfig;
 use tonic::transport::Identity;
+use tonic::transport::{Certificate, Endpoint};
 
 use crate::internal_err;
 use crate::Result;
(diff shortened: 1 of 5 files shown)


pingyu:
@limbooverlambda 

Thanks for your contribution !

Please fix the [error](https://github.com/tikv/client-rust/actions/runs/9556494802/job/26346594122?pr=459) of "make check" in CI. 

Rest LGTM~

limbooverlambda:
> @limbooverlambda
> 
> Thanks for your contribution !
> 
> Please fix the [error](https://github.com/tikv/client-rust/actions/runs/9556494802/job/26346594122?pr=459) of "make check" in CI.
> 
> Rest LGTM~

Thanks @pingyu . I fixed the issues with the formatting.

pingyu (approved):
LGTM~

PR #470, opened 2024-12-01
fix: make prometheus dep optional

johnbchron:
Right now the `prometheus` dependency is present regardless of the `prometheus` feature flag. This PR excludes the `prometheus` dependency when the flag is not present.


src/pd/retry.rs
```diff
@@ -75,15 +76,22 @@ impl<Cl> RetryClient<Cl> {
...
         let mut last_err = Ok(());
         for _ in 0..LEADER_CHANGE_RETRY {
             let res = $call;
 
+            #[cfg(feature = "prometheus")]
             match stats.done(res) {
                 Ok(r) => return Ok(r),
                 Err(e) => last_err = Err(e),
             }
+            #[cfg(not(feature = "prometheus"))]
```
pingyu, line 90: It looks a little too invasive to me.

How about like this:

Define another `DummyRequestStats` with no member and empty methods. Then change `pd_stats` and `tikv_stats` to return coresponding object accroding to the feature.

johnbchron: Sounds good. I'll get to it this evening.

Issue #493, opened 2025-06-06
“Connection refused ” caused by stopping tikv process when the tikv is marked as offline status


BryceCao:
we mark a tikv as offline status by sending a pd api for delete the store, its a normal opreation as shrinking tikv nodes. Then, we kill the tikv process to simulate hardware damage, in a client which is start before shrinking operation, a scan api fail with print "gRPC api error: status: Unavailable, message: "error trying to connect: tcp connect error: Connection refused (os error 111)", details: [], metadata: MetadataMap { headers: {} }". But ,in a new start client, a same scan api can return correct result. Other way, when we start the tikv process again, the old client can return the correct result too.

BryceCao:
I found that directly killing a TiKV process can also trigger this phenomenon. I suspect it is related to the leaders on the killed TiKV node — the client may have incorrectly accessed the killed TiKV instance.

BryceCao:
> I found that directly killing a TiKV process can also trigger this phenomenon. I suspect it is related to the leaders on the killed TiKV node — the client may have incorrectly accessed the killed TiKV instance.

Some start_marker scan can return normally. I suspect that this bug is triggered only when the Region being accessed has its leader located on the TiKV instance that was killed.

BryceCao:
https://github.com/tikv/client-rust/pull/495, i have submitted a PR, please review it.

PR #495, opened 2025-07-03
add handle operation toward rpc err in scan_inner

BryceCao:
fix https://github.com/tikv/client-rust/issues/493


src/request/plan.rs
```diff
@@ -345,6 +345,18 @@ pub(crate) async fn handle_region_error<PdC: PdClient>(
     }
 }
 
+pub(crate) async fn handle_rpc_error<PdC: PdClient>(
```
pingyu, line 348: I think it would be better to name as `invalidate_cache` for wider usage.

It would be nice if you can use this method for `handle_region_error` & `handle_other_error` to eliminate duplicated codes. But it is still OK if you do not want to change them in this PR.

pingyu (commented):
Thanks for you contribution !

Please fix the DCO error.

Issue #289, opened 2021-05-11
Synchronous API


ekexium:
Synchronous API is easier for new users to get started but can be less efficient.

The easiest way to do it is simply `block_on` each async API.

- [ ] Raw client #301 
- [ ] Transactional client
- [ ] Tests and examples

xuhui-lu:
Hey @ekexium , is this issue still open? I could help with it.

ekexium:
> Hey @ekexium , is this issue still open? I could help with it.

Yes it's open. Thank you!

eduralph:
Shipped via #517 — `SyncTransactionClient` / `SyncTransaction` / `SyncSnapshot`. Can this be closed?

PR #517, opened 2026-01-16
Add sync API for transactional client

odecode:
For issue #289  a synchronous API for transactional client.

I have programmed a fair bit but new to OSS and Rust. I consulted AI heavily for this to try and get an understanding of what this part of the project is about and how it works, but I tried using it a teacher rather than just "vibe coding".

## Summary by CodeRabbit

* **New Features**
  * Exposed blocking synchronous transaction and snapshot APIs that provide full, blocking access to transactional and read-only operations alongside existing async interfaces.
* **Bug Fixes**
  * Clear user-facing error when attempting to use the synchronous client from within an active async runtime.
* **Tests**
  * Extensive integration tests covering sync client usage, transactions, snapshots, scans, edge cases, and context validation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ reqwest = { version = "0.11", features = ["json", "native-tls-vendored"] }
 rstest = "0.18.2"
 serde_json = "1"
 serial_test = "0.5.0"
-tempfile = "3.6"
+tempfile = "=3.14.0"
 tokio = { version = "1", features = ["sync", "rt-multi-thread", "macros"] }
 
 [[test]]
--- a/proto-build/Cargo.toml
+++ b/proto-build/Cargo.toml
@@ -13,3 +13,5 @@ edition = "2021"
 [dependencies]
 glob = "0.3"
 tonic-build = { version = "0.10", features = ["cleanup-markdown"] }
+# Keep this compatible with rust-toolchain 1.84.1.
+tempfile = "=3.14.0"


odecode:
Thank you for your patience so far! I have made the review's suggested change.

tests/sync_transaction_tests.rs
```diff
@@ -0,0 +1,392 @@
...
+use tikv_client::Value;
+
+/// Helper to initialize and return a sync client
+fn sync_client() -> Result<SyncTransactionClient> {
+    SyncTransactionClient::new_with_config(pd_addrs(), Config::default().with_default_keyspace())
+}
+
+#[test]
+#[serial]
+fn sync_txn_get_timestamp() -> Result<()> {
```
pingyu: CI will not cover test cases with prefix `sync_txn_`. Please see the `Makefile`. Suggest to name as `txn_sync_`.

tests/common/mod.rs
```diff
@@ -74,6 +74,14 @@ pub async fn init() -> Result<()> {
     Ok(())
 }
 
+pub fn init_sync() -> Result<()> {
```
pingyu: I think we can simply implement as: `tokio::runtime::Builder::new_current_thread().enable_all().build()?.block_on(init())`.

As we can guarantee that this function must be the first of the test cases and never be called inside a tokio runtime.

It would be even better if we add a comment about the usage limitation.

odecode:
Is it possible for me to request review from Copilot myself? Then I could address all its concerns and not bother you on every iteration.

odecode:
added extra test coverage for the sync snapshot public api surface, and error handling to prevent panic when sync client is called from within an async runtime.

src/transaction/sync_transaction.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Rollback the transaction.
+    pub fn rollback(&mut self) -> Result<()> {
+        self.runtime.block_on(self.inner.rollback())
+    }
+
+    /// Send a heart beat message to keep the transaction alive.
+    pub fn send_heart_beat(&mut self) -> Result<u64> {
+        self.runtime.block_on(self.inner.send_heart_beat())
+    }
```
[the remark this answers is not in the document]
odecode: this is fixed

src/common/errors.rs
```diff
@@ -128,6 +128,12 @@ pub enum Error {
     KeyspaceNotFound(String),
     #[error("Transaction not found error: {:?}", _0)]
     TxnNotFound(kvrpcpb::TxnNotFound),
+    /// Attempted to create or use the sync client (including calling its methods) from within a Tokio async runtime context
```
pingyu, line 131: `Error::NestedRuntimeError`’s display string already wraps an inner `{0}` with guidance (`src/common/errors.rs:132`), but `check_nested_runtime()` also stores a long, guidance-heavy string (`src/transaction/sync_client.rs:30`). This likely produces duplicated advice and punctuation (e.g., “... any Tokio runtime.. Use TransactionClient ...”), and makes future message changes harder.

odecode: I changed it so that the guidance is only in Error::NestedRuntimeError, but left the {0} parameter there so that potential callers may add additional information if needed.

pingyu:
cc @iosmanthus

pingyu:
The build failure will be fixed by #532 .

pingyu:
@odecode Thanks for your contribution !

Issue #516, opened 2026-01-14
tests: failpoint_tests::txn_cleanup_2pc_locks is flaky


pingyu:
CI link: https://github.com/tikv/client-rust/actions/runs/20982618570/job/60310631494?pr=515

PR: https://github.com/tikv/client-rust/pull/515

```
Nextest run ID 1784fdea-b79b-49f5-ae6b-ac103c96cca0 with nextest profile: ci
    Starting 26 tests across 4 binaries (55 tests skipped)
        SLOW [> 60.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [>120.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [>180.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [>240.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [>300.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [>360.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [>420.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [>480.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [>540.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
 TERMINATING [>600.000s] (─────) tikv-client::failpoint_tests txn_cleanup_2pc_locks
     TIMEOUT [ 601.103s] ( 1/26) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        PASS [   9.742s] ( 2/26) tikv-client::failpoint_tests txn_cleanup_async_commit_locks
        PASS [   1.373s] ( 3/26) tikv-client::failpoint_tests txn_cleanup_locks_batch_size
        PASS [   5.370s] ( 4/26) tikv-client::failpoint_tests txn_cleanup_range_async_commit_locks
        PASS [   6.477s] ( 5/26) tikv-client::failpoint_tests txn_optimistic_heartbeat
        PASS [   1.349s] ( 6/26) tikv-client::integration_tests txn_bank_transfer
        PASS [   0.524s] ( 7/26) tikv-client::integration_tests txn_batch_mutate_optimistic
[...]

PR #523, opened 2026-01-20
flaky test wait for lock cleanup

odecode:
Related to issue #516 . I experienced different types of symptoms: Sometimes an assertion failed (assertion that remaining locks == 0), sometimes the assertion passed but there was panic or slowness elsewhere, most times the test passed fine.

My knowledge of the projects techniques is still a bit limited, so I consulted AI, but I try to use my best judgement in evaluating whether its suggestions make sense. It suggested that the test may experience a race condition regarding to how quickly the locks are cleaned up in a 2PC, and that we could wait for the lock cleanup process to let the background process finish and then check the assertions.

## Summary by CodeRabbit

* **Tests**
  * Improved lock count verification test assertions to enhance test reliability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tests/failpoint_tests.rs
+++ b/tests/failpoint_tests.rs
@@ -337,7 +337,8 @@ async fn txn_cleanup_2pc_locks() -> Result<()> {
         )
         .await?;
         let keys = write_data(&client, false, false).await?;
-        assert_eq!(count_locks(&client).await?, 0);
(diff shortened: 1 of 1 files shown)


odecode:
This was already fixed in master (same mechanism as here) so closing this PR

PR #527, opened 2026-01-25
raw: fix reverse_scan failing across multiple regions

### LINKED ISSUE #512: Reverse scan fail to find key when range spans multi region

HueCodes:
## Summary

Fix `reverse_scan` returning `key_not_in_region` error when the scan range spans multiple TiKV regions.

**Root cause:** `retryable_scan` selected the region based on `start_key` (lower bound), but TiKV reverse scan starts from the upper bound and goes backward. When bounds are in different regions, TiKV returns an error because the request context specifies the wrong region.

**Fix:**
- Select region based on `end_key` (upper bound) for reverse scans
- Return `region.start_key()` as continuation point (moving toward lower keys)
- Handle edge case when iteration lands exactly on a region boundary
- Add safety check to prevent infinite loops

Fixes #512

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo clippy --lib --tests` passes with no warnings
- [x] `cargo test --lib` passes (53 tests)
- [ ] Integration test with multi-region TiKV cluster (requires TiUP setup)

### DIFF
--- a/src/raw/client.rs
+++ b/src/raw/client.rs
@@ -873,14 +873,21 @@ impl<PdC: PdClient> Client<PdC> {
             if reverse {
                 let region_start = region.start_key();
                 if !region_start.is_empty() && *region_lookup_key == region_start {
-                    // Find the previous region by looking up a key just before this boundary.
-                    // Truncating the last byte gives a lexicographically smaller key.
-                    let mut prev_key: Vec<u8> = region_start.into();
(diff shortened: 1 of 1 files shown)


HueCodes:
Thanks for the feedback. Addressing these issues, I will fix and post and update today.

HueCodes:
Let me know if theres any other changes you would like me to make.

PR #538, opened 2026-06-03
fix: handle case where PD cluster has no leader

MananShukla7:
## Problem

In `try_connect_leader`, the previous leader is accessed with `.unwrap()`:

```rust
let previous_leader = previous.leader.as_ref().unwrap();
```

This panics when the PD cluster temporarily has no leader — a situation that occurs during a rolling restart of PD pods in a Kubernetes environment. During leader election, `GetMembersResponse.leader` can be `None`, causing the client to crash rather than returning a recoverable error.

## Fix

Replace `.unwrap()` with `.ok_or_else()` to return a descriptive `Err` instead of panicking. This is consistent with the style already used in this file.

**Before:**
```rust
let previous_leader = previous.leader.as_ref().unwrap();
```

**After:**
```rust
let previous_leader = previous
    .leader
    .as_ref()
    .ok_or_else(|| internal_err!("PD cluster has no leader"))?;
```

The same pattern is applied to the `resp.leader` lookup later in the function, replacing the existing `ok_or_else` error message with a clearer one.

## Reproduction

Trigger a rolling restart of PD pods while the TiKV client is connected. The client panics at `previous.leader.as_ref().unwrap()` when the new `GetMembersResponse` arrives with `leader: None` during leader election.

## Checklist

- [x] `cargo fmt --all`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test`
- [x] DCO sign-off (`git commit -s`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cluster connection logic to gracefully handle cases with no leader, returning a clear error instead of panicking, improving stability and observability.
* **Documentation**
  * Added explanatory comments describing leader-connection behavior and its non-panicking error semantics.

### DIFF
--- a/src/pd/cluster.rs
+++ b/src/pd/cluster.rs
@@ -283,6 +283,11 @@ impl Connection {
         }
     }
 
+    /// Attempts to connect to the PD cluster leader.
+    ///
+    /// Iterates over known members to find a responsive node, then connects
+    /// to the reported leader. Returns an error if no leader is present or
+    /// reachable.
     async fn try_connect_leader(
         &self,
         previous: &pdpb::GetMembersResponse,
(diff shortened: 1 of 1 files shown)


pingyu (approved):

iosmanthus (approved):
LGTM

pingyu:
@MananShukla7 Thanks for your contribution !

PR #544, opened 2026-07-05
transaction: resolve orphaned locks whose primary was never written

### LINKED ISSUE #543: Orphaned secondary lock is never resolved: reads fail with TxnNotFound forever (regression in #519)
## Summary

The `rollback_if_not_exist` escalation in `LockResolver::get_txn_status_from_lock` — the mechanism that recovers an **orphaned lock whose primary transaction was never written** — is unreachable. `check_txn_status` matches `Error::ExtractedErrors`, but its plan (`retry_multi_region → merge(CollectSingle) → extract_error`) can only deliver a per-key error as `Error::MultipleKeyErrors`. As a result, any key left under such an orphaned lock becomes **permanently unreadable and unwritable** through every API except `cleanup_locks`, even long after the lock's TTL expires.

This appears to be a regression introduced by #519 (`7d80f59`, "transaction: Resolve locks using kv_resolve_lock interface..."). The previous code sent a legacy `Cleanup` on the primary, which rolls back a missing primary unconditionally; #519 replaced it with the `CheckTxnStatus` flow whose escalation arm never fires for this error shape.

## Environment

- client-rust `master` @ `e53837d`
- TiKV / PD `v8.5.5`, single node, api-v1, no keyspace

## How the orphan arises (no API misuse required)

Prewrite requests are dispatched concurrently per region. A transaction whose keys span two regions can have its **secondary**'s prewrite land while its **primary**'s prewrite fails (lost write-write race) or never runs (client crash mid-prewrite). The result is a lock on the secondary key naming a primary that has no lock and no commit/rollback record — exactly the state `rollback_if_not_exist` exists to recover. Note this also arises from a plain crash between the parallel prewrites, with no client code running afterward.

## Reproduction

Using the public transactional API against a cluster where keys `P` and `S` fall in **different regions** (the orphan needs the two prewrites to be separate per-region batches; within one region the prewrite fails atomically and leaves nothing):

```rust
// P sorts first, S last; ensure a region boundary between them.
let mut txn = client.begin_optimistic().await?;
txn.lock_keys(vec![P.clone()]).await?;   // P becomes the primary
txn.put(S.clone(), b"orphaned".to_vec()).await?;

// Invalidate P from another transaction so `txn`'s prewrite on P loses:
let mut racer = client.begin_optimistic().await?;
racer.put(P.clone(), b"newer".to_vec()).await?;
racer.commit().await?;

let err = txn.commit().await;            // Err(WriteConflict) — S's prewrite already landed
drop(txn);                               // no rollback — i.e. a crash

// Now S is poisoned:
let read = another_txn.get(S).await;     // Err, forever — see below
```

**Observed** — every read of `S` (immediately, and long after the ~3s lock TTL expires):

```
MultipleKeyErrors([KeyError(KeyError {
    txn_not_found: Some(TxnNotFound { start_ts: <txn's ts>, primary_key: <P> }), .. })])
```

Only `TransactionClient::cleanup_locks` over the range recovers `S`.

## Root cause (file:line, `master` @ e53837d)

1. The read hits the lock and enters resolution: `resolve_locks` (`src/transaction/lock.rs:51`) → `get_txn_status_from_lock` (`:527`, `rollback_if_not_exist = false`) → `check_txn_status` (`:426`).
2. TiKV's CheckTxnStatus on the missing primary returns `KeyError { txn_not_found }`. `single_shard_handler` takes the key error and returns `Err(Error::MultipleKeyErrors(..))` (`src/request/plan.rs:220-222`). Because the error is already stripped from the response, the downstream `ExtractError` just propagates it (`plan.rs:842`) — `Error::ExtractedErrors` (built only at `plan.rs:844/846`) is impossible in this plan shape.
3. `check_txn_status`'s conversion arm matches only `Err(Error::ExtractedErrors(..))` (`lock.rs:469-479`) to turn `txn_not_found` into `Error::TxnNotFound`. `MultipleKeyErrors` falls through `Err(err) => return Err(err)` (`lock.rs:480`) unconverted.
4. `get_txn_status_from_lock`'s retry loop only special-cases `Err(Error::TxnNotFound)` (`lock.rs:565-573`) — the branch that, once the TTL has expired, sets `rollback_if_not_exist = true` and retries (writing the rollback record). It never runs; the raw error escapes (`lock.rs:589`).
5. No layer above re-enters with different state; every fresh read repeats identically. `cleanup_locks` heals only because it calls `check_txn_status` with `rollback_if_not_exist = true, current_ts = u64::MAX` (`lock.rs:325-327`).

For contrast, three sibling call sites place `.extract_error()` **before** `.merge(..)`, where `HasKeyErrors for Result<T,E>` (`src/store/errors.rs:163-174`) re-wraps into `ExtractedErrors` — so `send_heart_beat`, `check_all_secondaries`, and `commit_primary` convert correctly. `check_txn_status` alone has the adapter order swapped relative to its own match arm.

## Proposed fix

Accept both wrappers in `check_txn_status` (`src/transaction/lock.rs:469`):

```rust
Err(Error::ExtractedErrors(mut errors) | Error::MultipleKeyErrors(mut errors)) => {
    match errors.pop() {
        Some(Error::KeyError(key_err)) => {
            if let Some(txn_not_found) = key_err.txn_not_found {
                return Err(Error::TxnNotFound(txn_not_found));
            }
            return Err(Error::KeyError(key_err));
        }
        Some(err) => return Err(err),
        None => unreachable!(),
    }
}
```

Validated against this repro: the poisoned key self-heals after TTL expiry (no `cleanup_locks` needed); pre-TTL reads return a correctly-typed `Err(TxnNotFound)` after backoff. No regressions in the unit suite or in an integration suite covering CAS races, write-skew locking, pessimistic waits, and scans. (An equivalent fix reorders `check_txn_status`'s plan to the heartbeat shape, `.extract_error().merge(CollectSingle)`. Please don't unify the wrappers globally at `plan.rs:222` — `MultipleKeyErrors` feeds other merge/collect logic and public error surfaces.)

Happy to open a PR if the approach looks right.

## Impact

Any client crash between parallel per-region prewrites — or any failed commit not followed by an explicit `rollback()` — can leave keys permanently unreadable for all readers. A contributing factor: `Committer::commit` propagates a prewrite failure without cleaning up already-placed prewrite locks (unlike client-go, whose `twoPhaseCommitter.execute` runs a deferred best-effort `BatchRollback` on commit failure), so even orderly failed commits produce this state unless the caller rolls back — and `PessimisticRollback` does not clear prewrite locks, so pessimistic transactions can orphan even with that discipline.

## Workaround

Run `cleanup_locks(range, &safepoint, options)` periodically or on `TxnNotFound` read failures; always `rollback()` after a failed `commit()`.

## Related

- #519 (the regressing change / current resolution baseline), #528 (resolve-lock gaps for async-commit locks), #497 (`txn_not_found` seen in the wild), #315, #313.

eduralph:
check_txn_status matched only ExtractedErrors, but its plan delivers per-key errors as MultipleKeyErrors, so the Rollback_if_not_exist escalation in get_txn_status_from_lock was unreachable and an orphaned secondary lock poisoned its key permanently. Accept both wrappers.

Closes #543

Signed-off-by: Eduard Ralph <eduard@ralphovi.net>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction status handling when multiple errors are returned.
  * Transaction-not-found conditions are now reported consistently.
  * Expired locks with missing primary records are now properly rolled back and cleaned up.
* **Tests**
  * Added coverage for expired-lock cleanup and verified that lock resolution occurs exactly once.


eduralph:
@ekexium to raise awareness

ekexium (approved):

eduralph:
@pingyu - we're 1 LGTM away from getting this merged

pingyu:
How about replacing `ExtractedError` by reusing `MultipleKeyErrors` and introducing a new `MultipleRegionErrors` ?

It seems that the other interfaces have the same issue, and it could introduce potential bugs into future code.

p.s. Removing `ExtractedError` is a breaking change, but I think the next release will be a semver-major release anyway, since we have bumped the Rust toolchain and the protos upgrade is WIP.

eduralph:
Agreed, and I think the diagnosis is sharper than "the other interfaces have the same issue" — the two variants aren't just redundant, which one you get is decided by **plan shape**, invisibly, at a distance.

`ExtractedErrors` is produced only by `ExtractError`, and it carries *either* key errors *or* region errors. `MultipleKeyErrors` is produced by `single_shard_handler` and always carries key errors. Which one reaches the caller depends on whether a merge sits between `retry_multi_region` and `extract_error`: without one, `ExtractError` sees the per-shard `Vec<Result<_>>` and normalizes to `ExtractedErrors`; with `CollectSingle` (or `Collect`), the merge pops the single `Result` out first, so `MultipleKeyErrors` becomes the plan's own error and `ExtractError` re-raises it without ever calling `key_errors()`. That is precisely the bug this PR fixes.

That coupling is already visible in the public raw API, for the same class of failure:

| method | plan shape | error the user sees |
|---|---|---|
| `delete`, `put_with_ttl` | `retry_multi_region` → `merge(CollectSingle)` → `extract_error` | `MultipleKeyErrors` |
| `batch_delete`, `batch_put_with_ttl`, `delete_range` | `retry_multi_region` → `extract_error` | `ExtractedErrors` |

So a caller matching `Error::ExtractedErrors` to inspect key errors handles `batch_delete` and silently misses `delete`. Nothing about either signature suggests that.

On the internal side I went through the four places that match `ExtractedErrors` today. Only `LockResolver::check_txn_status` (this PR) was actually broken; `resolve_lock_with_retry`, `commit_primary_with_retry` and `CleanupLocks::execute` are correct — but only because their chains happen to have no collapsing merge. None of them is correct *by construction*. Adding a `merge(CollectSingle)` to any of those chains would silently flip the variant and break a distant `match` arm with no compile error and no test failure. `commit_primary_with_retry` is the one I'd least like to discover that way: it matches on `commit_ts_expired` to retry with a newer commit ts, so a silent variant change turns a recoverable commit into a failed one.

So yes to `MultipleKeyErrors` + `MultipleRegionErrors`, and dropping `ExtractedErrors`. Splitting by *what the payload is* instead of *which plan combinator produced it* is what makes the match arms verifiable — and it kills a whole class of "worked until someone recomposed the plan" bugs. Agreed too that a semver-major is the moment to do it.

One request on sequencing: could this land as a follow-up rather than in this PR? This one is a two-line match-arm widening for a live bug — an orphaned lock whose primary was never written currently poisons its key forever — and it's already approved. The refactor is a breaking change touching the error enum, both producer sites, and every consumer, and it deserves its own review rather than riding in on a bug fix.

Happy to open that PR — I'd plan to introduce `MultipleRegionErrors`, split `ExtractError`'s two branches across the two variants, update the four consumers, and add a test per public entry point pinning which variant it surfaces so plan recomposition can't silently change it again. Say the word and I'll put it up.

eduralph:
Put it up as #557 (draft, stacked on this branch) so it is concrete rather than hypothetical.

It removes `ExtractedErrors`, reports key errors as `MultipleKeyErrors` and region errors as a new
`MultipleRegionErrors`, and collapses this PR's two-variant arm back to one. It also drops a latent
bug in `CleanupLocks::execute`, which popped the last error to type-test it and on the key-error path
assigned the *remaining* vec to `key_error` — so a lone key error was dropped and `key_errors()`
reported `Some(vec![])`, an error carrying no errors.

One thing worth your eye, since it is the least obvious part: `HasKeyErrors for Result<T, Error>`
reports *any* per-shard `Err` as a key error, so `ExtractError` now wraps only what the response
itself reported (`KeyError` from a `kvrpcpb::KeyError` field, `KvError` from a raw endpoint's string
`error` field) and propagates transport or exhausted-retry errors unchanged. Testing for `KeyError`
alone would have left the raw operations plan-shape-dependent — the very thing the split removes.

Marked draft because it is stacked; happy to rebase it onto master once this one lands, or to fold it
in here instead if you would rather have them together.

eduralph:
@pingyu  - changes implemented as requested, also fiddled to get the tests to run clean.

PR #548, opened 2026-07-06
transaction: add contextual logging with start_ts (#267)

eduralph:
Enrich the transaction lifecycle, plan/retry, and region-cache logging with each transaction's start_ts and region/store context; add commit/rollback outcome logs, log dropping an active transaction under all check levels, and a terminal warn when region-error retries are exhausted. Truncate keys in the one full-key error log. Fix message typos/duplicates and remove dead logger-threading comments. Add a dependency-free capture-logger test and a RUST_LOG note to the README.

Refs #267

Signed-off-by: Eduard Ralph <eduard@ralphovi.net>

## Summary by CodeRabbit

* **Documentation**
  * Added a new Logging section covering how logging is handled, how to choose a backend, `RUST_LOG`, and how log levels correspond to transaction lifecycle and operational/error events.
* **Bug Fixes**
  * Fixed misleading debug output for reverse key scans.
* **Improvements**
  * Expanded transaction lifecycle/2PC logging with start timestamps and more timing/lock context; refined region error, retry, and cache invalidation logs.
  * Improved logged key prefixes by hex-encoding and truncating for readability.
* **Tests**
  * Added log-capture assertions to verify lifecycle logs include the transaction start timestamp.

### DIFF
--- a/src/transaction/lock.rs
+++ b/src/transaction/lock.rs
@@ -14,6 +14,7 @@ use tokio::time::sleep;
 use crate::backoff::Backoff;
 use crate::backoff::DEFAULT_REGION_BACKOFF;
 use crate::backoff::OPTIMISTIC_BACKOFF;
+use crate::kv::HexRepr;
 use crate::pd::PdClient;
 
 use crate::proto::kvrpcpb;
@@ -39,7 +40,7 @@ use crate::Result;
 
 pub(crate) fn format_key_for_log(key: &[u8]) -> String {
     let prefix_len = key.len().min(16);
-    format!("len={}, prefix={:?}", key.len(), &key[..prefix_len])
+    format!("len={}, prefix={}", key.len(), HexRepr(&key[..prefix_len]))
 }
 
 /// _Resolves_ the given locks. Returns locks still live. When there is no live locks, all the given locks are resolved.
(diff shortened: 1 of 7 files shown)


src/transaction/lock.rs
```diff
@@ -37,7 +37,7 @@ use crate::transaction::requests::TransactionStatusKind;
 use crate::Error;
 use crate::Result;
 
-fn format_key_for_log(key: &[u8]) -> String {
+pub(crate) fn format_key_for_log(key: &[u8]) -> String {
```
pingyu, line 41: How about use `kv::HexRepr` to format the `key[..prefix_len]` for more readable ?

pingyu (commented):
Rest LGTM.

src/transaction/lock.rs
eduralph: Good Point. 

{:?} on a byte slice was rendering the prefix as a decimal array (prefix=[104, 101, 108, 108, 111]); with HexRepr it now reads prefix=68656C6C6F. This also puts format_key_for_log back in line with how Key and KvPair render bytes — my earlier commit had swapped out the last HexRepr use in transaction.rs. Added two unit tests covering the hex encoding and the 16-byte truncation, neither of which had coverage before.

pingyu (approved):

iosmanthus (approved):

PR #535, opened 2026-04-27
request: replace try_join_all with JoinSet in plan handlers

pingyu:
## Summary

- Replace `futures::future::try_join_all` with `tokio::task::JoinSet` in `single_plan_handler` and `single_store_handler`
- Extract result collection into a reusable `collect_join_set_results` helper that:
  - Preserves spawn order via index tagging
  - Logs and propagates `JoinError` with structured context
  - Calls `shutdown().await` on the `JoinSet` before returning on error
- On shard iteration failure in `single_plan_handler`: abort already-spawned tasks via `shutdown().await`
- Pin dependency versions (`indexmap`, `icu_*`, `proptest`) for `rust-toolchain 1.84.1` compatibility

## Changes

- `src/request/plan.rs`:
  - Added `tokio::task::JoinSet` import; removed `futures::future::try_join_all`
  - Added `log::error` import for join failure logging
  - Introduced `collect_join_set_results()` — generic ordered result collector for `JoinSet<(usize, T)>`
  - Both `single_plan_handler` and `single_store_handler` now spawn index-tagged tasks via `JoinSet`
  - On `JoinError`: logs handler name and task count, shuts down remaining tasks, returns `Error::JoinError`
  - On shard stream `Err`: calls `shutdown().await` before returning, aborting in-flight tasks
  - Added unit test `test_join_set_results_keep_spawn_order` verifying result ordering
- `Cargo.toml` / `proto-build/Cargo.toml`: pinned `indexmap`, `icu_*` crates, and `proptest` for Rust 1.84.1 compatibility

## Review Findings

- https://github.com/pingyu/client-rust/blob/join-all-handles/src/request/plan.rs#L113-L114 and https://github.com/pingyu/client-rust/blob/join-all-handles/src/request/plan.rs#L157-L158: 🟡 risk: `join_set.shutdown().await` changes the old error path from "drop `JoinHandle`s and let already-spawned tasks continue detached" to "abort remaining shard/store tasks immediately." If any caller relies on in-flight fanout requests finishing after a later sharding/join failure, this is now a behavior change. Add a regression test for cancellation semantics or document that the new contract is intentional.

Validation: `cargo check --lib` passed, `cargo test --lib` passed (59 tests), and the new `test_join_set_results_keep_spawn_order` also passed. No other concrete correctness issues stood out in the final branch state.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal async task handling for multi-region and all-stores execution with stronger error handling and ordered result collection.
* **Tests**
[...]

PR #530, opened 2026-02-06
chore - Bump MSRV/toolchain to stable 1.93 and fix strict clippy

mingley:
## Summary

This PR upgrades the repository toolchain/MSRV to Rust `1.93.0` and applies strict-clippy compatibility fixes while preserving runtime behavior and public API compatibility.

Quality objective:
- keep the codebase clean under `-D warnings`,
- avoid unintended semantic changes,
- maintain `Error::GrpcAPI` usability for downstream code.

### Problem statement
Moving to Rust 1.93 and strict clippy surfaced new lints in generated/protocol and request/transaction paths. A prior approach introduced awkward conditional typing for `Error::GrpcAPI`; this PR keeps the API representation stable and solves lint pressure in a clearer way.

### Design changes
- Toolchain/MSRV:
  - `Cargo.toml`: `rust-version = "1.93.0"`
  - `rust-toolchain.toml`: channel `1.93.0`
- `Error::GrpcAPI` compatibility:
  - keep payload as `tonic::Status` (single representation),
  - keep `From<tonic::Status> for Error` conversion path,
  - remove `cfg(clippy)` payload-type split.
- Lint strategy for large error payload:
  - targeted `#![allow(clippy::result_large_err)]` in `src/lib.rs` and `examples/raw.rs`.
- Clippy cleanup (behavior-preserving simplifications):
  - remove redundant conversions,
  - replace `repeat(...).take(...)` with `repeat_n(...)`,
  - narrow generated-code dead-code handling in `src/proto.rs`.

### Risk analysis
- Primary risk: accidental API/behavior drift during lint cleanup.
- Mitigation:
  - explicit tests for error conversion and variant behavior,
  - full format/clippy/test gates.

## File Scope

- `Cargo.toml`
- `rust-toolchain.toml`
- `src/common/errors.rs`
- `src/lib.rs`
- `examples/raw.rs`
- `src/proto.rs`
- `src/pd/retry.rs`
- `src/store/request.rs`
- `src/transaction/requests.rs`
- `src/transaction/transaction.rs`
- `tests/integration_tests.rs`

## Testing Done

Executed locally:

1. `cargo fmt` -> pass
2. `cargo clippy --all-targets --all-features -- -D warnings` -> pass
3. `cargo test` -> pass

Added/maintained regression coverage in `src/common/errors.rs`:
- `from_tonic_status_produces_grpc_api_error`
- `grpc_api_variant_accepts_tonic_status_payload`
- `from_proto_region_error_wraps_region_error`
[...]

PR #549, opened 2026-07-14
request: replace try_join_all with JoinSet in plan handlers

### LINKED ISSUE #534: RetryableMultiRegion: try_join_all drops JoinHandles on error, detaching spawned tasks that panic on runtime shutdown
### Problem

When using tikv-client in a context where the Tokio runtime may be dropped (e.g., a short-lived runtime created inside a worker thread), the program may panic with:

```
A Tokio 1.x context was found, but it is being shutdown.
```

### Backtrace

The panic originates from `TimerEntry::poll_elapsed` in `tokio-1.26.0/src/runtime/time/entry.rs:550-551`:

```
TimerEntry::poll_elapsed                          ← PANIC: runtime timer driver is shut down
  ← tokio::time::sleep::Sleep::poll               ← sleep() called inside task
    ← tonic::transport::grpc_timeout::ResponseFuture::poll  ← gRPC timeout
      ← tonic AddOrigin service call
        ← tower buffer future
          ← tikv_client::ScanRequest::dispatch
            ← tikv_client::KvRpcClient::dispatch
              ← Dispatch<Req>::execute
                ← ResolveLock<P,PdC>::execute
                  ← RetryableMultiRegion::single_shard_handler  ← spawned via tokio::spawn
                    ← tokio multi_thread worker (async task)
```

### Scenario

The user intends to:

1. Create a short-lived Tokio runtime.
2. Create a `TransactionClient` on that runtime.
3. Perform scan operations via `snapshot.scan()` within `runtime.block_on()`.
4. Eventually drop the runtime when the worker thread exits.

During scan operations, tikv-client internally spawns tasks on the current runtime to shard the request across multiple regions. If one of these tasks errors and the user's runtime is short-lived enough to be dropped before the spawned tasks naturally terminate, the process panics.

### Root Cause

In `request/plan.rs`, `RetryableMultiRegion::single_plan_handler` spawns concurrent per-shard handler tasks via `tokio::spawn` and waits for them using `futures::future::try_join_all`:

```rust
// plan.rs:122-136
for shard in shards {
    let handle = tokio::spawn(Self::single_shard_handler(...));
    handles.push(handle);
}
let results = try_join_all(handles).await?;
```

`try_join_all` (from `futures-util 0.3.28`) **cancels remaining futures immediately when any one future returns an error** (source: `try_join_all.rs` lines 165-168 break on first error, line 182 drops remaining futures). This means:

1. When one `single_shard_handler` task fails (e.g., gRPC timeout, region error, leader-not-found), `try_join_all` drops the remaining `JoinHandle`s.
2. Tokio's `JoinHandle::Drop` does **not** cancel the spawned task — it detaches the task, which continues running on the runtime.
3. If the runtime is dropped while detached tasks are still alive, the tasks' subsequent `tokio::time::sleep` calls (via tonic's gRPC timeout wrapper or tikv-client's retry backoff) panic because the timer driver is shut down.

The same pattern exists in `RetryableAllStores::single_store_handler` (plan.rs:463).

### Proposed Fix

Instead of `try_join_all`, use `futures::future::join_all` to await all spawned tasks to completion before propagating errors:

```rust
let results = join_all(handles).await;
let mut err = None;
let mut outputs = Vec::with_capacity(results.len());
for r in results {
    match r.unwrap() { // JoinError only on panic
        Ok(ok) => outputs.push(Ok(ok)),
        Err(e) if err.is_none() => err = Some(e),
        _ => {}
    }
}
if let Some(e) = err {
    return Err(e);
}
Ok(outputs)
```

Alternatively, use `tokio::task::JoinSet` which provides `join_all()` semantics and tracks all spawned tasks.

eduralph:
Closes #534. Supersedes #535 — offered so the fix isn't blocked on rebase logistics. It carries @pingyu's two commits **unchanged, authorship and sign-off preserved**, plus one follow-up commit.

**Why superseding:** #535's merge conflict consists entirely of its dependency-pin commit for rust-toolchain 1.84.1, obsolete since #530 moved the toolchain to 1.93. With that commit dropped, the change rebases onto master with no conflicts. @pingyu — if you'd rather rebase #535 itself, say the word and I'll close this one.

**The follow-up commit:** raises tokio's floor to 1.21 (`JoinSet` stabilized there; `tokio = "1"` permitted resolutions that can't compile this crate), applies `cargo fmt`, and adds the doc comment the #535 review asked for — stating the contract that an error return means no further effects: remaining tasks are aborted, never detached (the old detach-on-error could complete stray writes after an observed failure, beyond the #534 panic).

**Verification:** `cargo test --lib` 67/67 (including the spawn-order test); clippy + fmt clean at 1.93; and a differential parity harness ([getwyrd/client-rust-test](https://github.com/getwyrd/client-rust-test)) driving client-go and client-rust through identical multi-region scenarios shows no wire-visible behavior change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for concurrent execution across multiple shards or stores.
  * Ensured results are returned in the same order as tasks were started.
  * Added safer handling for task failures, cancellations, and panics, preventing partial/incorrect aggregation.

* **Chores**
  * Updated the Tokio runtime dependency to 1.21.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ serde_derive = "1.0"
 serde_json = "1"
 take_mut = "0.2.2"
 thiserror = "1"
-tokio = { version = "1", features = ["sync", "rt-multi-thread", "macros"] }
+tokio = { version = "1.21", features = ["sync", "rt-multi-thread", "macros"] }
 tonic = { version = "0.12", features = ["tls", "gzip"] }
 
 [dev-dependencies]
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -94,6 +94,35 @@ pub(crate) fn is_grpc_error(e: &Error) -> bool {
     matches!(e, Error::GrpcAPI(_) | Error::Grpc(_))
 }
 
+async fn collect_join_set_results<T>(
+    mut join_set: JoinSet<(usize, T)>,
+    task_count: usize,
+    handler_name: &str,
+) -> Result<Vec<T>>
+where
+    T: Send + 'static,
+{
+    let mut results = (0..task_count).map(|_| None).collect::<Vec<_>>();
+    while let Some(join_result) = join_set.join_next().await {
+        match join_result {
+            Ok((idx, val)) => results[idx] = Some(val),
+            Err(e) => {
+                error!(
+                    "{}: failed to join task ({} tasks): {}",
+                    handler_name, task_count, e
+                );
+                join_set.shutdown().await;
+                return Err(Error::JoinError(e));
+            }
+        }
+    }
+
+    Ok(results
+        .into_iter()
+        .map(|result| result.expect("all spawned tasks should return a result"))
+        .collect())
+}
+
 pub struct RetryableMultiRegion<P: Plan, PdC: PdClient> {
     pub(super) inner: P,
     pub pd_client: Arc<PdC>,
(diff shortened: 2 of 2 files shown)


eduralph:
For the record, the earlier `integration-test-txn` failure was the pre-existing flake documented in #516 — same test (`txn_cleanup_2pc_locks`), same signature (hang until nextest's terminate kills it), observed there in January on #515 before this change existed. Mechanically, this PR only changes behavior on the join-failure and shard-iteration-error paths; that test's `return`-style failpoints are handled inside the per-shard tasks, so the success-path wait is identical to the old `try_join_all` — a hang in one would hang in both. It also passes repeatedly (3/3, <1s each) against a local api-v2 cluster on this branch, and the retriggered run is now green across the board.

pingyu (approved):
LGTM~

pingyu:
cc @iosmanthus

iosmanthus (approved):
Thanks a lot for picking this up and carrying @pingyu's commits forward with authorship preserved — that's a really considerate way to unblock the fix. 🙏

The change is well-reasoned and correct:
- Replacing `try_join_all`-over-`JoinHandle`s with `JoinSet` + `shutdown().await` properly fixes the detach-on-error behavior behind #534 — remaining tasks are now aborted-and-joined instead of racing on after the failure.
- Tagging each task with its spawn index and reassembling in order is the right call: it preserves parity with `try_join_all` and, importantly, keeps multi-region scan results in key order (a naive JoinSet migration collecting in completion order would have introduced a subtle ordering bug). The `test_join_set_results_keep_spawn_order` test guards this nicely.
- The `tokio = "1.21"` floor bump is the correct minimum for `JoinSet`/`shutdown`, and the doc comment documenting the abort-not-detach contract is excellent.

LGTM 👍

One non-blocking suggestion for a follow-up: the added test only covers the happy ordering path. A cheap, high-value addition would be a test that spawns a panicking task and asserts `collect_join_set_results` returns `Err(Error::JoinError(_))`, since the error path is the one this PR exists to fix. Not a blocker — happy to see this merge as-is.

PR #550, opened 2026-07-15
*: re-vendor kvproto at b41e863 (2026-06-22) and record the vendored revision

eduralph:
**Why:** the vendored protos are an unrecorded late-2023/early-2024 vintage (`errorpb.proto` /
`pdpb.proto` untouched since #324, one hand-patched addition from #519). Nothing records which
kvproto revision they came from, and the staleness now blocks real features.

**What:** all vendored protos + the include tree re-vendored at pingcap/kvproto `b41e863` — the
revision client-go pins today — plus a new **`proto/VERSION`** recording it. Regenerated with the
existing tonic-build pipeline (21 files); one mechanical source fix (`pdpb.RequestHeader` gained
`caller_component`/`caller_id`). Newly available surface includes `KvFlush` / `KvBufferBatchGet`
(pipelined DML), `HealthFeedback`, and `pdpb.BatchScanRegions`.

**Scope.** Per @pingyu's review, this PR is now the re-vendor and nothing else. The two new
protocol semantics that become *visible* with the refresh are split into follow-ups stacked on this
branch — #555 (`errorpb.UndeterminedResult`) and #556 (shared locks). Neither is a
regression introduced here: an undetermined result falls through to the generic region-error arm and
retries (the same default the follow-up keeps for non-commit-point plans), and a shared-lock wrapper
carries no key bytes on the wire whichever proto vintage parses it, so the keyspace hazard predates
this PR.

**Verification:** `make check` green (including a clean re-run of the proto-build pipeline — the
committed bindings match what it generates); 67 lib tests; txn/raw/failpoint integration suites green
against a local api-v2 cluster; and behaviour compared against client-go by driving both clients
through identical transactional and raw scenarios — no observable change.


pingyu:
Suggest to split the handling of `errorpb.UndeterminedResult` and shared locks into separate PRs.

So that this current PR focuses only on re-vendor kvprotos.

eduralph:
> Suggest to split the handling of `errorpb.UndeterminedResult` and shared locks into separate PRs.

Done — thanks, that's the right call. #550 is now the re-vendor and nothing else (59 files: the
protos, the regenerated bindings, `proto/VERSION`, and one mechanical fix where
`pdpb.RequestHeader` gained `caller_component`/`caller_id`).

The two semantics are split out into #555 (undetermined result) and #556 (shared locks),
both stacked on this branch.

Note the branch is now rebased onto master rather than carrying your `Merge branch 'master'` commit —
the split needed a rewrite, so the merge is gone but nothing it brought in is: #547 is in the base,
and your resolution of `transaction.rs` and mine came out identical. Thanks for the sync.

Worth stating explicitly, since it's the thing that would justify keeping them together: **neither
follow-up is fixing a regression that #550 introduces.** I checked both paths on the re-vendor-only
branch:

- `UndeterminedResult` falls through to the generic region-error arm, which backs off and retries.
  That is exactly the default #555 keeps for every plan that isn't a commit point — it only
  adds terminal treatment for raw CAS, the primary commit, and async-commit/1PC prewrites.
- A shared-lock wrapper leaves `primary_lock` unset on the wire whichever proto vintage parses it
  (TiKV's `SharedLocks::into_lock_info` sets only `lock_type`, `key` and `shared_lock_infos`), and
  the old bindings truncate that field just as unconditionally — so the `truncate_keyspace` panic
  #556 addresses predates this PR. The refresh only makes the wrapper *identifiable*.

So #550 is safe to merge on its own timeline.

pingyu (approved):
LGTM~

pingyu:
cc @YuJuncen @iosmanthus

YuJuncen (approved):

PR #553, opened 2026-07-22
tests: log phase boundaries in the cleanup tests to diagnose the CI hang (#516)

eduralph:
`failpoint_tests::txn_cleanup_2pc_locks` intermittently stalls in CI until nextest's 600s cap kills it (#516; `txn_cleanup_async_commit_locks` has hit the same thing). A killed test reports no timings, so the logs currently show only `SLOW → TERMINATING → TIMEOUT` with no indication of *which* step stopped making progress.

I tried to reproduce it and could not: single-node and 3-node api-v2 clusters (matching CI's `--kv 3` with `config/tikv.toml` / `config/pd.toml`), cold-started, and CPU-constrained to roughly a GitHub runner's budget — ~20 runs, all completing in 1–2s with every phase in milliseconds. So rather than guess at a fix, this PR makes the next occurrence self-explaining.

## What it adds (test-only)

- A `phase!` marker around each significant step of the two tests that have hung. The **entry** line is the point: when the test is killed, the last `phase >` line names the step that never finished. The exit line reports the duration, distinguishing "slow" from "stuck".
- `ensure_region_split` now logs its elapsed time and final region count, reusing the count its polling loop already fetched — diagnostics must not add a fallible request that could itself stall setup. That cost is charged to whichever test runs first, and that is the test CI kills, so this separates "setup was slow" from "the body stalled".

No client (`src/`) change, no assertion or control-flow change. CI already runs with `RUST_LOG=info`, so the markers appear without any workflow change.

Once a CI run captures a stall with these markers, the offending phase should be obvious and a real fix can follow.

## Testing

Full failpoint suite green (8/8) against a local 3-node multi-region api-v2 cluster; `make check` clean; verified the markers render as intended under `RUST_LOG=info`.

Note: this touches the same two test functions as #551, so whichever merges second will need a trivial rebase — happy to handle that.

## Summary by CodeRabbit

* **Tests**
  * Improved diagnostic logging for region setup checks, including elapsed time and the final observed region count.
  * Added timed phase markers to failpoint and integration cleanup tests, making CI test progress and failures easier to trace.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -113,8 +113,10 @@ async fn ensure_region_split(
 
     info!("splitting regions...");
     let start_time = std::time::Instant::now();
+    let mut observed_regions;
     loop {
-        if ctl::get_region_count().await? as u32 >= region_count {
+        observed_regions = ctl::get_region_count().await? as u32;
+        if observed_regions >= region_count {
             break;
         }
         if start_time.elapsed() > REGION_SPLIT_TIME_LIMIT {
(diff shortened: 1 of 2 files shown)


eduralph:
@pingyu - another ping for awareness. This is purely for diagnostic reasons

pingyu (approved):
LGTM~

pingyu:
cc @iosmanthus @YuJuncen

YuJuncen (approved):

PR #554, opened 2026-07-24
feat(transaction): add paginated range scanner

ariesdevil:
## Summary

- Add async `Transaction::scanner` and `Snapshot::scanner` APIs and expose the public `Scanner` type.
- Fetch the first non-empty page when the scanner is created, then scan TiKV one region at a time in fixed batches of 256 pairs.
- Merge remote pairs with buffered mutations through a linear mutation cursor:
  - buffered puts and inserts override remote values;
  - buffered deletes suppress matching keys.
- Keep additional memory bounded to one remote batch and one local mutation.
- Avoid inserting scanned remote pairs into the transaction read cache.
- Invalidate the scanner after EOF or an error, matching client-go iterator semantics.
- Handle empty and inverted ranges without issuing scan RPCs.
- Relocate only the current region when retrying after a region split or region error.

This provides behavior comparable to client-go's `KVTxn.Iter` and `KVSnapshot.Iter` without materializing the full range or fanning a scan out across every overlapping region.

## Tests

- Eager initial fetch and lazy pagination of subsequent pages.
- Empty and inverted ranges.
- Construction-time and subsequent-page errors.
- Scanner invalidation after EOF or an error.
- Sequential scanning across empty and non-empty regions.
- Region relocation after a split.
- Buffered put, insert, and delete overlays with and without keyspace encoding.
- Fixed RPC batch limits and read-cache isolation.

## Summary by CodeRabbit

* **New Features**
  * Added unbounded range scanning for both snapshots and transactions via a new lazy scanner that fetches results in batches until completion.
  * Introduced a public scanner API for iterating over key-value pairs.
  * Exposed the scanner type as part of the public transaction surface.
* **Tests**
  * Added/updated tests for the transaction start timestamp.
  * Added coverage for unbounded scan pagination, validating batch size, ordering, and returned values.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -164,6 +164,8 @@ pub use crate::transaction::Client as TransactionClient;
 #[doc(inline)]
 pub use crate::transaction::ProtoLockInfo;
 #[doc(inline)]
+pub use crate::transaction::Scanner;
+#[doc(inline)]
 pub use crate::transaction::Snapshot;
 #[doc(inline)]
 pub use crate::transaction::SyncSnapshot;
--- a/src/transaction/transaction.rs
+++ b/src/transaction/transaction.rs
@@ -40,6 +40,9 @@ use crate::KvPair;
 use crate::Result;
 use crate::Value;
 
+mod scanner;
+pub use scanner::Scanner;
+
 /// An undo-able set of actions on the dataset.
 ///
 /// Create a transaction using a [`TransactionClient`](crate::TransactionClient), then run actions
--- a/src/transaction/transaction/scanner.rs
+++ b/src/transaction/transaction/scanner.rs
@@ -1,0 +1,1099 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::cmp::Ordering;
+use std::sync::atomic;
+use std::sync::atomic::AtomicU8;
+use std::sync::Arc;
+
+use log::debug;
+use tokio::time::sleep;
+
+use super::Transaction;
+use super::TransactionStatus;
+use crate::backoff::Backoff;
+use crate::pd::PdClient;
+use crate::pd::PdRpcClient;
+use crate::proto::pdpb::Timestamp;
+use crate::region::RegionWithLeader;
+use crate::request::plan::handle_region_error;
+use crate::request::plan::is_grpc_error;
+use crate::request::EncodeKeyspace;
+use crate::request::KeyMode;
+use crate::request::Keyspace;
+use crate::request::Plan;
+use crate::request::PlanBuilder;
+use crate::request::RetryOptions;
+use crate::request::TruncateKeyspace;
+use crate::store::HasKeyErrors;
+use crate::store::HasRegionError;
+use crate::transaction::buffer::MutationIterator;
+use crate::transaction::lowering::new_scan_request;
+use crate::BoundRange;
+use crate::Error;
+use crate::Key;
+use crate::KvPair;
+use crate::Result;
+use crate::Value;
+
+impl<PdC: PdClient> Transaction<PdC> {
+    /// Create a paginated transactional scanner over a range, the equivalent
+    /// of client-go's `KVTxn.Iter(start, end)`.
+    ///
+    /// The returned [`Scanner`] fetches region-sequentially: each page is one
+    /// scan request clamped to a single region and limited to
+    /// `SCANNER_BATCH_SIZE` pairs. Construction fetches the first non-empty
+    /// page, later pages are fetched only as the scanner is advanced.
+    /// Fetched pairs are not inserted into the transaction's read
+    /// cache; beyond the transaction's existing buffer, the scanner retains at
+    /// most one fetched batch and one buffered mutation.
+    ///
+    /// Construction fetches the first non-empty page, so it fails with an RPC
+    /// or lock-resolution error when that fetch fails. It also fails fast if
+    /// the transaction no longer allows operations (e.g. it is already
+    /// committed or rolled back), or if the range is reversed (start key
+    /// greater than end key): reverse iteration is not supported, use
+    /// [`Transaction::scan_reverse`] for a bounded reverse scan instead.
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// # use tikv_client::{Key, KvPair, Value, Config, TransactionClient};
+    /// # use futures::prelude::*;
+    /// # futures::executor::block_on(async {
+    /// #   let client = TransactionClient::new(vec!["192.168.0.100", "192.168.0.101"]).await.unwrap();
+    /// #   let mut txn = client.begin_optimistic().await.unwrap();
+    /// #   let key1: Key = b"bar".to_vec().into();
+    /// #   let key2: Key = b"foo".to_vec().into();
+    /// #   let mut scanner = txn.scanner(key1..key2).await.unwrap();
+    /// #   while let Some(_pair) = scanner.next().await.unwrap() {
+    ///       // Process the pair...
+    /// #   }
+    /// #   // Finish the transaction...
+    /// #   txn.commit().await.unwrap();
+    /// # });
+    /// ```
+    pub async fn scanner(&mut self, range: impl Into<BoundRange>) -> Result<Scanner<'_, PdC>> {
+        debug!("creating transactional scanner");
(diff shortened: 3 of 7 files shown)


src/transaction/transaction.rs
```diff
@@ -448,6 +461,70 @@ impl<PdC: PdClient> Transaction<PdC> {
...
+            // data may remain; advance past the last returned key with
+            // `next_key()` (the same key+'\x00' successor trick client-go
+            // uses) and fetch again. If the total is an exact multiple of the
+            // batch size, this costs one extra empty page fetch, same as
+            // client-go discovering EOF on its next getData.
+            let full_page = page.len() as u32 == SCAN_UNBOUNDED_BATCH_SIZE;
+            if let Some(last) = page.last() {
+                start = last.key().clone().next_key();
+            }
+            out.extend(page);
```
you06: We can't estimate the data size in a range, so buffer them in memory is danger to me, it can easily cause OOM in the client side.

you06 (commented):
rest LGTM

src/transaction/transaction.rs
ariesdevil: Fixed, aligned to the `client-go`'s behavior, PTAL again.

src/lib.rs
```diff
@@ -164,14 +164,15 @@ pub use crate::transaction::Client as TransactionClient;
...
+pub use crate::transaction::Scanner;
+#[doc(inline)]
 pub use crate::transaction::Snapshot;
 #[doc(inline)]
 pub use crate::transaction::SyncSnapshot;
 #[doc(inline)]
 pub use crate::transaction::SyncTransaction;
 #[doc(inline)]
 pub use crate::transaction::SyncTransactionClient;
-#[doc(inline)]
```
you06, line 174: Is it a mis-deletion?

ariesdevil: fixed

ariesdevil:
Hi @ekexium, any concerns about this PR?

src/transaction/transaction.rs
```diff
@@ -1128,6 +1163,76 @@ const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_millis(MAX_TTL / 2);
...
+        loop {
+            if let Some(pair) = self.cache.next() {
+                return Ok(Some(pair));
+            }
+            if self.eof {
+                return Ok(None);
+            }
+            let page: Vec<KvPair> = self
+                .txn
+                .scan(
```
pingyu: I think we should not reuse `Transaction::scan` here.

That path fans the remaining range out to every overlapping region, applies the limit per region, and caches every merged response before globally truncating it. With 100 populated regions, one logical 256-pair page can fetch and retain up to 25,600 pairs, and repeated pages can grow the transaction buffer with the entire range.

Seems that we need a new non-caching, region-sequential fetch path with a global batch bound.

ekexium (commented):
Mostly LGTM. I was a bit confused by the name "scan_unbounded", how about renaming to something like "iter" or "scanner"?

ariesdevil:
Thank you @ekexium, @pingyu for your patient and careful review; PTAL again.

src/transaction/transaction/scanner.rs
```diff
@@ -0,0 +1,997 @@
...
+    /// #   let key2: Key = b"bar".to_vec().into();
+    /// #   let mut scanner = txn.scanner(key1..key2).await.unwrap();
+    /// #   while let Some(_pair) = scanner.next().await.unwrap() {
+    ///       // Process the pair...
+    /// #   }
+    /// #   // Finish the transaction...
+    /// #   txn.commit().await.unwrap();
+    /// # });
+    /// ```
+    pub async fn scanner(&mut self, range: impl Into<BoundRange>) -> Result<Scanner<'_, PdC>> {
```
ekexium, line 75: If it doesn't support reverse range, we'd better return an error. Also, maybe we should first check whether the operation is allowed.

ekexium (commented):
Looks like the `scan` method doesn't correctly handle `Insert` and `CheckNotExist`. Once this PR lands they would have different behavior. Could you help fix that as well?

ariesdevil:
Yup, I'll fix it

src/transaction/transaction/scanner.rs
ariesdevil: Fixed, and I'll impl `iter_reverse` in next PR.

ariesdevil:
Hi @ekexium , PTAL again.

src/transaction/transaction/scanner.rs
```diff
@@ -0,0 +1,1043 @@
...
+
+        loop {
+            let range: BoundRange = (self.next_start.clone(), self.end.clone()).into();
+            let (enc_start, enc_end) = range.encode_keyspace(keyspace, KeyMode::Txn).into_keys();
+            let enc_start: Vec<u8> = enc_start.into();
+            let enc_end: Vec<u8> = enc_end.map(Into::into).unwrap_or_default();
+
+            // Locate only the first region. On every retry this is rebuilt from
+            // `next_start`, so a split can only select the first child.
+            let mut regions =
```
pingyu: `PdClient::region_for_key` should be enough since we only need the first region.

src/transaction/transaction/scanner.rs
```diff
@@ -0,0 +1,1043 @@
...
+                region_stream_for_range((enc_start, enc_end.clone()), self.rpc.clone());
+            let Some(((span_start, span_end), region)) = regions.next().await.transpose()? else {
+                self.eof = true;
```
pingyu: No region for the range should be considered as error.

src/transaction/transaction/scanner.rs
```diff
@@ -0,0 +1,1043 @@
...
+                continue;
+            }
+            if let Some(errors) = response.key_errors() {
+                return Err(Error::MultipleKeyErrors(errors));
+            }
+
+            let page: Vec<KvPair> = response.pairs.into_iter().map(Into::into).collect();
+            if page.len() as u32 == SCANNER_BATCH_SIZE {
+                let last = page.last().expect("a full page is never empty");
+                self.next_start = last.key().clone().truncate_keyspace(keyspace).next_key();
```
pingyu: `next_start` is possible to equal to `end` ?

ariesdevil:
@pingyu fixed, PTAL

ariesdevil:
The CI failure seems not related to this PR.

src/transaction/transaction/scanner.rs
```diff
@@ -0,0 +1,1092 @@
...
+}
+
+/// The TiKV side of the union iterator.
+///
+/// It owns all RPC, region-retry, and pagination state and stays positioned on
+/// at most one remote pair. The public `Scanner` only needs to compare and
+/// consume that current pair.
+struct RemoteScanner<PdC: PdClient> {
+    status: Arc<AtomicU8>,
+    rpc: Arc<PdC>,
```
pingyu: Nit: `rpc` -> `pd` or `pd_client`.

pingyu (approved):

src/transaction/transaction/scanner.rs
```diff
@@ -0,0 +1,1092 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// # use tikv_client::{Key, KvPair, Value, Config, TransactionClient};
+    /// # use futures::prelude::*;
+    /// # futures::executor::block_on(async {
+    /// #   let client = TransactionClient::new(vec!["192.168.0.100", "192.168.0.101"]).await.unwrap();
+    /// #   let mut txn = client.begin_optimistic().await.unwrap();
+    /// #   let key1: Key = b"foo".to_vec().into();
```
pingyu: The public example scans foo..bar, but foo > bar. scanner() rejects this reversed range, so .await.unwrap() panics at runtime. Swap the example bounds.

pingyu:
@ariesdevil 
It seems that your last commit is incorrect. I see the `region_stream_for_range` come back.

ariesdevil:
> @ariesdevil It seems that your last commit is incorrect. I see the `region_stream_for_range` come back.

Yes, I rebased the wrong branch, I'll fix it soon.

ariesdevil:
Done, PTAL again, and thx for your patient review.

pingyu:
LGTM~

pingyu:
cc @ekexium

ekexium (approved):

PR #399, opened 2023-07-07
Migrate to tonic

andylokandy:
Use tonic as the gRPC service.

Closes https://github.com/tikv/client-rust/issues/397

PR #555, opened 2026-07-25
request: treat an undetermined apply outcome as terminal at commit points

eduralph:
> Stacked on #550 — please review the second commit only. Until #550 merges, the diff below also
> contains its re-vendor.

**Why:** the re-vendored kvproto exposes `errorpb.UndeterminedResult` — the server telling us the
raft apply outcome is UNKNOWN. Today it falls through to the generic region-error arm and is retried
like any transient failure. For most requests that's right; for a commit point it isn't. "Failed"
must never be claimed for a transaction that may already be durable.

**What:** follow client-go's shape. Its transport never retries `UndeterminedResult` ("should not
retry … processed by the caller", `region_request.go`) and each *action* decides. So the plan layer
keeps **retry as the default** — replaying an idempotent request resolves the uncertainty, and on
backoff exhaustion the error escapes *unchanged* so callers can still classify it — while plans for
which a replay is unsafe opt in to terminal-on-first-sight:

- **raw CAS** — a replay would compare against its *own* effect and report `succeed = false` for a
  write that happened.
- **the primary commit** — client-go returns `ErrResultUndetermined` here (`commit.go`) rather than
  retrying.
- **async-commit / 1PC prewrites** — the prewrite *is* the commit point. Stricter than client-go, in
  the safe direction.

**Shard aggregation is fixed alongside:** `collect::<Result<_>>()` returns the first error *by shard
index*, so a lower-index shard's determinate error could mask a higher-index shard's
`UndeterminedResult` and tell the caller the request definitely failed. An undetermined error from
any shard now wins.

Commit paths classify the surfaced error as `UndeterminedError`. Two places are deliberately
conservative and say so in comments: a 1PC prewrite that sharding downgraded to 2PC, and an
undetermined error arriving from the commit plan's `resolve_lock` layer rather than the commit
dispatch. Both only *over*-report uncertainty, which makes the caller verify instead of replay.

**Verification:** 4 new tests — terminal for opted-in plans under a *generous* backoff (proving
terminal, not merely retry-exhausted), preserved unchanged on retry exhaustion, recognizer
positive/negative, and the shard-masking hazard. 71 lib tests; `make check` green; txn/raw/failpoint
integration suites green against a local api-v2 cluster.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction commit handling when the final outcome cannot be determined.
  - Prevented unsafe retries after requests may already have been sent.
  - Preserved and surfaced undetermined-region and dispatch errors.
  - Improved error prioritization when multiple regions report failures.

- **New Features**
  - Added configurable multi-region retry behavior for undetermined results and ambiguous outcomes.
  - Operations can now stop immediately when outcomes are uncertain, providing clearer transaction results.

### DIFF
--- a/src/request/plan.rs
+++ b/src/request/plan.rs
@@ -132,6 +132,14 @@ where
         .collect())
 }
 
+/// Did the server say the request's outcome is UNKNOWN — e.g. a raft timeout where the
+/// apply result was never observed (`errorpb.UndeterminedResult`)? A commit receiving
+/// this must be reported as undetermined: reporting plain failure would invite the
+/// caller to retry effects that may already be durable.
+pub(crate) fn is_undetermined_region_error(e: &Error) -> bool {
+    matches!(e, Error::RegionError(re) if re.undetermined_result.is_some())
+}
+
 pub struct RetryableMultiRegion<P: Plan, PdC: PdClient> {
     pub(super) inner: P,
     pub pd_client: Arc<PdC>,
@@ -141,6 +149,30 @@ pub struct RetryableMultiRegion<P: Plan, PdC: PdClient> {
     /// If true, return Ok and preserve all regions' results, even if some of them are Err.
     /// Otherwise, return the first Err if there is any.
     pub preserve_region_results: bool,
+
+    /// Terminal treatment of `errorpb.UndeterminedResult` (an unknown raft apply
+    /// outcome). client-go's transport never retries it and each ACTION decides
+    /// (region_request.go: "should not retry ... processed by the caller"); its one
+    /// terminal action is the primary non-async commit (commit.go: returns
+    /// ErrResultUndetermined). Plans opt in when a replay could produce a WRONG
+    /// RESULT (raw CAS replaying against its own effect) or when the request is a
+    /// commit point and a later, different retry error must not overwrite the
+    /// uncertainty (primary commit; async/1PC prewrite — stricter than client-go,
+    /// in the safe direction). Everything else retries: re-applying an idempotent
+    /// request resolves the uncertainty, and on backoff exhaustion the error
+    /// escapes UNCHANGED for the caller to classify.
+    pub terminal_on_undetermined: bool,
+
+    /// Terminal treatment of a DISPATCH-stage gRPC error, for the request whose
+    /// replay is not idempotent with respect to its own result: raw CAS. Once the
+    /// request may have been sent, a lost response is as ambiguous as
+    /// `errorpb.UndeterminedResult` — a replay would compare against the first
+    /// attempt's own write and could report `succeed = false` for a write that
+    /// happened. Sharding/connection errors still retry (the request was never
+    /// sent), and commit points do NOT set this: replaying a commit is idempotent —
+    /// it can only resolve the uncertainty — and client-go's transport likewise
+    /// retries RPC errors there (region_request.go, onSendFail).
+    pub terminal_on_dispatch_error: bool,
 }
 
 impl<P: Plan + Shardable, PdC: PdClient> RetryableMultiRegion<P, PdC>
@@ -155,6 +187,8 @@ where
         backoff: Backoff,
         permits: Arc<Semaphore>,
         preserve_region_results: bool,
+        terminal_on_undetermined: bool,
+        terminal_on_dispatch_error: bool,
     ) -> Result<<Self as Plan>::Result> {
         let shards = current_plan.shards(&pd_client).collect::<Vec<_>>().await;
         let shards_len = shards.len();
@@ -182,6 +216,8 @@ where
                         backoff,
                         permits,
                         preserve_region_results,
+                        terminal_on_undetermined,
+                        terminal_on_dispatch_error,
                     )
                     .await,
                 )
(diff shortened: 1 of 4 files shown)


src/request/plan.rs
```diff
@@ -1062,4 +1128,172 @@ mod test {
...
+        // later different error must not overwrite the uncertainty. A GENEROUS
+        // backoff proves the error is terminal, not merely retry-exhausted.
+        let plan = RetryableMultiRegion {
+            inner: UndeterminedPlan,
+            pd_client: Arc::new(MockPdClient::default()),
+            backoff: Backoff::no_jitter_backoff(1, 2, 10),
+            preserve_region_results: false,
+            terminal_on_undetermined: true,
+        };
+        let err = plan.execute().await.unwrap_err();
```
pingyu, line 1235: The test passes even if `UndeterminedResult` is retried until exhaustion because every attempt returns the same error. Suggest to count `UndeterminedPlan::execute` calls and assert exactly one dispatch.

pingyu (commented):
Rest LGTM.

src/request/plan.rs
eduralph: Good catch — the error alone cannot distinguish "terminal" from "retried to exhaustion", because every attempt resurfaces the same error. Done: `UndeterminedPlan` now counts dispatches behind a shared `Arc`, and the terminal test asserts **exactly one dispatch** under a backoff that would have allowed 10 retries. The exhaustion test gained the complementary assertion (`RETRIES + 1` dispatches), so it now proves the default path actually retried rather than merely erred. I also checked the new assertion bites: with the terminal early-return disabled, the test fails with 11 dispatches counted.

Also rebased onto master now that #550 has merged, so the diff is down to this PR's own commit.

pingyu (approved):

pingyu:
cc @ekexium

src/transaction/transaction.rs
```diff
@@ -1340,7 +1340,31 @@ impl<PdC: PdClient> Committer<PdC> {
...
+                // PRECISION NOTE: if 1PC was configured but sharding disabled it on
+                // the actual requests (multi-region transactions), this is
+                // conservative — it can report UNDETERMINED for a prewrite that could
+                // not have committed. That is the safe direction: the caller verifies
+                // instead of replaying. The per-request 1PC state is not observable
+                // on the error path; precise gating belongs to the upfront
+                // single-region 1PC decision (client-go's checkOnePC shape) that full
+                // 1PC support will introduce.
+                return if (self.options.async_commit || self.options.try_one_pc)
+                    && crate::request::plan::is_undetermined_region_error(&e)
```
ekexium: Should we also check `is_grpc_error(&e)` here?

src/transaction/transaction.rs
```diff
@@ -1475,9 +1511,20 @@ impl<PdC: PdClient> Committer<PdC> {
...
+                // and propagate the error to the user.
+                //
+                // PROVENANCE NOTE: this plan wraps the commit in `resolve_lock`, so an
+                // undetermined error could in principle originate from an auxiliary
+                // CheckTxnStatus/ResolveLock RPC rather than the commit dispatch itself. That
+                // only over-reports (marks undetermined when the commit was actually rejected),
+                // which is the SAFE direction — the caller verifies instead of assuming a clean
+                // outcome. Distinguishing per-layer error provenance is a precision improvement,
+                // not a safety fix, and is left as follow-up.
+                if matches!(e, Error::Grpc(_))
```
ekexium: Should it use `is_grpc_error` here to match GrcpAPI error as well?

src/transaction/transaction.rs
eduralph: Yes — done. client-go agrees: it marks the committer undetermined for async-commit/1PC prewrites on any RPC error (prewrite.go, prewrite1BatchReqHandler.drop), alongside the UndeterminedResult region-error path, since for those modes a lost prewrite response leaves the commit state unknown. Both this site and the commit-primary one now share a single predicate, is_commit_outcome_unknown (= is_grpc_error || is_undetermined_region_error), with a unit test.

src/transaction/transaction.rs
eduralph: Yes — and your question uncovered that the Error::Grpc(_)-only match has been effectively dead since the tonic migration (#399). Under grpcio, RPC failures surfaced as Error::Grpc; with tonic, every KV dispatch failure takes the then-new Error::GrpcAPI(Status) (the impl_request! macro's .map_err(Error::from) on tonic::Status), and Error::Grpc narrowed to connection establishment — so the undetermined flag could no longer fire on a real dispatch failure. Switched to the shared is_commit_outcome_unknown helper, which matches client-go's sender.GetRPCError() semantics (any error from the RPC call — status errors and connection failures alike, commit.go). The new unit test pins the GrpcAPI case the old match missed

eduralph:
Pulling on the same thread surfaced one more gap, fixed in this push: raw CAS was terminal on UndeterminedResult but still replayed after a dispatch-stage gRPC error — yet a lost response is exactly as ambiguous, and the replay could compare against the first attempt's own write and report succeed = false for a write that happened. CAS now opts into retry_multi_region_terminal_on_ambiguous_outcome, which is additionally terminal on a dispatch-stage gRPC error. Deliberately narrow: sharding/connection errors still retry (the request was never sent), and commit points keep retrying gRPC errors — replaying a commit is idempotent and client-go's transport does the same (region_request.go, onSendFail). For CAS this is stricter than client-go, whose rawkv sendReq loop replays both cases; same safe-direction call as the async/1PC prewrite handling above. Two new dispatch-count tests pin both behaviors.

@pingyu  - you might want to re-look at the additional code

pingyu:
> Pulling on the same thread surfaced one more gap, fixed in this push: raw CAS was terminal on UndeterminedResult but still replayed after a dispatch-stage gRPC error — yet a lost response is exactly as ambiguous, and the replay could compare against the first attempt's own write and report succeed = false for a write that happened. CAS now opts into retry_multi_region_terminal_on_ambiguous_outcome, which is additionally terminal on a dispatch-stage gRPC error. Deliberately narrow: sharding/connection errors still retry (the request was never sent), and commit points keep retrying gRPC errors — replaying a commit is idempotent and client-go's transport does the same (region_request.go, onSendFail). For CAS this is stricter than client-go, whose rawkv sendReq loop replays both cases; same safe-direction call as the async/1PC prewrite handling above. Two new dispatch-count tests pin both behaviors.
> 
> @pingyu - you might want to re-look at the additional code

LGTM~

ekexium (approved):

PR #556, opened 2026-07-25
transaction: refuse to resolve shared locks instead of mis-handling them

eduralph:
> Stacked on #550 — please review the second commit only. Until #550 merges, the diff below also
> contains its re-vendor.

**Why:** the re-vendored kvproto exposes shared locks (`Op::SharedLock` /
`Op::SharedPessimisticLock`). Their contract is unusual: a shared lock's real holders live *only* in
`kvrpcpb.LockInfo.shared_lock_infos`, and the proto comment is explicit that you must "DO NOT read
from the wrapper LockInfo" — whose own `lock_version` and `primary_lock` are left unset (its `key`
is not; see below).

**What:** this client doesn't implement shared-lock resolution, and every *partial* handling of one
is worse than none — resolving the wrapper checks transaction 0; filtering on the wrapper's fields
(`use_async_commit`, `lock_type`) silently drops the real members; and the pessimistic-lock special
cases in the resolver don't know `SharedPessimisticLock`. So resolution **refuses** them with an
explicit error: in `resolve_locks`, in `LockResolver`'s crate-public entry point, and in
`CleanupLocks::execute` before any filter runs. Until real support lands, an explicit error is the
only answer that cannot roll back a live transaction or skip a dead one. Servers predating shared
locks never produce them, so this is a no-op there.

The API-v2 keyspace codecs are made shared-lock-aware separately, because they run on scan results
that never reach the resolver. They now take the shape of client-go's `codecV2.decodeLockInfo`
(`internal/apicodec/codec_v2.go`), which decodes a `LockInfo`'s own `Key` / `PrimaryLock` /
`Secondaries` and *then* recurses into `SharedLockInfos`, with no wrapper special case. This client
now does the same: convert the lock's **own** key fields, then recurse into the members.

**What the wrapper actually contains.** An earlier revision skipped a wrapper's own fields entirely,
reading the proto's "DO NOT read from the wrapper LockInfo" as covering them. Checking the writer —
`SharedLocks::into_lock_info` in TiKV's `components/txn_types/src/lock.rs`, identical at `v8.5.7`
and on `master` — that's only half right, and the half it gets wrong matters:

```rust
info.set_shared_lock_infos(shared_locks.into());
info.set_key(raw_key);
```

A wrapper sets `lock_type`, `shared_lock_infos` and `key` — the key it locks — and leaves
`primary_lock`/`lock_version` at their defaults. Each member is built from that *same* raw key plus
its own primary and version. So the caveat scopes the per-transaction fields, not the locked key.
Skipping the wrapper wholesale would have handed `scan_locks` a physical key beside decoded member
keys; converting it wholesale would have hit the length assertion on the unset primary. Hence no
wrapper special case, just a per-field guard — both fields are exercised by tests.

I'd still welcome a correction if that reading of the writer is off. It can't be settled by test at
today's pin: CI runs `TIKV_VERSION: v8.5.5`, which predates shared locks, so no integration test in
this repo can produce one — the coverage below is unit-level by necessity, and the refusal path is
unreachable against that server. Worth noting for a separate change: `v8.5.6` is the first release
carrying shared locks, so bumping the CI pin within the same release family would make these paths
reachable. Not proposing it here, given this PR is already a split-out.

The two directions guard differently, on purpose:

- **Truncating**, empty bytes can only mean *unset* (an encoded key always carries its 4-byte
  prefix), so unset fields are skipped rather than run into `pretruncate_bytes`' length assertion.
- **Encoding**, the input is a *logical* key and the empty logical key is valid in API v2, so
  nothing is skipped. `scan_locks` → `resolve_locks` round-trips locks through truncate-then-encode,
  and skipping empties there would strand a lock on the empty key with no prefix, sending resolution
  after an empty physical key. Wrappers need no encode-side guard because resolution refuses them
  first.

The panic hazard predates the re-vendor: a wrapper arrives the same way whichever proto vintage
parses it.

Full shared-lock support is deliberately follow-up work.

**Verification:** 4 new tests — the resolver refusal (a plain lock passes; a wrapper carrying members
and a lock marked shared only by its op are both refused); the codec converting a wrapper's own key
alongside its members; a wrapper's unset fields surviving truncation (the other reading, so both are
pinned); and a lock on the empty logical key round-tripping through truncate-then-encode. 71 lib
tests; `make check` green; txn/raw/failpoint integration suites green against a local api-v2 cluster;
and behaviour compared against client-go by driving both clients through identical scenarios — no
observable change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction lock cleanup when lock fields are missing, empty, or malformed.
  * Prevented shared-lock records from being mistakenly treated as ordinary locks during cleanup.
  * Preserved empty logical keys through encoding and decoding.
  * Added validation so unsupported shared-lock representations return an error instead of being silently altered.
  * Improved handling of nested lock information and lock cleanup reliability.

### DIFF
--- a/src/request/keyspace.rs
+++ b/src/request/keyspace.rs
@@ -160,23 +160,63 @@ impl TruncateKeyspace for Vec<KvPair> {
     }
 }
 
+/// Is this key field UNSET, as opposed to carrying the empty logical key?
+///
+/// An encoded key always carries its 4-byte keyspace prefix, so on the wire an empty
+/// byte string means "not set" — never "the empty logical key", which encodes to the
+/// bare prefix. Guarding on this keeps the codecs off `pretruncate_bytes`' length
+/// assertion for fields a shared-lock wrapper leaves unset, and mirrors client-go's
+/// `codecV2.DecodeKey`, which returns early on a zero-length key.
+fn is_unset_key(key: &[u8]) -> bool {
+    key.is_empty()
+}
+
 impl TruncateKeyspace for Vec<crate::proto::kvrpcpb::LockInfo> {
     fn truncate_keyspace(mut self, keyspace: Keyspace) -> Self {
         if !matches!(keyspace, Keyspace::Enable { .. }) {
             return self;
         }
         for lock in &mut self {
-            take_mut::take(&mut lock.key, |key| {
-                Key::from(key).truncate_keyspace(keyspace).into()
-            });
-            take_mut::take(&mut lock.primary_lock, |primary| {
-                Key::from(primary).truncate_keyspace(keyspace).into()
-            });
+            // Convert this lock's OWN key fields, then recurse into any shared-lock
+            // members — the shape of client-go's `codecV2.decodeLockInfo`, which
+            // decodes Key/PrimaryLock/Secondaries and *then* walks SharedLockInfos,
+            // with no wrapper special case.
+            //
+            // Per TiKV's writer (`SharedLocks::into_lock_info`, txn_types/src/lock.rs)
+            // a wrapper sets `lock_type`, `shared_lock_infos` and `key` — the key it
+            // locks — and leaves `primary_lock`/`lock_version` at their defaults. Each
+            // member is built from that SAME raw key plus its own primary and version.
+            // So the wrapper's key must be converted like any other (skipping it would
+            // hand `scan_locks` a physical key beside decoded member keys), while its
+            // unset fields must be left alone — hence no wrapper special case, just the
+            // per-field guard below.
+            if !is_unset_key(&lock.key) {
+                take_mut::take(&mut lock.key, |key| {
+                    Key::from(key).truncate_keyspace(keyspace).into()
+                });
+            }
+            if !is_unset_key(&lock.primary_lock) {
+                take_mut::take(&mut lock.primary_lock, |primary| {
+                    Key::from(primary).truncate_keyspace(keyspace).into()
+                });
+            }
             for secondary in lock.secondaries.iter_mut() {
+                // Unlike `key`/`primary_lock` above, this guard is not an unset-field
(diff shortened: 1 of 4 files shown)


src/request/keyspace.rs
```diff
@@ -160,23 +160,57 @@ impl TruncateKeyspace for Vec<KvPair> {
...
+                    Key::from(key).truncate_keyspace(keyspace).into()
+                });
+            }
+            if !is_unset_key(&lock.primary_lock) {
+                take_mut::take(&mut lock.primary_lock, |primary| {
+                    Key::from(primary).truncate_keyspace(keyspace).into()
+                });
+            }
             for secondary in lock.secondaries.iter_mut() {
+                if is_unset_key(secondary) {
```
pingyu, line 210: Here seems to be an malformed-input tolerance rather than an “unset secondary” case.

An unset secondaries field becomes an empty Vec. A present empty element inside secondaries is not a normal “unset” representation.

Suggest to comment about this.

pingyu (commented):
Rest LGTM.

src/request/keyspace.rs
eduralph: You are right — an unset `secondaries` is an empty `Vec` with no elements to visit, so a present-but-empty *element* can only be malformed input, and the guard there is tolerance rather than unset-field handling. Commented it as such (distinguishing it from the `key`/`primary_lock` guards above, which do meet genuinely unset fields), and added a small test — `a_malformed_empty_secondary_is_tolerated_not_truncated` — so the distinction is pinned rather than only narrated.

Also rebased onto master now that #550 has merged, so the diff is down to this PR's own commit.

pingyu (approved):

pingyu:
cc @ekexium

ekexium (approved):

PR #558, opened 2026-08-05
pd: bound every wait on the PD path — a frozen peer must not hang the client

eduralph:
**Why:** #516's `txn_cleanup_2pc_locks` hang, diagnosed end to end. In two CI runs (31042420081, 31043996146), the phase logs from #553 show every phase of the test completing in milliseconds; the only statement after the last phase is `count_locks`, which begins with `client.current_timestamp()` — a TSO request. The cluster-logs artifact shows the trigger: all three TiKVs logged `monotonic time jumped back` by ~65s simultaneously (a hypervisor-level pause of the runner VM), PD then measured a 2.6s scheduling gap on its 50ms TSO loop, lost its 1.67s election lease, resigned, and closed the TSO allocator — and **recovered ~15s later**, serving normally for the rest of the job, while the test stayed hung until nextest's 600s kill. Whatever holds the client after that point is client-side. And nothing about the trigger is CI-specific: any hypervisor pause or live migration, cgroup throttling under Kubernetes, or a middlebox silently expiring an idle long-lived connection produces the same frozen-open stream in production.

**The defects.** Nothing on the PD path bounded a wait against a peer that goes silent *without* terminating the connection:

1. The TSO receive loop waits on `responses.next()` unbounded — a request parked in `pending_requests` on a frozen-open stream hangs forever. (Termination is recoverable; silence is not.)
2. TSO stream creation (`pd_client.tso(...)`) is unbounded the same way.
3. The dial: a frozen peer's kernel still completes the TCP handshake, then the HTTP/2 preface never arrives and `connect()` hangs unbounded. `Connection::connect` even took a timeout parameter and ignored it (`_timeout`), and its `get_members` call carried no bound either.
4. Teardown alone cannot deliver failures: the cancelled gRPC request body owns the request channel inside the connection's task, which may not run again — releasing nothing — until the connection thaws (observed in reproduction: sends kept succeeding into a dead worker's channel).

**What:** follow client-go's shape — its TSO dispatcher arms a per-batch deadline with `defaultPDTimeout` = 3s and cancels the stream on expiry (`pd/client/clients/tso/dispatcher.go`), bounds stream creation via `checkStreamTimeout` (`stream.go`), and bounds dials — using the client's configured request timeout (`Config::timeout`, whose documented contract covers PD requests). Every bound is armed only while something is owed:

- The receive loop waits in whole timeout windows while batches are outstanding: a batch pending across one full silent window declares the stream stalled and the worker exits. A batch dispatched mid-window gets one full window, so teardown takes between one and two windows of genuine silence. Idle streams are never torn down.
- Stream creation is guarded by a watchdog with the same window rule, armed only while a batch is pending — PD (a gRPC-go server) sends response headers with its first response, so `tso(..)` legitimately waits while the client is idle; HTTP/2 lets request batches flow (and become pending) before response headers arrive, so a stalled creation is still detected.
- The worker's exit is *signaled*: `get_timestamp` selects on a watch channel the worker closes on exit — while enqueueing as well as while awaiting the response, since a dead worker's request channel can remain open with a full buffer and `send` would otherwise block forever.
- Dials are bounded, HTTP/2 handshake included: `SecurityManager` grows `connect_with_timeout` (the existing `connect` signature is preserved, now bounded by the default request timeout, so no API break), `Connection::connect` uses its hitherto-ignored timeout, and `get_members` is bounded like every other request in that path.

Stream termination is also observable now: a clean end stays info-level; stalls, creation timeouts, termination with batches outstanding, and stream status errors (previously swallowed) surface as warnings with the outstanding count — resolving the old TODO about indistinguishable terminations.

**Considered and omitted:** a timer-based backstop in `get_timestamp` (client-go has no analogue — ambient contexts cover it — and the worker-exit signal is event-driven and precise), and gRPC keepalive (connection-level health; client-go applies it to TiKV connections; a natural follow-up).

**What this means for #516:** the unbounded hang is gone. Under a short freeze the test recovers and passes; under a freeze longer than the retry budget it fails fast with a descriptive error instead of a 600s timeout — the trigger itself is outside any client's control.

**Verification:** 10 new unit tests — the receive loop and the creation watchdog driven under a paused clock (stall detected within [1, 2] windows; a mid-window batch granted its full window; idle streams and idle creation never torn down; termination clean only when nothing is owed; a stream status error surfaced; the happy path), `get_timestamp` failing fast when the worker dies, and enqueueing observing worker death with a full request buffer. The stall test bites: with the stall branch disabled it runs unbounded. 82 lib tests, `make check`, and the txn/raw/failpoint integration suites green against a local api-v2 cluster.

Manually, with `docker pause` on PD for 45s (SIGSTOP: TCP open, no FIN — the client-side view of a VM pause): **before**, one silent 44.5s hang ended only by the unpause; **after**, a 5s-idle client keeps its worker silently, the stall is detected in ~4s, reconnects against the frozen peer fail in 2s each, callers receive bounded, descriptive errors while the freeze lasts, and the client recovers within ~3s of the peer thawing.

Ref: #516

## Summary by CodeRabbit

- **Bug Fixes**
  - Added timeout handling for cluster connections, security handshakes, and initial membership requests.
  - Timestamp requests now honor configurable time limits and fail promptly when services stall or become unavailable.
  - Improved handling of timestamp stream errors, idle termination, and worker shutdowns to prevent requests from remaining unresolved.
  - Reconnection attempts now apply the configured timeout consistently.
- **Tests**
  - Expanded coverage for connection timeouts, timestamp allocation, stream failures, and worker shutdown behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ rstest = "0.18.2"
 serde_json = "1"
 serial_test = "0.5.0"
 tempfile = "=3.14.0"
-tokio = { version = "1", features = ["sync", "rt-multi-thread", "macros"] }
+tokio = { version = "1", features = ["sync", "rt-multi-thread", "macros", "test-util"] }
 
 [[test]]
 name = "failpoint_tests"
--- a/src/common/security.rs
+++ b/src/common/security.rs
@@ -76,10 +76,27 @@ impl SecurityManager {
     }
 
     /// Connect to gRPC server using TLS connection. If TLS is not configured, use normal connection.
-    pub async fn connect<Factory, Client>(
+    ///
+    /// The dial is bounded by the default request timeout — see
+    /// [`Self::connect_with_timeout`] to choose the bound.
+    pub async fn connect<Factory, Client>(&self, addr: &str, factory: Factory) -> Result<Client>
+    where
+        Factory: FnOnce(Channel) -> Client,
+    {
+        self.connect_with_timeout(addr, crate::config::DEFAULT_REQUEST_TIMEOUT, factory)
+            .await
+    }
+
+    /// Like [`Self::connect`], with `connect_timeout` bounding the whole dial,
+    /// HTTP/2 handshake included. A frozen peer can complete the TCP handshake
+    /// from its kernel (which keeps running when the process does not) and then
+    /// never answer the HTTP/2 preface, hanging an unbounded connect until the
+    /// peer thaws (#516).
+    pub async fn connect_with_timeout<Factory, Client>(
         &self,
         // env: Arc<Environment>,
         addr: &str,
+        connect_timeout: Duration,
         factory: Factory,
     ) -> Result<Client>
     where
@@ -91,7 +108,15 @@ impl SecurityManager {
         } else {
             self.default_channel(addr).await?
         };
-        let ch = channel.connect().await?;
+        let ch = tokio::time::timeout(connect_timeout, channel.connect())
+            .await
+            .map_err(|_| {
+                internal_err!(
+                    "connecting to {} timed out after {:?}",
+                    addr,
+                    connect_timeout
+                )
+            })??;
 
         Ok(factory(ch))
     }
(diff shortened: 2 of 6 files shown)


eduralph:
@pingyu for attention

Issue #528, opened 2026-01-27
Resolve lock bug


Surrina-ki:
https://github.com/tikv/client-rust/commit/7d80f59128196c80940cc864a90941f68514af8d#diff-3ac08ac53a85faf27689aba58b5ba02e85714d1d4e325835d018a76d48c60bf1
In this commit, the logic related to lock resolution was modified, but I've found that the current implementation is still not comprehensive enough. For example, with async commit, if only prewrite is executed and the secondary commits are never performed, subsequent reads of the written data will keep failing with "Failed to resolve lock".

Surrina-ki:
Currently, when checking the transaction status and encountering a lock, nothing is executed. However, I noticed that in client-go, when a lock is encountered and it belongs to an async commit transaction, the client does send requests such as CheckSecondaryLocksRequest and ResolveLockRequest to attempt to resolve the lock. Could you please take a look? @pingyu

pingyu:
Async commit is not implemented yet. Please refer to description of #519:

> Resolving pessimistic & async commit locks is not implemented yet.

Surrina-ki:
> Async commit is not implemented yet. Please refer to description of [#519](https://github.com/tikv/client-rust/pull/519):
> 
> > Resolving pessimistic & async commit locks is not implemented yet.

Is there a planned timeline for implementing async commit lock support?

Issue #509, opened 2026-01-05
tests: "txn_cleanup_range_async_commit_locks" is flaky


pingyu:
URL: https://github.com/tikv/client-rust/actions/runs/20656538490/job/59314998376

```
Nextest run ID cec4f7d2-35ba-4b4f-9df0-ed5b8bd98b27 with nextest profile: ci
    Starting 26 tests across 4 binaries (54 tests skipped)
        PASS [   2.648s] ( 1/26) tikv-client::failpoint_tests txn_cleanup_2pc_locks
        SLOW [> 60.000s] (─────) tikv-client::failpoint_tests txn_cleanup_async_commit_locks
        SLOW [>120.000s] (─────) tikv-client::failpoint_tests txn_cleanup_async_commit_locks
        PASS [ 171.758s] ( 2/26) tikv-client::failpoint_tests txn_cleanup_async_commit_locks
        PASS [   0.923s] ( 3/26) tikv-client::failpoint_tests txn_cleanup_locks_batch_size
        FAIL [   0.713s] ( 4/26) tikv-client::failpoint_tests txn_cleanup_range_async_commit_locks
        PASS [   6.105s] ( 5/26) tikv-client::failpoint_tests txn_optimistic_heartbeat
        PASS [   0.874s] ( 6/26) tikv-client::integration_tests txn_bank_transfer
        PASS [   0.123s] ( 7/26) tikv-client::integration_tests txn_batch_mutate_optimistic
        PASS [   4.155s] ( 8/26) tikv-client::integration_tests txn_batch_mutate_pessimistic
        PASS [   0.115s] ( 9/26) tikv-client::integration_tests txn_crud
        PASS [   0.124s] (10/26) tikv-client::integration_tests txn_get_for_update
        PASS [   0.019s] (11/26) tikv-client::integration_tests txn_get_timestamp
        PASS [   0.103s] (12/26) tikv-client::integration_tests txn_insert_duplicate_keys
        PASS [   0.107s] (13/26) tikv-client::integration_tests txn_key_exists
        PASS [  12.661s] (14/26) tikv-client::integration_tests txn_lock_keys
        PASS [  12.667s] (15/26) tikv-client::integration_tests txn_lock_keys_error_handle
        PASS [   0.105s] (16/26) tikv-client::integration_tests txn_pessimistic
        PASS [   0.127s] (17/26) tikv-client::integration_tests txn_pessimistic_delete
[...]

PR #561, opened 2026-08-31
fix(transaction): recover expired async commit locks

ariesdevil:
Recover expired async-commit locks by checking all secondary locks before resolving, instead of retrying CheckTxnStatus forever (#528).

Also bundled:
- cleanup_locks: take max with the primary lock's min_commit_ts when computing the commit version from secondaries
- CheckSecondaryLocks merge: return an error instead of panicking on conflicting commit TS across regions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery of expired asynchronous transaction locks, allowing reads to proceed after interrupted commits.
  * Improved handling of stale, missing, conflicting, or invalid locks across regions.
  * Improved rollback behavior for transactions with mismatched primary locks.
  * Added clearer errors for invalid transaction responses.
  * Prevented incomplete or unresolved transaction states from being incorrectly cached.

* **Tests**
  * Added coverage for asynchronous commit recovery, rollback scenarios, multi-region transactions, lock validation, and transaction status resolution.


ariesdevil:
**NOTE:** This is the initial impl, subsequent PRs will align the impl of client-go.

ariesdevil:
The failed CI seems to be a flaky test, first seen in #509 , I'll try to fix it.

ariesdevil:
Hi @ekexium, if you have time, please take a look at this PR.

ekexium:
Async-commit protocol is error-prone, is this implementation ported from client-go? AI review finds a subtle issue:

One potential issue in the pessimistic-lock path: LockNotExistDoNothing is mapped to RolledBack, leading to a region-wide rollback that could remove valid locks after a primary switch. client-go [excludes this action from final rollback](https://github.com/tikv/client-go/blob/52c1e76cec993571493c81de442bcbef90cdc106/txnkv/txnlock/lock_resolver.go#L138) and [uses pessimistic rollback](https://github.com/tikv/client-go/blob/52c1e76cec993571493c81de442bcbef90cdc106/txnkv/txnlock/lock_resolver.go#L629).
Could you check this case?

ariesdevil:
Nice catch; confirmed and fixed, thanks, PTAL again.

`TransactionStatusKind` ignored `action`, so `LockNotExistDoNothing` / `TtlExpirePessimisticRollback` became `RolledBack` and triggered a region-wide `ResolveLock`. Worse, that status was also cached, so a later prewrite lock of the same txn in the same call would be swept too.

Fix mirrors client-go: a pessimistic lock whose txn is no longer `Locked` now gets a single-key `PessimisticRollback` only (skipped when the lock is its own primary), and `is_cacheable` now requires a determined status (`NoAction` / `LockNotExistRollback` / `TtlExpireRollback`), like `IsStatusDetermined`. The GC path keeps rolling those back, matching `BatchResolveLocks`.

And added tests for all four reply kinds, the primary-key skip, and the cache case; they fail on the old code.

ekexium (approved):

ariesdevil:
Hi @pingyu, if you have time, PTAL as well.

